PR #1, opened 2018-06-30
fixed paths and tag dictionary

kashif:


kashif:
No worries! I see the same zip loop in other places... what does it do I couldn’t figure it out...

PR #11, opened 2018-07-27
GH-10: Added text classification model and trainer

### LINKED ISSUE #10: Add code for training a text-classifier
Currently, we cannot use flair to train a text-classifier. 

We want to add a 
- data fetcher for loading a labeled corpus
- text classification model
- trainer for the text-classification model

tabergma:
closes #10


alanakbik:
Great work!

Issue #17, opened 2018-07-29
Add Trainer for custom language models


alanakbik:
Flair currently supports training of custom sequence taggers, but not custom language models. Many people want to try out contextual string embeddings for new languages (#2) or domains. 

Task: add a trainer class to facilitate training of new language model embeddings.

alanakbik:
LanguageModelTrainer class added to release 0.2 branch

Issue #15, opened 2018-07-27
Increase test coverage


tabergma:
We should add tests for 
- all embeddings
- the data classes

We could also add tests for training, to make sure training works without issues.

tabergma:
closed via https://github.com/zalandoresearch/flair/pull/30

PR #18, opened 2018-07-30
GH-17: language model embeddings

alanakbik:
- Added LanguageModelTrainer class in trainer package to enable training of customer char LM embeddings (#17)

- simplified loader for standard character dictionary

- updated stored character LMs to new package structure

- Added pytest for LanguageModelTrainer (#15)


tabergma:
:+1:

Issue #19, opened 2018-07-30
simplify SequenceTagger and SequenceTaggerTrainer


alanakbik:
possible simplifications to the SequenceTagger:

- use of Pytorch builtin LearningRateSchedulers
- auto-spawning on GPUs
- switching of evaluation metrics (F1, accuracy)

alanakbik:
Ready for release 0.2!

PR #22, opened 2018-07-30
GH-19: simplify sequence tagger

alanakbik:
- added learning rate scheduler to SequenceTaggerTrainer and removed old annealing code
- now use Metric class in SequenceTaggerTrainer
- added auto-spawn on GPU for all three model classes
- removed move-to-gpu commands in train.py and tutorials (no longer necessary due to auto-spawn)

### DIFF
--- a/flair/embeddings.py
+++ b/flair/embeddings.py
@@ -70,7 +70,7 @@ class TokenEmbeddings(Embeddings):
 
 
 class DocumentEmbeddings(Embeddings):
-    """Abstract base class for all document-level embeddings. Ever new type of document embedding must implement these methods."""
+    """Abstract base class for all token-level embeddings. Ever new type of word embedding must implement these methods."""
 
     @property
     @abstractmethod
@@ -208,6 +208,9 @@ class WordEmbeddings(TokenEmbeddings):
                 else:
                     word_embedding = np.zeros(self.embedding_length, dtype='float')
 
+                # if torch.cuda.is_available():
+                #     word_embedding = torch.cuda.FloatTensor(word_embedding)
+                # else:
                 word_embedding = torch.FloatTensor(word_embedding)
 
                 token.set_embedding(self.name, word_embedding)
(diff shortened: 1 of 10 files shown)


flair/embeddings.py
```diff
@@ -70,7 +70,7 @@ def embedding_type(self) -> str:
 
 
 class DocumentEmbeddings(Embeddings):
-    """Abstract base class for all document-level embeddings. Ever new type of document embedding must implement these methods."""
+    """Abstract base class for all token-level embeddings. Ever new type of word embedding must implement these methods."""
```
tabergma: Redo renaming?

flair/trainers/sequence_tagger_trainer.py
```diff
@@ -254,12 +240,18 @@ def evaluate(self, evaluation: List[Sentence], evaluate_with_fscore: bool = True
             main_result = re.sub('accuracy', 'acc', main_result)
 
             f_score = float(re.findall(r'\d+\.\d+$', main_result)[0])
-
             return f_score, fp, main_result
 
-        precision: float = tp / (tp + fp)
+        if evaluation_method == 'accuracy':
+            score = metric.accuracy()
+            accuracy: float = tp / (tp + fp)
```
tabergma: Why calculating accuracy on your own when using the metric class?

flair/trainers/sequence_tagger_trainer.py
```diff
@@ -254,12 +240,18 @@ def evaluate(self, evaluation: List[Sentence], evaluate_with_fscore: bool = True
...
+            print(accuracy)
+            return score, fp, str(score)
 
-        return precision, fp, str(precision)
+        if evaluation_method == 'F1':
+            print(metric.accuracy())
```
tabergma: Why are you printing the accuracy if you are using f-score as metric?

flair/embeddings.py
alanakbik: good point! will redo!

PR #35, opened 2018-08-02
GH-20: Improve text classifier

### LINKED ISSUE #20: Improve TextClassifier
- TextClassifier is currently not working on gpu
- Use of Pytorch builtin LearningRateSchedulers

tabergma:
closes #20 

- use ReduceLROnPlateau in TextClassifier
- Fix bugs in training TextClassifier on GPU
- TextClassifier now takes a DocumentEmbeddings instead of a list of TokenEmbeddings
- DocumentLSTMEmbedding now takes additional parameters to define (1) the output dimension of the reprojection of words and (2) whether only the last representation or a combination of last and first representation of the LSTM should be taken as document embedding

### DIFF
--- a/flair/embeddings.py
+++ b/flair/embeddings.py
@@ -357,13 +357,6 @@ class CharLMEmbeddings(TokenEmbeddings):
         self.detach = detach
 
         self.is_forward_lm: bool = self.lm.is_forward_lm
-        if self.is_forward_lm:
-            print('FORWARD language mode loaded')
-        else:
-            print('BACKWARD language mode loaded')
-
-        print('on cuda:')
-        print(next(self.lm.parameters()).is_cuda)
 
         dummy_sentence: Sentence = Sentence()
         dummy_sentence.add_token(Token('hello'))
@@ -433,7 +426,7 @@ class DocumentMeanEmbeddings(DocumentEmbeddings):
         super().__init__()
 
         self.embeddings: StackedEmbeddings = StackedEmbeddings(embeddings=word_embeddings)
-        self.name: str = 'word_mean'
+        self.name: str = 'document_mean'
         self.reproject_words: bool = reproject_words
         self.static_embeddings: bool = not reproject_words
 
@@ -441,6 +434,10 @@ class DocumentMeanEmbeddings(DocumentEmbeddings):
         self.__embedding_length = self.embeddings.embedding_length
 
         self.word_reprojection_map = torch.nn.Linear(self.__embedding_length, self.__embedding_length)
+        torch.nn.init.xavier_uniform_(self.word_reprojection_map.weight)
+
+        if torch.cuda.is_available():
+            self.cuda()
 
     @property
     def embedding_length(self) -> int:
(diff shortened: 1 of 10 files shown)


flair/embeddings.py
```diff
@@ -420,51 +413,56 @@ def _add_embeddings_internal(self, sentences: List[Sentence]) -> List[Sentence]:
                 offset_backward -= 1
                 offset_backward -= len(token.text)
 
-                token.set_embedding(self.name, embedding.cpu())
```
alanakbik: Why remove the .cpu() command? Embeddings should always be returned on cpu.

flair/embeddings.py
```diff
@@ -297,7 +297,7 @@ def _add_embeddings_internal(self, sentences: List[Sentence]):
                 character_embeddings[d[i]] = chars_embeds_temp[i]
 
             for token_number, token in enumerate(sentence.tokens):
-                token.set_embedding(self.name, character_embeddings[token_number].cpu())
```
alanakbik: Why remove the .cpu() command? Embeddings should always be returned on cpu.

flair/embeddings.py
```diff
@@ -477,49 +475,66 @@ def embed(self, paragraphs: Union[List[Sentence], Sentence]):
...
+        if self.reproject_words and reproject_words_dimension is not None:
+            self.embeddings_dimension = reproject_words_dimension
 
         # bidirectional LSTM on top of embedding layer
-        self.word_reprojection_map = torch.nn.Linear(self.length_of_all_word_embeddings,
-                                                     self.length_of_all_word_embeddings)
-        self.rnn = torch.nn.LSTM(self.length_of_all_word_embeddings, hidden_states, num_layers=num_layers,
+        self.word_reprojection_map = torch.nn.Linear(self.length_of_all_token_embeddings,
+                                                     self.embeddings_dimension)
+        self.rnn = torch.nn.GRU(self.embeddings_dimension, hidden_states, num_layers=num_layers,
```
alanakbik: why GRU and not LSTM?

flair/models/text_classification_model.py
```diff
@@ -89,26 +78,20 @@ def load_from_file(cls, model_file):
         # ATTENTION: suppressing torch serialization warnings. This needs to be taken out once we sort out recursive
         # serialization of torch objects
         warnings.filterwarnings("ignore")
-        state = torch.load(model_file, map_location={'cuda:0': 'cpu'})
```
alanakbik: I believe this is needed if a model was trained on GPU but is loaded on CPU

flair/embeddings.py
tabergma: I moved the `.cpu()` command directly to the method `set_embeddings()`. This way, you don't have to remember that you should call it when setting the embeddings.

flair/models/text_classification_model.py
tabergma: It is not working if you load the GPU model on a GPU machine again. I'll try to load a model trained on GPU on a CPU machine.

flair/embeddings.py
tabergma: During training GRU worked slightly better than LSTMs on the small datasets.

flair/embeddings.py
alanakbik: ahh ok!!

flair/embeddings.py
alanakbik: :+1:

alanakbik:
great!!

Issue #8, opened 2018-07-19
Documentation


tabergma:
- Fixing typos
- Add missing documentation/tutorials
- Add missing method documentation

PR #40, opened 2018-08-03
GH-8: embedding tutorial

alanakbik:
Added tutorial on how to train LM embeddings

Added explanation of batch tagging


tabergma:
:+1:

PR #49, opened 2018-08-07
docs: use GloVe and English news embeddings for AGNews example

stefan-it:
Hi,

this PR corrects both word and character embeddings for the AGNews text classification example. Now GloVe embeddings are used. Originally, embeddings for German are used, which are not suitable for English news articles.


alanakbik:
Thanks for correcting this!

Issue #5, opened 2018-07-11
Add class for FastText out-of-vocabulary word embeddings


joelkuiper:
Seems like 

https://github.com/zalandoresearch/flair/blob/master/flair/embeddings.py#L139

 Does not make use of the OoV functionality present in FastText, seems like it would be a nice addition!

davidsbatista:
I also just recently noticed that the fastText corpus downloaded by Flair doesn't contain n-grams stored, is anyone taking the this? I could eventually look into it, but I guess it depends on having fastText models containing subword (i.e., n-gram) available.

alanakbik:
Yes, we haven't gotten around to looking into this, so currently we only use FastText without the subword features. So any help here would be appreciated! 

As a note of caution, we've observed ourselves and heard from other groups that the subword features often negatively impact downstream task performance. One guess is that this is because of how unknown words are handled - without subwords, they are simply marked as UNKs so the model can learn to deal with unknown words. But with subwords, an embedding is always produced based on ngrams which may not always be a good embedding and thus produce errors.

pranaychandekar:
Hey guys. As a part of experimentation, I needed this feature and so I wrote one file to do the same. Anyone who needs this functionality, feel free to use it - https://github.com/pranaychandekar/fasttext-embeddings-with-flair

alanakbik:
Hey that's great - can't wait to try it out! Would you like to add this in a PR?

pranaychandekar:
Sure @alanakbik ! Would love to do that.

alanakbik:
Closed by #879 :)

PR #60, opened 2018-08-10
GH-5: Make use of subword embeddings if supported

jbaiter:
This is an attempt at implementing https://github.com/zalandoresearch/flair/issues/5.
No extra classes were neccessary, this PR just changes the approach taken for checking if a word is OOV.

Previously, flair would store a set of all known words from the Gensim `KeyedVectors` class.
I changed this to use the `__contains__` implementation on `KeyedVectors`, which does just that, without any extra data structures. However, if the embeddings are of type `FastTextKeyedVectors`, Gensim will also use the subword-embeddings in case of an OOV (https://github.com/RaRe-Technologies/gensim/blob/develop/gensim/models/keyedvectors.py#L1928).

Unfortunately, this can't currently be used with the FastText-embeddings shipped with flair, since they don't actually have any subword-embeddings, the `KeyedVectors` class is just `Word2VecKeyedVectors` (i.e., only word embeddings).

### DIFF
--- a/flair/embeddings.py
+++ b/flair/embeddings.py
@@ -189,8 +189,6 @@ class WordEmbeddings(TokenEmbeddings):
 
         self.precomputed_word_embeddings = gensim.models.KeyedVectors.load(embeddings)
 
-        self.known_words = set(self.precomputed_word_embeddings.index2word)
-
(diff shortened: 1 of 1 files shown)


alanakbik:
Looks good, thanks very much! As next step, we'll prepare and test some subword models for shipping with Flair!

PR #64, opened 2018-08-13
GH-62: Update get_item for sentence and add repr for token

### LINKED ISSUE #62: getting items in sentence indices
2 issues here

- the ```__getitem__``` should return something for sentence[0] (in python this is usually the first thing)
- the ```flair.data.Token``` object should display the string in it's __repr__ method

tabergma:
closes #62

### DIFF
--- a/flair/data.py
+++ b/flair/data.py
@@ -175,7 +175,7 @@ class Sentence:
                 self.add_token(Token(word))
 
     def __getitem__(self, token_id: int) -> Token:
-        return self.get_token(token_id)
+        return self.get_token(token_id + 1)
 
     def __iter__(self):
         return iter(self.tokens)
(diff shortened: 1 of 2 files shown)


flair/data.py
```diff
@@ -175,7 +175,7 @@ def __init__(self, text: str = None, use_tokenizer: bool = False, labels: List[s
                 self.add_token(Token(word))
 
     def __getitem__(self, token_id: int) -> Token:
-        return self.get_token(token_id)
+        return self.get_token(token_id + 1)
```
alanakbik: get_token does a full iteration over all tokens in the sentence to identify the one with the right id. 

perhaps more efficient would be: 

`token = self.tokens[idx]`

that saves us the iteration.

alanakbik:
the example in TUTORIAL_BASICS.md also needs to be updated with the new index

flair/data.py
alanakbik: to clarify, the whole __getitem__ would then be: 

```python
    def __getitem__(self, idx: int) -> Token:
        token = self.tokens[idx]
        if token is None:
            raise IndexError()
        return token
```

Issue #65, opened 2018-08-14
Simplify data fetchers


alanakbik:
There is a lot of redundant code in the data fetcher helper routines. Simplify by creating a generic 'CoNLL-column' data reader that can be passed the column definition, so it is applicable to CoNLL03, CoNLL2000 and any other sequence labeling data that is formatted in a similar column style.

alanakbik:
Data fetchers now simplified. You can read any CoNLL column formatted file by passing a dict that specifies what column is what field. 

For instance: 

```python
from flair.data_fetcher import NLPTaskDataFetcher

sentences = NLPTaskDataFetcher.read_column_data('/path/to/conll03/data', column_name_map={0: 'text', 3: 'ner'})

for sentence in sentences:
    print(sentence.to_tagged_string()) 
```

Will read a CoNLL-03 formatted file and map the first column (index 0) to the lexical value of each word (i.e. the token text) and column index 3 as the NER tag.

PR #66, opened 2018-08-14
GH-65: refactored to generic column format and string enums

alanakbik:
Refactored to generic column format

- removed task-specific reader functions such as read_germeval() and read_conll03() etc.
- added generic reader function to which one can pass a column format as dict
- enum is now string instead of int
- more comments


flair/data_fetcher.py
```diff
@@ -7,250 +7,252 @@
...
+
             else:
                 # print(line)
                 fields: List[str] = re.split("\s+", line)
-                token = Token(fields[0])
-                token.add_tag('pos', fields[1])
-                token.add_tag('np', fields[2])
-                if len(fields) > 3:
-                    token.add_tag('ner', fields[3])
+                # print(len(fields))
```
tabergma: Could be removed.

flair/data_fetcher.py
```diff
@@ -7,250 +7,252 @@
...
+            if column_name_map[column] == 'text':
+                text_column = column
 
         sentence: Sentence = Sentence()
         for line in lines:
 
+            if line.startswith('#'):
+                continue
+
             if line == '':
```
tabergma: Could be replaced by `if not line`.

tabergma:
:+1:

Issue #37, opened 2018-08-02
Underutilization of GPU


ducalpha:
When running an experiment with NER ([run_ner.py](https://gist.github.com/ducalpha/fff18d22ab1ebdd6bfcd7952de3ba4be)), I found that the model does not utilize my GPU fully. I have 2 GPUs Nvidia Titan Xp on my machine but only 1 of them is used with the utilization varying from 10-60%.

```
+-----------------------------------------------------------------------------+
| NVIDIA-SMI 396.45                 Driver Version: 396.45                    |
|-------------------------------+----------------------+----------------------+
| GPU  Name        Persistence-M| Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp  Perf  Pwr:Usage/Cap|         Memory-Usage | GPU-Util  Compute M. |
|===============================+======================+======================|
|   0  TITAN Xp            Off  | 00000000:03:00.0 Off |                  N/A |
| 40%   61C    P2    65W / 250W |   5367MiB / 12196MiB |     15%      Default |
+-------------------------------+----------------------+----------------------+
|   1  TITAN Xp            Off  | 00000000:04:00.0 Off |                  N/A |
| 23%   35C    P8    10W / 250W |     10MiB / 12188MiB |      0%      Default |
+-------------------------------+----------------------+----------------------+
                                                                               
+-----------------------------------------------------------------------------+
| Processes:                                                       GPU Memory |
|  GPU       PID   Type   Process name                             Usage      |
|=============================================================================|
|    0     30384      C   python                                      5357MiB |
+-----------------------------------------------------------------------------+
```

Training my data [NeuroNLP2](https://github.com/XuezheMax/NeuroNLP2) yields utilization of both GPUs at more than 90%.

alanakbik:
Very interesting! Multi-GPU support is definitively on our list, possibly for release 0.3. With regards to GPU usage that is strange - we have to run some tests to see what happens here.

alipetiwala:
@alanakbik 

same issue here on NER model

+-----------------------------------------------------------------------------+
| NVIDIA-SMI 410.79       Driver Version: 410.79       CUDA Version: 10.0     |
|-------------------------------+----------------------+----------------------+
| GPU  Name        Persistence-M| Bus-Id        Disp.A | Volatile Uncorr. ECC |
| Fan  Temp  Perf  Pwr:Usage/Cap|         Memory-Usage | GPU-Util  Compute M. |
|===============================+======================+======================|
[...]

PR #69, opened 2018-08-16
Gh 47 refactor crf

blythed:
Speed up pull request, still lots of work to do on speed.


alanakbik:
Partially addresses #37 with a rough x2 speedup in our default sequence tagging setup.

Issue #54, opened 2018-08-09
Enhancement: NER Output not as string but nested dict


pwichmann:
It would be great if there was a way to obtain a nested dictionary as output of the NER instead of a string with <...> tags. The string is quite tedious to work with.

I imagine an output like:

```
"sentence": {
    "text": "Facebook, Inc. is a company, and Google is one as well.",
    "named_entities": [
        {
            "mention_text": "Facebook, Inc.",
            "start_pos": 0
            "end_pos": ...
            "type": "ORG",
            "confidence": 0.9
        },
        {
        ...
        }
    ]
} 
```

pwichmann:
Thank you!

That does not solve the problem though.
It could be the case that words or word sequences are spelled the same but have different NER tags - depending on their context. That information is lost without the offset information which allows me to unambiguously identify the entity in the text.

I just had a similar problem with named entity mentions that contained other named entity mentions. It was a mess to process these without offset information. In fact, I just had to give up. :(

tabergma:
in release-0.3

You need to call `sentence.to_dict(tag_type='ner')` to obtain the dictionary with the original sentence text, entities and labels.

PR #72, opened 2018-08-16
GH-54: token offsets

alanakbik:
- Added `whitespace_after` field to Token (see #54)
- Sentence tokenizer now records `whitespace_after`
- Added method to infer `whitespace_after` field for already tokenized data
- Distringuish between `to_plain_string()` and `to_tokenized_string()` of `Sentence` - the former produces original string, the latter whitespace-tokenized string

### DIFF
--- a/flair/data.py
+++ b/flair/data.py
@@ -382,13 +382,13 @@ class Sentence:
     def __str__(self) -> str:
         return 'Sentence: "' + ' '.join([t.text for t in self.tokens]) + '" - %d Tokens' % len(self)
 
-    def to_plain_string(self) -> str:
-        return ' '.join([t.text for t in self.tokens])
-
     def __len__(self) -> int:
         return len(self.tokens)
 
-    def to_real_string(self):
+    def to_tokenized_string(self) -> str:
+        return ' '.join([t.text for t in self.tokens])
+
+    def to_plain_string(self):
         plain = ''
         for token in self.tokens:
             plain += token.text
--- a/flair/data.py
+++ b/flair/data.py
@@ -216,7 +216,7 @@ class Sentence:
                 # determine offsets for whitespace_after field
                 index = text.index
                 running_offset = 0
-                last_word_offset = -2
+                last_word_offset = -1
                 last_token = None
                 for word in tokens:
                     token = Token(word)
@@ -224,8 +224,8 @@ class Sentence:
                     try:
                         word_offset = index(word, running_offset)
                     except:
-                        word_offset = last_word_offset = +1
-                    if word_offset - 1 == last_word_offset:
+                        word_offset = last_word_offset + 1
+                    if word_offset - 1 == last_word_offset and last_token is not None:
                         last_token.whitespace_after = False
                     word_len = len(word)
                     running_offset = word_offset + word_len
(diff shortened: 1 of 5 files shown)


flair/data.py
```diff
@@ -198,22 +200,76 @@ def __init__(self, text: str = None, use_tokenizer: bool = False, labels: List[L
...
+                running_offset = 0
+                last_word_offset = -2
+                last_token = None
+                for word in tokens:
+                    token = Token(word)
+                    self.add_token(token)
+                    try:
+                        word_offset = index(word, running_offset)
+                    except:
+                        word_offset = last_word_offset = +1
```
stefan-it: @alanakbik This line looks a bit weird, I mean the `= +1` operator here.

alanakbik: @stefan-it yeah, unfortunately in some (very) rare cases the segtok tokenization produces a strange error in which it produces a token that is not actually in the sentence. this can cause the index search to fail, so just to be sure there is a fallback in such cases to default to the next position in the sentence.

flair/data.py
```diff
@@ -198,22 +200,76 @@ def __init__(self, text: str = None, use_tokenizer: bool = False, labels: List[L
...
+                    if word_offset - 1 == last_word_offset:
+                        last_token.whitespace_after = False
```
tabergma: If we fail to determine the word_offset for the first word in the sentence, the `word_offset` would be -1 and the `last_word_offset` would be -2. In this case `word_offset - 1 == last_word_offset` would be true and you would try to set `whitespace_after` to False for `None`. This will throw an exception.

tabergma: Can't you set `last_word_offset` to -1 and check for `word_offset - 1 == last_word_offset and last_token is not None`?

flair/data.py
```diff
@@ -198,22 +200,76 @@ def __init__(self, text: str = None, use_tokenizer: bool = False, labels: List[L
...
+        quote_count: int = 0
+        # infer whitespace after field
+
+        for token in self.tokens:
+            if token.text == '"':
+                quote_count += 1
+                if quote_count % 2 != 0:
+                    token.whitespace_after = False
+                elif last_token is not None:
+                    last_token.whitespace_after = False
```
tabergma: This does not work for a sentence like: `xyz "abc"`. With the current code you would set `whitespace_after` to False for `xyz`.

flair/data.py
alanakbik: oh yes, you're right!

flair/data.py
alanakbik: I think it would work: 

```python
sentence: Sentence = Sentence('xyz "abc".', use_tokenizer=True)
print(sentence.to_tokenized_string())
print(sentence.to_plain_string())
```
gives me: 
 
```
xyz " abc " .
xyz "abc".
```

whitespace_after here depends on the `quote_count`.

tabergma:
:+1:

djstrong:
@alanakbik Why (Flair) embeddings are calculated over `to_tokenized_string` and not over `to_plain_string`? `to_tokenized_string` inserts spaces between tokens, especially before punctuation. It is inconsistent with trained language model.

Issue #85, opened 2018-08-21
sequence_tagger_trainer: `save_model` should be `True` by default


stefan-it:
Hi,

at the moment, no trained models are saved by default. This affects almost all training examples.

Can I open a PR to set the `save_model` to `True` in the `sequence_tagger_trainer` class?

alanakbik:
Hi, yes you're right! It would be better to set `save_model` to `True` as default! We'd appreciate a PR!

PR #86, opened 2018-08-22
GH-85: Set save_model to True as default

stefan-it:
Hi,

as discussed in #85 this PR sets the `save_model` parameter in `sequence_tagger_trainer` to `True` as new default.


alanakbik:
Thanks!

PR #92, opened 2018-08-27
GH-8: tutorials

alanakbik:
Small simplifications to tutorials:
- moved explanation on text classification data format from BASICS to TRAINING A MODEL, since the format is only used in context of training a classification model. This makes the basics section a bit lighter and the classification explanation more grouped. 
- corrected typo


tabergma:
:+1:

Issue #97, opened 2018-08-29
Named entity starts with "I" tag


pwichmann:
I came across an unexpected NER tagging result where a named entity starts with the ```I-MISC``` tag.
I had assumed named entities can only start with ```B``` (beginning) or ```S``` (single token).

**The pre-tokenized raw sentence:**
```Laurel : Walter Asonevich , president of Pennsylvania Highlands Community College , recently was presented with the Boy Scouts of America Laurel Highlands Council 2018 Distinguished Citizen Award .```

**The tagged result:**
```Laurel <S-LOC> : Walter <B-PER> Asonevich <E-PER> , president of Pennsylvania <B-ORG> Highlands <I-ORG> Community <I-ORG> College <E-ORG> , recently was presented with the Boy <B-ORG> Scouts <I-ORG> of <I-ORG> America <I-ORG> Laurel <I-ORG> Highlands <I-ORG> Council <E-ORG> 2018 Distinguished Citizen <I-MISC> Award <E-MISC> .```

**The unexpected results / potential bug**
Note the last two tags: Citizen Award gets recognised as a named entity - but the sequence starts with ```I-MISC``` (after the previous named entity has ended on ```E-ORG```). ```I``` should be reserved for inner tokens, or am I mistaken?

This messed up my tag parser since I assumed named entities HAVE to start with ```B``` or ```S``` type tags.

Did I make a mistake?

Added information:
* basic model: 'ner'

alanakbik:
Thanks for reporting this. This seems strange - I'd like to take a closer look. Are you using the master branch or the latest pip version? How often does such an error occur in, say, 1000 sentences?

pwichmann:
@alanakbik It is a rare occurrence (maybe <1 in 1000 sentences). That's why it was so surprising and took me hours to find as the root cause.

I used latest pip version.

Please assume I am an absolute noob. But given the sentence, you should be able to replicate the error.

alanakbik:
Ah ok! Such errors can happen in hopefully very rare cases because the `B` and `S` logic is not hard-coded in the decoder. The CRF learns this logic from the training data, but there may be sentences where it still believes that an I- tag following an untagged word is most plausible given the model.

We'll look into this some more and see if hard-coding BIO or BIOES logic makes a difference!

@blythed: any thoughts?

pwichmann:
[...]

Issue #79, opened 2018-08-19
call eval.pl bugs


TanyaZhao:
when the code call eval.pl for evaluation, it causes this bug:

Traceback (most recent call last):
  File "char_lm.py", line 68, in <module>
    max_epochs=150)
  File "E:\users\v-tizhao\vs\flair\trainers\sequence_tagger_trainer.py", line 114, in train
    embeddings_in_memory=embeddings_in_memory)
  File "E:\users\v-tizhao\vs\flair\trainers\sequence_tagger_trainer.py", line 235, in evaluate
    p = run(eval_script, stdout=PIPE, input=eval_data, encoding='utf-8')
  File "E:\users\v-tizhao\anaconda\lib\subprocess.py", line 403, in run
    with Popen(*popenargs, **kwargs) as process:
  File "E:\users\v-tizhao\anaconda\lib\subprocess.py", line 709, in __init__
    restore_signals, start_new_session)
  File "E:\users\v-tizhao\anaconda\lib\subprocess.py", line 997, in _execute_child
    startupinfo)
OSError: [WinError 193] %1 is not a valid Win32 application

btw, I run the code on windows server, 64bit machine.
Could you help me? Thank you !

stefan-it:
The problem is, that a perl script is downloaded and used for the evaluation process. I guess this causes some problems on Windows.

In the upcoming release of `flair` could come with  a pure python implementation of that evaluation script, see  #75 :)

TanyaZhao:
one more questions：how could I get ontonotes 5.0 dataset? 
I saw that in your data_fetcher.py, there is "train.conllu", "test.conllu", "dev.conllu". Are these ontonotes data which have converted into conll format?  Where could I get them?
Could you please tell me the path? 
Thank you very much!

stefan-it:
This repository claims to have the Ontonotes 5.0 in CoNLL format:

https://github.com/ontonotes/conll-formatted-ontonotes-5.0

tabergma:
in release-0.3

PR #113, opened 2018-09-14
GH 75 spans

alanakbik:
Adds methods to retrieve BIO/BIOES annotated spans from sentences and evaluate using F1 score. 

Specifically: 
- An initial simple Span class in data.py that holds a Token span. This class may be expanded in future commits.
- A method in Sentence to extract spans of a tag type, such as NER spans.
- Test methods to check if span extraction handles BIO/BIOES and other formats and irregularities (#97)
- Change of evaluation method in SequenceTaggerTrainer to use our own evaluation routine instead of perl script (#79 )
- A new ItemWeigher helper class to count weighted strings

### DIFF
--- a/flair/data.py
+++ b/flair/data.py
@@ -421,7 +421,7 @@ class Sentence:
         for index, tag in enumerate(tags):
             self.tokens[index].add_tag(tag_type, tag)
 
-    def _infer_space_after(self):
+    def infer_space_after(self):
         """
         Heuristics in case you wish to infer whitespace_after values for tokenized text. This is useful for some old NLP
         tasks (such as CoNLL-03 and CoNLL-2000) that provide only tokenized data with no info of original whitespacing.
--- a/flair/data.py
+++ b/flair/data.py
@@ -312,8 +312,8 @@ class Sentence:
             if tag[0:2] in ['B-', 'S-']:
                 starts_new_span = True
 
-            # if previous_tag[0:2] in ['S-', 'E-']:
-            #     starts_new_span = True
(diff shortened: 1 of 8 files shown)


flair/trainers/sequence_tagger_trainer.py
```diff
@@ -175,38 +171,46 @@ def evaluate(self, evaluation: List[Sentence], out_path=None, evaluation_method:
...
+                    # get the predicted tag
+                    predicted_tag = self.model.tag_dictionary.get_item_for_index(pred_id)
+                    token.add_tag('predicted', predicted_tag)
+
+                gold_tags = []
+                predicted_tags = []
+
+                # get spans
+                for tag in sentence.get_spans(self.model.tag_type):
+                    gold_tags.append(tag.__str__())
```
tabergma: Instead of `tag.__str__()` you should use `str(tag)`.

tabergma: Could actually be rewritten to:

`gold_tags = [str(tag) for tag in sentence.get_spans(self.model.tag_type)]`

flair/data.py
```diff
@@ -368,6 +420,44 @@ def convert_tag_scheme(self, tag_type: str = 'ner', target_scheme: str = 'iob'):
...
+                    last_token.whitespace_after = False
+
+                if token.text.startswith('\''):
+                    last_token.whitespace_after = False
+
+            if token.text in ['(']:
+                token.whitespace_after = False
+
+            last_token = token
+        return self
```
tabergma: The return value can be removed as it is not used (also from the docs).
Also the method name should not start with `_` as it is used outside the sentence class.

tabergma:
:+1:

PR #125, opened 2018-09-27
Gh 23 memory embeddings

alanakbik:
- SequenceTagger now predicts tags + confidence score
- Span predictions now also have confidence score
- save model is split into save_final_model and checkpoint for more control over training
- many refactorings to lower CPU cost and increase legibility
- add experimental "memory" embeddings


tabergma:
:+1:

Issue #137, opened 2018-10-10
Memory management for training on large data sets


alanakbik:
In uses cases where training data sets are large or there is little available RAM, language model embeddings cannot be stored in memory (see #135 ). 

**Current solution**: Currently, the only way to still train a model in such cases is to set the 
`embeddings_in_memory` tag to `False` in the trainer classes (`TextClassifierTrainer` or `SequenceTaggerTrainer`). With this flag, embeddings get generated on the fly at each epoch and immediately discarded after use, which solves the memory issue but is computationally expensive since there is no re-use of already computed embeddings. 

**Idea**: Use a key-value store to persist embeddings to disk and enable quick lookup of already computed embeddings. A nice side-effect is that if we run several experiments on the same dataset, embeddings from earlier runs can be re-used, thus speeding up parameter-sweep experiments.

alanakbik:
will be part of release-0.3 and activated by default for `CharLMEmbeddings` (can still be turned off to save disk space)

PR #142, opened 2018-10-11
GH-137: memory optimizations

alanakbik:
This PR adds a method for writing embeddings computed with `CharLMEmbeddings` to disk for fast reuse in and across experiments, therefore: 
- If a model is trained using CharLMEmbeddings is run over the same dataset several times (i.e. for experimenting with parameters), embeddings are only computed on the first run and then reused, thus potentially saving a lot of time.
- If embeddings are too large to fit into memory, you can set `embeddings_in_memory=False` as parameter in the `.train()` function and still reuse computed embeddings, thus again potentially saving a lot of time.

### DIFF
--- a/flair/data.py
+++ b/flair/data.py
@@ -169,10 +169,10 @@ class Token:
         return self.sentence.get_token(self.head_id)
 
     def __str__(self) -> str:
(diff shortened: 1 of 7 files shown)


tabergma:
:+1:

PR #153, opened 2018-10-18
GH-151: Reset cache settings when saving model

### LINKED ISSUE #151: Set `use_cache` to False when saving a model
When training a model using embeddings that are using the cache options (embeddings are written to disk during training to speed up training time) the cache option is saved along the model. When loading the model from disk again on a different machine to predict some data points, the cache files might not be there. This results in an error loading the model. 

Solution: Disable cache option once the model is saved.

tabergma:
closes #151

### DIFF
--- a/flair/embeddings.py
+++ b/flair/embeddings.py
@@ -452,6 +452,8 @@ class CharLMEmbeddings(TokenEmbeddings):
         state = self.__dict__.copy()
         # Remove the unpicklable entries.
         state['cache'] = None
+        state['use_cache'] = False
+        state['cache_directory'] = None
         return state
 
     @property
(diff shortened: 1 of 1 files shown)


flair/embeddings.py
```diff
@@ -460,8 +462,11 @@ def embedding_length(self) -> int:
 
     def _add_embeddings_internal(self, sentences: List[Sentence]) -> List[Sentence]:
 
+        cache_path = '{}-tmp-cache.sqllite'.format(self.name) if self.cache_directory is None else os.path.join(
+            self.cache_directory, '{}-tmp-cache.sqllite'.format(os.path.basename(self.name)))
```
alanakbik: This fails if self.cache_directory is not set.

tabergma: It should not as `self.cache_directory` is set to `None` by default and we are checking in this line if `self.cache_directory is None` or not.

PR #160, opened 2018-10-18
GH-158: remove unneeded AWS and Jinja2 dependencies

### LINKED ISSUE #158: Flair depends on AWS CLI (boto library etc) --- should be removed/optional
I assume that the usual Flair user does not need any AWS dependency, therefore the hard AWS CLI dependency (defined in `setup.py`) should be removed or at least made optional.

hjacobs:
Fixed #158.

Both the AWS and Jinja dependencies are not really used (as far as my grep search and running tests tell me). Both dependencies pull a lot of other stuff (esp. the `awscli` dependency) and cause only headaches without providing value.


NilsRethmeier:
This (the hardset awscli version) also cause issues with facebook visdom installation, so thank you @hjacobs

PR #161, opened 2018-10-18
WIP: how to run tests with Pipenv

### LINKED ISSUE #159: Document how to run tests locally (py.test)
The local setup (virtualenv / pipenv) and how to run tests (py.test) should be clearly documented. 

Additionally I would propose to change the default config for tests to exclude the external files (downloads from S3). Currently the env var `TRAVIS=true` needs to be set in order to exclude external S3 files.

hjacobs:
WORK IN PROGRESS

Fixes #159.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -104,3 +104,5 @@ venv.bak/
 
 # mypy
 .mypy_cache/
+
+Pipfile.lock
--- a/Pipfile
+++ b/Pipfile
@@ -1,0 +1,14 @@
+[[source]]
+url = "https://pypi.org/simple"
+verify_ssl = true
+name = "pypi"
+
+[packages]
+flair = {editable = true, path = "."}
+
+[dev-packages]
+pytest = "*"
+"flake8" = "*"
+
+[requires]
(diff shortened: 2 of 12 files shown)


stefan-it:
:+1: for introducing `pathlib` :) 

I think it would also be a good idea to replace the `.format()`  constructs with f-strings? They're available with Python 3.6 (and >= 3.6 is recommended by `flair`) :)

hjacobs:
@stefan-it yes, I totally agree and also told @alanakbik about both `pathlib` and f-strings. I just did not want to change too many things at once (even `pathlib` is only introduced for now where needed to make the change work).

hjacobs:
I also propose getting rid of the `TRAVIS` env var to skip tests and instead have some flag to run all tests ("slow tests" or similar).

tabergma:
:+1: 
Thanks for the PR! I created two issues out of your suggestions:
https://github.com/zalandoresearch/flair/issues/177
https://github.com/zalandoresearch/flair/issues/176
Feel free to contribute to them :)

Issue #75, opened 2018-08-17
implement pure Python "span-F1" evaluation routine applicable to BIO/BIOS


miracle24:
Hello. Thanks for your wonderful code. I find you also calculate the F1 score using the perl script. I think there is a problem if this script is used to calculate the F1 score for the BIOS tag scheme, since this perl script cannot deal with those tag start with 'S-'. So I wonder if you guys correct the script code or you convert the tag scheme of tokens before using the script? If I am not right, please let me know. Thanks again.

alanakbik:
Hi Miracle, many thanks for spotting this - the script indeed does not have any special provisions for the S-tags, but seems to default to chunkstart=True and chunkend=True for any non-BIE entity tag which is good. I've ran some tests and found some rare cases where it fails to work as intended, so this will probably have a minor effect on evaluation numbers and means that I have to take a closer look. I think for the upcoming release we might end up just implementing our own evaluation routine in Python, which will make a lot of this easier!

tabergma:
in release-0.3

PR #164, opened 2018-10-19
Added class-based metrics

fsonntag:
With the changes in GH-75, the use of the CoNLL evaluation script was removed.
Nevertheless I found it very useful for evaluation class-based predictions, and GH-75 didn't replace this functionality.
So I added this functionality into the `Metrics` object.

Please feel free to suggest any changes and improvements.

### DIFF
--- a/flair/trainers/sequence_tagger_trainer.py
+++ b/flair/trainers/sequence_tagger_trainer.py
@@ -146,11 +146,11 @@ class SequenceTaggerTrainer:
                 log.info("EPOCH {0}: lr {1:.4f} - bad epochs {2}".format(epoch + 1, learning_rate, bad_epochs))
                 if not train_with_dev:
                     log.info("{0:<4}: f-score {1:.4f} - acc {2:.4f} - tp {3} - fp {4} - fn {5} - tn {6}".format(
-                        'DEV', dev_metric.f_score(), dev_metric.accuracy(), dev_metric._tp, dev_metric._fp,
-                        dev_metric._fn, dev_metric._tn))
+                        'DEV', dev_metric.f_score(), dev_metric.accuracy(), dev_metric.get_tp(), dev_metric.get_fp(),
+                        dev_metric.get_fn(), dev_metric.get_tn()))
                 log.info("{0:<4}: f-score {1:.4f} - acc {2:.4f} - tp {3} - fp {4} - fn {5} - tn {6}".format(
-                    'TEST', test_metric.f_score(), test_metric.accuracy(), test_metric._tp, test_metric._fp,
-                    test_metric._fn, test_metric._tn))
+                    'TEST', test_metric.f_score(), test_metric.accuracy(), test_metric.get_tp(), test_metric.get_fp(),
+                    test_metric.get_fn(), test_metric.get_tn()))
 
                 with open(loss_txt, 'a') as f:
                     dev_metric_str = dev_metric.to_tsv() if dev_metric is not None else Metric.to_empty_tsv()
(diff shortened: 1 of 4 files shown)


tabergma:
Great idea! 
Unfortunately, your changes are breaking `Plotter.plot_training_curves()` as the tsv format changed. So we cannot merge you changes yet.
We could also reuse the adapted metric to calculate the class metrics in the text classifier. And the changes should also allow us to add the calculation of the micro and macro average score to the metric class itself. I'll take a close look in the next couple of days.

fsonntag:
Thanks for the feedback!
I've reverted the `to_tsv` function, as you seem to use it for plotting the process and I think it makes sense to only plot the overall metrics.
I've adjusted the `TextClassifierTrainer` and removed the `calculate_class_metrics` and the `calculate_micro_avg_metric` as they seem to be superfluous now. 
So I'm not sure if this all goes with how you have the overall architecture in mind, so please let me know if you have something else in mind!

flair/training_utils.py
```diff
@@ -15,64 +17,94 @@ class Metric(object):
...
     def to_empty_tsv():
         return '_\t_\t_\t_\t_\t_\t_\t_'
 
     def __str__(self):
-        return '{0:<10}\ttp: {1} - fp: {2} - fn: {3} - tn: {4} - precision: {5:.4f} - recall: {6:.4f} - accuracy: {7:.4f} - f1-score: {8:.4f}'.format(
-            self.name, self._tp, self._fp, self._fn, self._tn, self.precision(), self.recall(), self.accuracy(), self.f_score())
+        all_classes = self.get_classes()
+        all_lines = [
+            '{0:<10}\ttp: {1} - fp: {2} - fn: {3} - tn: {4} - precision: {5:.4f} - recall: {6:.4f} - accuracy: {7:.4f} - f1-score: {8:.4f}'.format(
+                MICRO_AVG_METRIC if cls == None else cls,
```
tabergma: I would take `self.name` if `cls is None` as the metric can be used in many different ways and it does not need to be the micro average of some classes.

flair/training_utils.py
```diff
@@ -15,64 +17,94 @@ class Metric(object):
...
+            for cls in all_classes]
+        return '\n'.join(all_lines)
+
+    def get_classes(self) -> List:
+        all_classes = list(set(itertools.chain(*[list(keys) for keys
+                                                 in [self._tps.keys(), self._fps.keys(), self._tns.keys(),
+                                                     self._fns.keys()]])))
+
+        all_classes.sort(key=lambda x: (x is not None, x))
+        return all_classes
```
tabergma: I would love to see methods to calculate the micro and macro average for all classes. Do you think you can add those? Otherwise I'll do it later on.

flair/trainers/text_classification_trainer.py
```diff
@@ -233,8 +229,7 @@ def evaluate(self, sentences: List[Sentence], eval_class_metrics: bool = False,
             batches = [sentences[x:x + mini_batch_size] for x in
                        range(0, len(sentences), mini_batch_size)]
 
-            y_pred = []
-            y_true = []
+            metric = Metric('')
```
tabergma: A metric should get a meaningful name. I would suggest either `MIRCO_AVG` or the data set type (e.g. `TEST`, `DEV` or `TRAIN`).

flair/trainers/sequence_tagger_trainer.py
```diff
@@ -215,21 +215,27 @@ def evaluate(self, evaluation: List[Sentence], out_path=None, evaluation_method:
...
 
-                for gold in gold_tags:
-                    if gold not in predicted_tags:
+                for tag, gold in gold_tags:
+                    if (tag, gold) not in predicted_tags:
                         metric.fn()
+                        metric.fn(tag)
+                    else:
+                        metric.tn()
+                        metric.tn(tag)
```
tabergma: The metrics per tag are collected but not printed at all. Do you want to add a print of the test metric after training is done?

fsonntag:
Thanks for the feedback! Took me a while to address the issues, but I tackled them all

flair/trainers/sequence_tagger_trainer.py
fsonntag: True, I added appropriate logging

flair/training_utils.py
fsonntag: They're added as methods to `Metric`

flair/trainers/sequence_tagger_trainer.py
```diff
@@ -247,6 +250,16 @@ def evaluate(self, evaluation: List[Sentence], out_path=None, evaluation_method:
             score = metric.f_score()
             return score, metric
 
+    def log_metric(self, metric: Metric, dataset_name: str):
+        log.info("{0:<4}: f-score {1:.4f} - acc {2:.4f} - tp {3} - fp {4} - fn {5} - tn {6}".format(
+            dataset_name, metric.f_score(), metric.accuracy(), metric.get_tp(), metric.get_fp(),
+            metric.get_fn(), metric.get_tn()))
+        for cls in metric.get_classes():
+            log.info("{0:<4}: f-score {1:.4f} - acc {2:.4f} - tp {3} - fp {4} - fn {5} - tn {6}".format(
```
tabergma: As `metric.get_classes()` returns a list of all classes containing `None` this currently fails when training a sequence tagger due to 
```
TypeError: unsupported format string passed to NoneType.__format__
```
I would remove the `None` class from `metric.get_classes()`, so you don't log the results twice. Or you can use `dataset_name` in case of `cls is None` instead of using `cls` as name.

fsonntag: I removed the `None` class from get_classes(), as it seems more consistent! Thanks!

fsonntag:
Did a full rebase on the master branch, everything worked fine!

tabergma:
Great! Thanks a lot for improving our metric class!

alanakbik:
@fsonntag thanks for your help!

fsonntag:
You're welcome, thanks for flair :)

PR #182, opened 2018-11-01
GH-170: Add possibility to load custom embeddings

### LINKED ISSUE #170: Using custom word embeddings
Hi there,

I am experimenting with training a SequenceTagger for NER in french, and I would like to use custom word embeddings for the language.

Is there a simple way to do this, that doesn't involve changing the code of `WordEmbeddings` ?

Cheers, and thank you for the great work !

tabergma:
- Add possibility to load custom embeddings
- Add example data to train the sequence tagger in `train.py`

closes #170

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -106,3 +106,5 @@ venv.bak/
 .mypy_cache/
 
 Pipfile.lock
+
+resources/taggers/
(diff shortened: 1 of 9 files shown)


alanakbik:
Awesome!

PR #183, opened 2018-11-01
GH-177: Update tests

### LINKED ISSUE #177: Get rid of `TRAVIS` environment variable
As proposed in https://github.com/zalandoresearch/flair/pull/161, we should get rid of the `TRAVIS` env var to skip tests and instead have some flag to run all tests ("slow tests" or similar). The default unit tests should be self-contained and fast.

tabergma:
- Get rid of `TRAVIS` env
- Per default only "fast" tests are executed when running `pytest`
- To run all test you have to execute `pytest --runslow`
- Update all tests to use pathlib.

closes #177


tests/test_model_integration.py
```diff
@@ -148,9 +155,20 @@ def test_load_use_serialized_tagger():
...
-    corpus = NLPTaskDataFetcher.fetch_data(NLPTask.IMDB)
+    loaded_model: SequenceTagger = SequenceTagger.load('pos')
+
+    loaded_model.predict(sentence)
+    loaded_model.predict([sentence, sentence_empty])
+    loaded_model.predict([sentence_empty])
+
+
+@pytest.mark.slow
+def test_train_load_use_classifier(results_base_path, tasks_base_path):
```
alanakbik: Do the integration tests really need to be skipped on travis? They only make small models and are important for catching lots of errors.

tabergma:
Update: Travis runs unit and integration test on a pull request. On a normal push just the unit tests are executed. Slow tests (such as `test_embeddings.py`) are not executed at all on Travis.

To run only unit tests: `pytest`
To run unit + integration tests: `pytest --runintegration`
To run unit + slow tests: `pytest --runslow`

alanakbik:
Great work, thanks!

PR #186, opened 2018-11-05
Correct suppression of torch warnings

hcl14:
Suppress torch warnings according to [warnings manual](https://docs.python.org/3/library/warnings.html#temporarily-suppressing-warnings) to resolve the issue https://github.com/zalandoresearch/flair/issues/82


alanakbik:
Thanks for your help!

PR #195, opened 2018-11-07
GH-175: Disable autograd in Sequence Tagger & Language Model

### LINKED ISSUE #175: Disable autograd for prediction / evaluation speedups
As discussed in #167, Torch offers a way to disable autograd computation which is [recommended](https://discuss.pytorch.org/t/model-eval-vs-with-torch-no-grad/19615/6) during evaluation / prediction. 

**Task**: Work in `with torch.no_grad():` in the prediction routines and benchmark the difference in speed / memory for prediction / evaluation tasks.

tabergma:
- Disable autograd in Sequence Tagger & Language Model
- Update logging in Sequence Tagger: Only log class metrics at the end of training instead of every epoch.

Performance improvements in Sequence Tagger: up to 10% faster for prediction

closes #175

### DIFF
--- a/flair/models/sequence_tagger_model.py
+++ b/flair/models/sequence_tagger_model.py
@@ -154,12 +154,12 @@ class SequenceTagger(torch.nn.Module):
 
     @classmethod
     def load_from_file(cls, model_file):
-        # suppress torch warnings: 
+        # suppress torch warnings:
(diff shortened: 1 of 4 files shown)


alanakbik:
Great, thanks!

PR #214, opened 2018-11-15
GH-210: Add torch.no_grad to generate_text

### LINKED ISSUE #210: Add torch.no_grad() to LanguageModel.generate_text()
The autograd engine is not required when using an LM to generate text.

So, as pointed out in #167, `torch.no_grad()` needs to be added to `LanguageModel.generate_text()` for better performance and to avoid out of memory issues.

tabergma:
closes #210


alanakbik:
Great!

PR #215, opened 2018-11-15
GH-200: Remember best score of language model

### LINKED ISSUE #200: Loading a saved LM model forgets previous best validation score
When i load a previous trained LM model, and run few more epochs even if my model gets worst score then before it will update best-lm.pt

as code suggests, as trainig start  Best val score is set to 10000 by default even if my previous score was far less than that.

tabergma:
closes #200


alanakbik:
Awesome!

Issue #226, opened 2018-11-19
Non-whitespaced tokenized text results into an infinite loop


jfilter:
I though my text was whitespace-tokenized, but it actually wasn't. I run into an infinite loop here:

https://github.com/zalandoresearch/flair/blob/86485ec920e30b787c43b6b3ff094fbc4fa0c253/flair/data.py#L539-L541

A safeguard would have been helpful and I am about to send a PR to add one.

alanakbik:
Hi @jfilter cool thanks for pointing this out!

I am trying to reproduce the error with a string that is not whitespace tokenized but not getting this error. Could you post a code snippet that causes the error?

jfilter:
Okay, I investigated it a little bit more. I only happens in certain constellations. For instance here:

```python
from flair.models import SequenceTagger
from flair.data import Sentence

tagger = SequenceTagger.load('ner-ontonotes')

text = 'It all boils down to a simple question:    If you are happy with the globalism that has infected our nation on countless levels with New World Order establishment favorites like HW Bush, Bill, W Bush and Obama the past 24 years and want to further erode America, choose Hillary, Ted Cruz or John Kasich who are all more of the same controlled puppets of the elite banks and world power players who only care about making more money and gaining more power for themselves. '

tagger.predict(Sentence(text))[0].to_dict(tag_type='ner')
```

The four spaces after `to a simple question:` break it. If you remove one space, it works fine.

alanakbik:
Wow very interesting. It seems the error only occurs in the predict method - I'll take a closer look! Thanks for spotting this!!

alanakbik:
Thanks again for spotting this. The minimum example to reproduce the error is: 

```python
from flair.data import Sentence

text = ':    nation on '
sentence = Sentence(text)
print(sentence.to_original_text())
```

The error already happens as you noted already during creating the Sentence without a tokenizer. While PR #227 would fix the error, it would prevent people from using the class with text where multiple whitespaces occur (which some might want to do). 

There's a solution in which the indexes get correctly determined even with multiple whitespaces, namely in the Sentence constructor, instead of: 


```python
                for word in text.split(' '):
                    if word:
                        try:
                            word_offset = text.index(word, offset)
                        except:
                            word_offset = offset

                        token = Token(word, start_position=word_offset)
                        self.add_token(token)
                        offset += len(word) + 1
```

use this to determine offsets: 

```python
# add each word in tokenized string as Token object to Sentence
                word = ''
[...]

PR #227, opened 2018-11-19
GH-226: add safeguard for non-tokenized input

jfilter:
raise a ValueError if the input text is not whitespace-tokenized and no tokenizer is used.

See #226.

There is [one failing test](https://github.com/zalandoresearch/flair/blob/master/tests/test_data.py#L96) but I think it should fail. You would as well run into an infinite loop further done. Should the test be adapted to fail?

### DIFF
--- a/flair/data.py
+++ b/flair/data.py
@@ -321,6 +321,10 @@ class Sentence:
             # otherwise assumes whitespace tokenized text
             else:
                 # add each word in tokenized string as Token object to Sentence
+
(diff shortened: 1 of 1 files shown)


tabergma:
Thanks for your contribution and pointing out this issue!
We fixed the root cause of the problem in https://github.com/zalandoresearch/flair/pull/236, thus I will close this PR. It will be merged into master with release-0.4.

Issue #189, opened 2018-11-06
Unified model trainer


alanakbik:
We currently use different trainer classes for sequence taggers and text classifiers (`SequenceTaggerTrainer` and `TextClassifierTrainer`), but these classes have very similar code since they handle the same training-related steps (annealing, evaluation, logging etc.). This leads to lots of redundant code, difficulties in keeping both trainers aligned and been a source of lots of bugs. Also this creates complexity that is not really needed. 

Solution: 
- Create a singular `ModelTrainer` class that can take as argument either a `SequenceTagger` or a `TextClassifier`
- In the process, unify the interfaces of `SequenceTagger` and `TextClassifier` - both models probably need to implement a `predict()` and a `forward_and_loss()` method.
- Make the `evaluate()` function of the `ModelTrainer` static to make it easier to evaluate a trained model

PR #235, opened 2018-11-22
GH-189: Update single trainer

tabergma:
- `EvaluationMetric` needs to be defined when starting training
- Update `Metric` class: Class name needs to be provided when adding a `tp`, `fp`, `tn`, or `fn`
- Removed old trainer classes

related to #189


alanakbik:
Thanks!

Issue #228, opened 2018-11-19
Add better learning rate annealing scheduler + Learning rate finder


mhham:
Two ideas that could allow improving training :

1. Add learning rate finder : as explained in [here](https://www.jeremyjordan.me/nn-learning-rate/), and implemented [here](https://github.com/fastai/fastai/blob/98588ff0d088041259ee77ca63cdf0cc117c3bf1/fastai/train.py#L22)

2. Cosine Annealing with restart : https://github.com/roveo/pytorch/pull/1/commits/35891c46dbdec46e55a5fb725418c2880f631dac

alanakbik:
That sounds like a great idea - if fact we've been recently thinking a lot about better methods to find good parameters than a full parameter sweep. Would you be interested in testing/integrating such approaches?

alanakbik:
@kashif what do you think?

Issue #220, opened 2018-11-16
allow custom optimizers


kashif:
Allow the trainer to use other optimizers rather than the default SGD

PR #240, opened 2018-11-23
added AdamW and SGDW and corresponding schedular

kashif:
for issue #220 and #228

### DIFF
--- a/flair/trainers/language_model_trainer.py
+++ b/flair/trainers/language_model_trainer.py
@@ -11,6 +11,7 @@ from torch.optim.lr_scheduler import ReduceLROnPlateau
 
 from flair.data import Dictionary
 from flair.models import LanguageModel
+from flair.optim import ReduceLRWDOnPlateau
 
 
 log = logging.getLogger(__name__)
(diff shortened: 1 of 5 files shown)


alanakbik:
Awesome!

flair/trainers/language_model_trainer.py
```diff
@@ -186,9 +191,15 @@ def train(self,
 
             epoch = 0
             best_val_loss = self.model.best_score if self.model.best_score is not None else 100000000
-            optimizer = torch.optim.SGD(self.model.parameters(), lr=learning_rate)
-            scheduler: ReduceLROnPlateau = ReduceLROnPlateau(optimizer, verbose=True, factor=anneal_factor,
-                                                             patience=patience)
+            optimizer = self.optimzer(self.model.parameters(), lr=learning_rate, **kwargs)
+            if optimizer.__class__.__name__ in ['AdamW', 'SGDW']:
```
tabergma: Due to consistency I would prefer `type(optimizer) is AdamW or type(optimizer) is SGDW`. Same for the other occurrences.

kashif:
🙇

flair/trainers/language_model_trainer.py
kashif: Can I do:

```python
if isinstance(optimizer, (AdamW, SGDW)):
    ...
```

Issue #243, opened 2018-11-26
NLP Dataset Downloader


alanakbik:
Some NLP datasets, such as the Universal Dependencies corpora, are freely available online. To make experimentation easier, add a feature for automatically fetching these datasets and putting them in a default folder structure. 

Perhaps the existing NLPTaskDataFetcher can be extended such that: 

```python
corpus = NLPTaskDataFetcher.fetch_data(NLPTask.UD_ENGLISH)
```

When this is called, checks first if the dataset is already there. If not, triggers a download action in which the corresponding UD corpus gets downloaded and upacked into default folder structure.

PR #247, opened 2018-11-26
GH-243: dataset downloader

alanakbik:
addresses #243 

- DataFetcher can now download universal dependencies corpora for 30 languages, WikiNER corpora for 8 languages and some CoNLL tasks. 
- If there is only a training data file, now samples both dev and test data (#3)
- rename all fetch_* methods to load_* methods

So now you can load a dataset like this: 

```python
from flair.data_fetcher import NLPTaskDataFetcher, NLPTask

# load one corpus
corpus = NLPTaskDataFetcher.load_corpus(NLPTask.UD_ENGLISH)
print(corpus)

# load a MultiCorpus of two UD corpora
corpus = NLPTaskDataFetcher.load_corpora([NLPTask.UD_ENGLISH, NLPTask.UD_GERMAN])
print(corpus)
```

You no longer need to download the UD corpus yourself. The method will check if it is there and if not download the corpus.


tabergma:
:+1: Great!

PR #252, opened 2018-11-27
fix error in setup.py

kashif:


alanakbik:
Thanks for spotting this!

Issue #248, opened 2018-11-26
RuntimeError: CUDA error: out of memory


omerarshad:
when training on dataset of emerging entitiy recognition, memory error occurs, even when using small batch size

alanakbik:
Hi @omerarshad that's strange. Could you give us more info: How much Cuda memory do you have? What embeddings are you using? Where exactly do you get the oom error?

omerarshad:
I have 12 Gb memory, i am using crawl, mix english forward, mix eng backward embeddings to train sequence model on W-NUT emerging entities. I am using example provided in docs.

here is the error : 

![error](https://user-images.githubusercontent.com/16164105/49067329-116e8400-f224-11e8-8536-0b997fdc0585.png)

alanakbik:
could you try setting embeddings_in_memory to False in the trainer? Does it run through then?

omerarshad:
i tried tha but same error persists

tabergma:
Can you please share the following things with us:

- Version: Which flair version are you using?
- Environment: What kind of operation system are you using?
- Code: Can you please share the exact code you are using with us, so that we might be able to reproduce the problem.
- What did you already tried to solve the issue?
- When does the error occur exactly?

The more details you share the easier it is to actually solve the problem :). Thanks.

omerarshad:
Flair version : 0.3.2
Linux, ubuntu 

Code : 

from flair.embeddings import CharLMEmbeddings
from flair.data import Sentence

glove_embedding = WordEmbeddings('crawl')

from flair.data_fetcher import NLPTaskDataFetcher, NLPTask
from flair.data_fetcher import TaggedCorpus

tag_dictionary = corpus.make_tag_dictionary(tag_type='ner')

from flair.models import SequenceTagger

tagger: SequenceTagger = SequenceTagger(hidden_size=256,
                                        embeddings=glove_embedding,
                                        tag_dictionary=tag_dictionary,
                                        tag_type='ner',
                                        use_crf=True)

# 6. initialize trainer
from flair.trainers import SequenceTaggerTrainer

trainer: SequenceTaggerTrainer = SequenceTaggerTrainer(tagger, corpus)

# 7. start training
trainer.train('resources/taggers/example-ner',
              learning_rate=0.1,
              mini_batch_size=10,
              embeddings_in_memory = False,
              
              max_epochs=150)

omerarshad:
Error occurs after few epochs 
[...]

Issue #199, opened 2018-11-08
Emerging Entity Detection


ZhaofengWu:
The readme mentions that this framework achieves 50.20 F1 on WNUT-17 dataset for Emerging Entity Detection, but is there any way I can reproduce this result? I looked over the [How to Reproduce Experiments](https://github.com/zalandoresearch/flair/blob/master/resources/docs/EXPERIMENTS.md) document but didn't see this task mentioned.

Or is there a way I can train some model for this task using this framework?

Thanks!

alanakbik:
Hi @ZhaofengWu - I think we just used the standard parameters to train this model, so we never performed any parameter selection. I would probably start with a configuration in which you use the 'crawl' embeddings for English, paired with mixed language models forward and backward. Like this:

```python
embedding_list = [
    WordEmbeddings('crawl'),
    CharLMEmbeddings('mix-forward'),
    CharLMEmbeddings('mix-backward'),
]

embeddings: StackedEmbeddings = StackedEmbeddings(embeddings=embedding_list)

# initialize sequence tagger
from flair.models import SequenceTagger

tagger: SequenceTagger = SequenceTagger(hidden_size=256,
                                        embeddings=embeddings,
                                        tag_dictionary=tag_dictionary,
                                        tag_type=tag_type)

# initialize trainer
trainer: ModelTrainer = ModelTrainer(tagger, corpus)

trainer.train('resources/taggers/wnut',
              learning_rate=0.1,
              mini_batch_size=32,
              train_with_dev=True,
              patience=3)
```

Then, you could try experimenting with learning rates and mini batch sizes to improve the results. Remember to do parameter selection against dev data, to train_with_dev then should be set to False.

ZhaofengWu:
Thanks! What should I pass in for the `corpus` argument? For crawl, I see that you have
```python
# 1. get the corpus
corpus: TaggedCorpus = NLPTaskDataFetcher.fetch_data(NLPTask.ONTONER)
print(corpus)

# 2. what tag do we want to predict?
tag_type = 'ner'

# 3. make the tag dictionary from the corpus
tag_dictionary = corpus.make_tag_dictionary(tag_type=tag_type)
```
But I'm not sure what I should do for W-NUT.

Also, what's the difference between `SequenceTaggerTrainer` and `ModelTrainer`? Why are we using the latter here?

alanakbik:
Ah my mistake, not the `ModelTrainer`, you should use `SequenceTaggerTrainer`. 

[...]

PR #253, opened 2018-11-27
GH-243: add WNUT-17 dataset to downloader

alanakbik:
addresses #243 

Since there is a lot of interest in the Flair community (#199 #248), added one more dataset to the downloader, namely the WNUT-17 shared task on emerging entity detection. 

Just load with: 

```python
corpus = NLPTaskDataFetcher.load_corpus(NLPTask.WNUT_17)
```

Dataset will get downloaded automatically.


tabergma:
:+1:

PR #254, opened 2018-11-27
GH-176: Use pathlib in entire project

### LINKED ISSUE #176: Use `pathlib` in entire project
As suggested in https://github.com/zalandoresearch/flair/pull/161 make use `pathlib` in entire flair project.

tabergma:
closes #176


flair/data_fetcher.py
```diff
@@ -451,15 +432,15 @@ def read_column_data(path_to_column_file: str,
...
+    def read_conll_ud(path_to_conll_file: Path) -> List[Sentence]:
         """
        Reads a file in CoNLL-U format and produces a list of Sentence with full morphosyntactic annotation
        :param path_to_conll_file: the path to the conll-u file
        :return: list of sentences
        """
         sentences: List[Sentence] = []
 
-        lines: List[str] = open(path_to_conll_file, encoding='utf-8'). \
+        lines: List[str] = open(str(path_to_conll_file), encoding='utf-8'). \
```
stefan-it: The `open` function directly supports `Path` objects :) So the `str` conversion is not needed :)

tabergma: Thanks!

alanakbik:
Great!

Issue #249, opened 2018-11-26
cache directory always resets to None


ericlief:
I am running out of memory during sequence tagging and found out no cache directory existed, despite my passing it as follows:

 clm_fw = CharLMEmbeddings("/home/liefe/lm/fw_p25/best-lm.pt", use_cache=True, cache_directory="/home/liefe/tag/cache")  
 clm_bw = CharLMEmbeddings("/home/liefe/lm/bw_p25/best-lm.pt", use_cache=True, cache_directory="/home/liefe/tag/cache")     

What looks like is happening when I inspected the _add_embeddings_internal method, is that it will always be set to None because the path is to a file that has not yet been created. When I fixed this, it now saves to the path.
def _add_embeddings_internal(self, sentences: List[Sentence]) -> List[Sentence]:
        
        # this whole block is for compatibility with older serialized models  TODO: remove in version 0.4
        if 'cache' not in self.__dict__ or 'cache_directory' not in self.__dict__:
            self.use_cache = False
            self.cache_directory = None
        else:
            cache_path = '{}-tmp-cache.sqllite'.format(self.name) if not self.cache_directory else os.path.join(
                self.cache_directory, '{}-tmp-cache.sqllite'.format(os.path.basename(self.name)))
           
            if not os.path.exists(cache_path):
                self.use_cache = False
                self.cache_directory = None

alanakbik:
Ah yes thanks for spotting this! I suspect some changes for 0.3.2 caused this bug. 

We'll include a fix for this in the upcoming 4.0 release!

PR #256, opened 2018-11-27
GH-249: refactor caching in CharLMEmbeddings

alanakbik:
addresses #249  

- Refactor CharLMEmbeddings caching mechanism to address bug #249 
- Relax some PathLib signatures to also take string


tabergma:
:+1:

Issue #257, opened 2018-11-27
Learning rate output during training of language model


stefan-it:
Hi,

after the first epoch (during language model training) the learning rate output is often:

```bash
| learning rate 0.00
```

Can I open a PR to display at least four digits (like it is displayed during sequence tagging)?

alanakbik:
Thanks for adding this! #258

PR #258, opened 2018-11-27
GH-257: adjust learning rate output information 

stefan-it:
Hi,

this PR adjusts the learning rate output information during language model training (now four digits are displayed for learning rate in output).


alanakbik:
Great, thanks!

PR #263, opened 2018-11-28
GH-217: Resume training

### LINKED ISSUE #217: Resume training (keeping epoch, losses, and best model score)
It would be nice to be able to keep the epoch numbers and losses when resuming training on a model loaded from disk.

(https://pytorch.org/tutorials/beginner/saving_loading_models.html#saving-loading-a-general-checkpoint-for-inference-and-or-resuming-training)

#200 already mentioned keeping the scores in the pickled .pt object, but keeping the logs is also a nice add in (not overwriting them).

tabergma:
closes #217 

To be able to resume training you need to train with `checkpoint` set to `True`. 
If `checkpoint=True` the model plus the training parameters (e.g. loss, epoch, optimizer state dict) are saved to disk after every epoch. 
To continue training, for example for a `SequenceTagger`, you need to execute:
```
trainer = ModelTrainer.load_from_checkpoint('checkpoint.pt', 'SequenceTagger', corpus)
trainer.train(out_path)
```

### DIFF
--- a/flair/models/language_model.py
+++ b/flair/models/language_model.py
@@ -5,6 +5,9 @@ import torch
 import math
 from torch.autograd import Variable
 from typing import List
+
+from torch.optim import Optimizer
+
 from flair.data import Dictionary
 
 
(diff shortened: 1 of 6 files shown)


alanakbik:
:+1:

Issue #242, opened 2018-11-26
Wrapper for hyperopt


tabergma:
Finding the optimal parameters for a model is hard. To help you to find the best parameter combination we want to include a wrapper for [hyperopt](https://github.com/hyperopt/hyperopt) so that you can find the correct parameters for your `SequenceTagger` and `TextClassifier` faster.

PR #264, opened 2018-11-28
GH-242: Take average over training runs as score

tabergma:
related to #242 

To overcome the issues of random noise in the training, we average the last three scores of the training. Additionally, you can run one parameter combination x times and take the average over the x runs as score.


alanakbik:
:+1:

PR #268, opened 2018-11-28
GH-212: Write training log to file in experiment directory

### LINKED ISSUE #212: Saving the training log in the experiment directory
It would be nice that the training log is also automatically dumped into the experiment directory.

tabergma:
closes #212


alanakbik:
Great!

PR #278, opened 2018-12-02
improve find_learning_rate API

kashif:
- [x] use running mean instead of logging all the losses O(1) instead of O(itr) space
- [x] cast base path to Path if its a string
- [x] takes file name as a parameter which allows comparing different parameter plots
- [x] stop if loss becomes nan
- [x] plot with skipped first few and last few iterations


alanakbik:
Cool, can't wait to try it out :)

PR #281, opened 2018-12-03
GH-280: Fix language model training.

### LINKED ISSUE #280: Language model training
Hi,

with the latest `0.4` branch, training of a language model does not work. More precisely, the `TextCorpus` class expects a `Path`-like object as training path (see [here](https://github.com/zalandoresearch/flair/blob/release-0.4/flair/trainers/language_model_trainer.py#L19)). With a `Path`-like object the following comprehension won't work:

```python
self.train_files = sorted(
    [f for f in self.train_path.iterdir() if (self.train_path / f).exists()])
```

-> `(self.train_path / f).exists()` will always be empty, so the list of training files will be empty. I could temporarily fix it with using only `f.exists()` but then another error appeared.

It just want to mention this issue before the official `0.4` version is released :)

tabergma:
closes #280

### DIFF
--- a/flair/trainers/language_model_trainer.py
+++ b/flair/trainers/language_model_trainer.py
@@ -23,7 +23,7 @@ class TextCorpus(object):
         self.train_path = path / 'train'
 
         self.train_files = sorted(
-            [f for f in self.train_path.iterdir() if (self.train_path / f).exists()])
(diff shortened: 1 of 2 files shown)


alanakbik:
:+1:

PR #283, opened 2018-12-03
GH-282: Load logging config from file.

### LINKED ISSUE #282: Disable logging from dependencies
Currently, the logging configuration is overwritten by the BERT logging set up. Everything is written to stderr, log level is set to DEBUG and no file handler is applied anymore.

To avoid this, we should read our logging configuration from a file and disable existing loggers as described here: https://docs.python.org/3.5/howto/logging.html#configuring-logging

tabergma:
closes #282


alanakbik:
:+1:

PR #284, opened 2018-12-03
GH-276: annealing options

### LINKED ISSUE #276: ModelTrainer annealing options
Currently, the `train_with_dev` parameter in the `ModelTrainer` impacts the annealing behavior: If set to True, we anneal against training loss, otherwise against current score computed on dev data. 

This is a bit problematic since: 

- We may want more control over annealing behavior: For instance, we might want to anneal against train loss even if train_with_dev is set to False.
- This behavior is implicit, while is may be easier to understand if it can be explicitly set 

**Task**: Add more annealing options to `ModelTrainer` to make it possible to explicitly control annealing behavior.

alanakbik:
closes #276 

- add option to `ModelTrainer` to anneal either against test loss or dev score
- during parameter selection, never test on test data
- renamed some variables in `ModelTrainer` for clarity

### DIFF
--- a/flair/embeddings.py
+++ b/flair/embeddings.py
@@ -6,7 +6,6 @@ from typing import List, Union, Dict
 import gensim
 import numpy as np
 import torch
-from torch.utils.data import TensorDataset, DataLoader, SequentialSampler
 from deprecated import deprecated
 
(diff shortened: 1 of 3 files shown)


tabergma:
:+1:

PR #290, opened 2018-12-06
GH-289: Add eval_mini_batch_size to ModelTrainer

### LINKED ISSUE #289: Introduce eval_mini_batch_size
It should be possible to set a separate `mini_batch_size` for evaluation as this does not have any impact on training, but on performance. If no `mini_batch_size` for evaluation is set, we should set it equal to the `mini_batch_size` of training.

tabergma:
closes #289


alanakbik:
:+1:

PR #292, opened 2018-12-07
GH-276: return test score and add twitter embeddings

### LINKED ISSUE #291: Add Twitter Embeddings
We used to have pre-trained Twitter embeddings in earlier releases. We somehow removed them, but they are important for good scores on the WNUT task. So put them back in.

alanakbik:
Two small changes: 

- added back in pre-computed twitter embeddings (closes #291 )
- returns test score in hyperparameter optimization


tabergma:
:+1:

Issue #251, opened 2018-11-27
Add BERT embeddings


kashif:
Would be handy to have Bert embeddings especially for stacked embeddings etc.

alanakbik:
Wow that was fast :) How many BERT layers did you use?

We still need to match BERT tokenization to our token structure. Currently, their tokenizer may split words that we treat as one word: For instance the number "250,000" is split by BERT into three tokens "250",  "," and  "000" which it embeds individually. We currently just take the first embedding as token embedding, but this was just a quick fix that will affect performance especially for languages other than English. Once we fix this, I would expect POS tagging results to improve!

stefan-it:
I used the default parameters of the Bert Embedding class :) But I'm going to test it with more layers + follow this PR for further fixes :)

alanakbik:
First version added to release-0.4 branch. You can call `BertEmbeddings` like any other embeddings: 

```python
from flair.embeddings import BertEmbeddings

# instantiate BERT embeddings
bert_embeddings = BertEmbeddings()

# make example sentence
sentence = Sentence('I love Berlin.', use_tokenizer=True)

# embed sentence
bert_embeddings.embed(sentence)

# print embedded tokens
for token in sentence:
    print(token)
    print(token.embedding)
```

This also makes it possible to mix and match BERT, Flair and ELMo embeddings with the `StackedEmbeddings` class: 

```python
embeddings = StackedEmbeddings(embeddings=[
    WordEmbeddings('glove'),
    BertEmbeddings('bert-large-uncased'),
    FlairEmbeddings('news-forward'),
    FlairEmbeddings('news-backward'),
])
```

You can pass two pooling operations to BERT: 'first' and 'mean'. This affects how token embeddings are build from word pieces (either use embedding of first word piece or average over all pieces). You can also choose which layers to concat to form embeddings - it is currently set to the last 4 layers, which in the BERT paper performed best for feature-based usage. 

However, very initial experiments do not give us great results with BERT embeddings. This could be because our parameter settings aren't great and more experimentation is required, or we are doing something else wrong. 

Anyway, we'd be happy how they work if somebody's already experimenting with the 0.4 branch!

stefan-it:
[...]

PR #305, opened 2018-12-12
GH-251: Raise ValueError if BertModel does not exist.

tabergma:
related to #251

### DIFF
--- a/flair/embeddings.py
+++ b/flair/embeddings.py
@@ -9,7 +9,7 @@ import torch
 from deprecated import deprecated
 
 from pytorch_pretrained_bert.tokenization import BertTokenizer
-from pytorch_pretrained_bert.modeling import BertModel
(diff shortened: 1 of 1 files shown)


flair/embeddings.py
```diff
@@ -734,6 +734,10 @@ def __init__(self,
         """
         super().__init__()
 
+        if bert_model not in ['bert-base-uncased', 'bert-large-uncased', 'bert-base-cased', 'bert-large-cased',
+                              'bert-base-multilingual-cased', 'bert-base-multilingual-uncased', 'bert-base-chinese']:
+            raise ValueError('Provided bert-model is not available.')
```
stefan-it: I think it would be better to use the internal `PRETRAINED_MODEL_ARCHIVE_MAP` variable from the `pytorch_pretrained_bert` module, see [here](https://github.com/huggingface/pytorch-pretrained-BERT/blob/master/pytorch_pretrained_bert/modeling.py#L41) :)

alanakbik:
:+1:

PR #313, opened 2018-12-14
Bump version of pytorch-pretrained-bert

stefan-it:
Hi,

today version 0.4.0 of the `pytorch-pretrained-bert` was released (see [Changelog](pytorch-pretrained-bert)) so it would be great to have the latest version of that module in the upcoming `flair` release.


alanakbik:
Thanks!

PR #314, opened 2018-12-17
GH-51: Add fine-grained POS Tagger for German Tweets

### LINKED ISSUE #51: Model for Fine-grained POS Tagging of German Tweets
Hi,

recently I received the training, dev and test data from the "Fine-grained POS Tagging of German Tweets" paper (which can be found [here](https://pdfs.semanticscholar.org/82c9/90aa15e2e35de8294b4a721785da1ede20d0.pdf)). The highest accuracy mentioned in that paper is 89.42%.

I trained a model with German fasttext embeddings and forward/backward lm embeddings and after 40 epochs (on CPU) an accuracy ~ 92% could be achieved. I double checked that result with an own evaluation script that loads the trained model, predicts the tags and compares the predicted tags with the gold standard and this also achieves 92%.

So I would like to ask if the core team is interested in such a pos tagging model for German twitter data? I would really like to share the model, but then I have to ask the paper author for a permission to share a trained model.

tabergma:
Thanks @stefan-it for providing the model!

closes #51

### DIFF
--- a/flair/models/sequence_tagger_model.py
+++ b/flair/models/sequence_tagger_model.py
@@ -629,6 +629,12 @@ class SequenceTagger(flair.nn.Model):
                                   'de-pos-ud-v0.2.pt'])
             model_file = cached_path(base_path, cache_dir=cache_dir)
 
+        elif model.lower() == 'de-pos-fine-grained':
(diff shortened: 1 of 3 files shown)


alanakbik:
:+1:

PR #317, opened 2018-12-18
Update the pytest command

staeff:
Otherwise pytest could run test in the `lib` folder on all dependencies of the package. That happened to me.
Update: I just realized that I only ran into the issue, because it did do the installation with pip in a toplevel venv. This was neccessary, because the pipenv installation did not work. So this PR might not useful.


tabergma:
:+1: Thanks!

PR #323, opened 2018-12-21
GH-322: generate text

### LINKED ISSUE #322: generate_text() throws error with small temperatures
The 'generate_text()' method of the LanguageModel class throws an overflow error in certain cases, especially when the 'temperature' variable is set to be small. 

**To Reproduce**
```python

language_model = FlairEmbeddings('news-forward').lm

language_model.generate_text(temperature=0.01, number_of_characters=100)
```

alanakbik:
closes #322 

This PR fixes the overflow problem and cleans up the LM code a bit. Also adds a unit test for the issue.

### DIFF
--- a/flair/models/language_model.py
+++ b/flair/models/language_model.py
@@ -60,9 +60,9 @@ class LanguageModel(nn.Module):
 
     def init_weights(self):
         initrange = 0.1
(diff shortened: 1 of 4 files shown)


tabergma:
:+1:

Issue #340, opened 2018-12-30
TypeError: unsupported operand type(s) for /: 'str' and 'str' in language_model_trainer.py


xuxingya:
An error occured when I try to train my own embedding following Tutorial 9. My system is Ubuntu 18 and python3.6.
```
corpus = TextCorpus('/data/resume_extract_file/corpus',
                    dictionary,
                    is_forward_lm,
                    character_level=True)
```
![image](https://user-images.githubusercontent.com/13343428/50547009-9fe38480-0c6c-11e9-8bab-ffcbca7a73d1.png)
It looks like the problem was occurred by the 24 line of flair/flair/trainers/language_model_trainer.py
self.train_path = path / 'train'.

stefan-it:
Should be:

```python
from pathlib import Path

corpus = TextCorpus(Path('/data/resume_extract_file/corpus'),
                    dictionary,
                    is_forward_lm,
                    character_level=True)
```

stefan-it:
I made a PR to correct this in the documentation :)

xuxingya:
Thank you for you fast respond :)

alanakbik:
Ah, great! @xuxingya thanks for spotting this! @stefan-it thanks for fixing this :)

alanakbik:
fixed in #341

PR #341, opened 2018-12-30
GH-340: use Path object for specifying corpus folder

stefan-it:
Hi,

this PR corrects the tutorial for training an own language model. The corpus folder must be a `Path` object (which is passed to the `TextCorpus` instance).


alanakbik:
Thanks!

Issue #349, opened 2019-01-02
Lower default dropout for language models


alanakbik:
Currently, the `LanguageModel` class sets a dropout of 0.5 by default. However, in most cases there is little danger of overfitting, since text corpora used to train LMs tend to be very large. We noted that we get improved results if we reduce the dropout to 0.1 or even no dropout at all (see discussion in #329). 

Task: Set a lower default value for dropout in the `LanguageModel`.

alanakbik:
closed by #350

PR #350, opened 2019-01-02
GH-349: language model optimizations

alanakbik:
Small optimizations to the `LanguageModel` and `LanguageModelTrainer` classes: 

- lowers default dropout of the `LanguageModel` to 0.1 which will be better for most use cases (#349) 
- small improvements to `generate_text()` method: Newline character is no longer added by default and method now additionally returns average log likelihood of generated string. 
- only send mini-batches to GPU for better GPU memory usage


kashif:
👍

PR #353, opened 2019-01-03
Gh 347 sequence tagging enhancement

Nilabhra:
Changes in the predict() function:

- Faster inferencing by presorting sequences. Also, let's the user figure out the optimal batch size early on.
- Token embeddings are cleared after processing each batch, thereby saving a lot of RAM.
- Progress bar that can be toggled by setting the argument `verbose = True`, useful when processing a large number of sequences.

### DIFF
--- a/flair/models/sequence_tagger_model.py
+++ b/flair/models/sequence_tagger_model.py
@@ -17,6 +17,7 @@ from typing import List, Tuple, Union
 
 from flair.training_utils import clear_embeddings
 
+
 log = logging.getLogger('flair')
 
 START_TAG: str = '<START>'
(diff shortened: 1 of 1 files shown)


flair/models/sequence_tagger_model.py
```diff
@@ -252,41 +255,57 @@ def forward_labels_and_loss(self, sentences: Union[List[Sentence], Sentence]) ->
...
-        # first, sort sentences by number of tokens
-        sentences.sort(key=lambda x: len(x), reverse=True)
-        longest_token_sequence_in_batch: int = len(sentences[0])
+        # if grad is enabled, sort sentences by number of tokens
+        if torch.is_grad_enabled():
+            sentences.sort(key=lambda x: len(x), reverse=True)
 
         lengths: List[int] = [len(sentence.tokens) for sentence in sentences]
         tag_list: List = []
+        longest_token_sequence_in_batch: int = len(sentences[0])
```
alanakbik: during eval phase of training, grad is not enabled, causing sentences to not be sorted. the longest sentence is then not always in the first position (sentences[0]). 

this raises an exception during training (eval phase). 

to reproduce error: 

```python
corpus: TaggedCorpus = NLPTaskDataFetcher.load_corpus(NLPTask.UD_ENGLISH).downsample(0.01)

tag_dictionary = corpus.make_tag_dictionary(tag_type='pos')

                                        embeddings=WordEmbeddings('glove'),
                                        tag_type='pos')

trainer.train('resources/taggers/example-ner')

```

Nilabhra: Ah yes, I thought the predict method would be used in the training phase (which now sorts the sequences). I'll see what can be done here. Thanks!

Nilabhra: I made the fix. But, how is that the test's passed both locally and on travis?

alanakbik: Yeah I was wondering about this as well, the test should not have passed. The problem was that the error occurs only if the first sentence is not longest sentence in a mini-batch, which in the tiny datasets we use for integration tests apparently never happens. We need to add an additional test for this.

alanakbik:

PR #361, opened 2019-01-05
Fixes #345. Allows user to specify a field name to use for embedding …

### LINKED ISSUE #345: add column-name-based embeddings
There are other token-level types that can be embedded, e.g. part of speech tags or UMLS medical concepts. Right now we can only use embeddings for the token column in the CoNLL format.

This class roughly implements what I'd like. Not sure whether it should be its own class though because it can also be implemented by adding a few lines to existing classes.

```
class FieldEmbeddings(WordEmbeddings):

    def __init__(self, path : str, field : str):
        super().__init__(path)
        self.field = field

    def _add_embeddings_internal(self, sentences: List[Sentence]) -> List[Sentence]:
        for i, sentence in enumerate(sentences):
            for token, token_idx in zip(sentence.tokens, range(len(sentence.tokens))):
                token: Token = token
                text = token.get_tag(self.field).value
                if text in self.precomputed_word_embeddings:
                    word_embedding = self.precomputed_word_embeddings[text]
                    # self.hits += 1
                elif text.lower() in self.precomputed_word_embeddings:
                    word_embedding = self.precomputed_word_embeddings[text.lower()]
                elif re.sub(r'\d', '#', text.lower()) in self.precomputed_word_embeddings:
                    word_embedding = self.precomputed_word_embeddings[re.sub(r'\d', '#', text.lower())]
                elif re.sub(r'\d', '0', text.lower()) in self.precomputed_word_embeddings:
                    word_embedding = self.precomputed_word_embeddings[re.sub(r'\d', '0', text.lower())]
                else:
                    word_embedding = np.zeros(self.embedding_length, dtype='float')
                word_embedding = torch.FloatTensor(word_embedding)
                token.set_embedding(self.name, word_embedding)
        return sentences
```
And used with:
```
    column_format = {0:'word', 1:'pos', 2:'cui', 3:'ner'}
    embeddings = FieldEmbeddings("resources/embeddings/cui_embeddings.gensim", 'cui')

tmills:
…lookup.

I used None as default because I assume that would be a faster comparison than using the string 'text'.

### DIFF
--- a/flair/embeddings.py
+++ b/flair/embeddings.py
@@ -144,7 +144,7 @@ class StackedEmbeddings(TokenEmbeddings):
 class WordEmbeddings(TokenEmbeddings):
     """Standard static word embeddings, such as GloVe or FastText."""
 
-    def __init__(self, embeddings: str):
+    def __init__(self, embeddings: str, field: str = None):
         """
         Initializes classic word embeddings. Constructor downloads required files if not there.
         :param embeddings: one of: 'glove', 'extvec', 'crawl' or two-letter language code.
@@ -206,6 +206,8 @@ class WordEmbeddings(TokenEmbeddings):
 
         self.precomputed_word_embeddings = gensim.models.KeyedVectors.load(str(embeddings))
 
+        self.field = field
+
         self.__embedding_length: int = self.precomputed_word_embeddings.vector_size
         super().__init__()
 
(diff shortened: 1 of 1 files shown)


tmills:
Sorry, but I can't understand why this test fails, it passes locally for me. Any ideas?

alanakbik:

Issue #363, opened 2019-01-05
Duplicate definition of ELMoEmbeddings


ceh:
There exists two class definitions of `ELMoEmbeddings`, see https://github.com/zalandoresearch/flair/blob/548b630477af8a96403e581d320adfe9464fa30f/flair/embeddings.py#L244 and https://github.com/zalandoresearch/flair/blob/548b630477af8a96403e581d320adfe9464fa30f/flair/embeddings.py#L788

PR #364, opened 2019-01-05
GH-363: Remove duplicate definition of ELMoEmbeddings

ceh:
Proposed fix for https://github.com/zalandoresearch/flair/issues/363


alanakbik:
Ugh - I guess duplication must have happened during a merge. Thanks for spotting and fixing this!

PR #365, opened 2019-01-05
Fix link to contributor guidelines

ceh:
File was moved in https://github.com/zalandoresearch/flair/pull/354


alanakbik:
Great, thanks for spotting and fixing this!

PR #368, opened 2019-01-06
GH-367: sentence embeddings

### LINKED ISSUE #367: Inconsistent shape of Token and Sentence embeddings
Currently, word-level embeddings generated with one of the `TokenEmbeddings` classes are vectors, whereas sentence-level embeddings generated with one of the `DocumentEmbeddings` classes are unsqueezed vectors. This is inconsistent and counterintuitive.

It would be better if sentence-level embeddings were also normal vectors. 

Example code to reproduce: 

```python
word_embeddings = WordEmbeddings('glove')
sentence = Sentence('hello world')
word_embeddings.embed(sentence)
for token in sentence:
    print(token.embedding.size())

document_embeddings = DocumentPoolEmbeddings([word_embeddings])
sentence = Sentence('hello world')
document_embeddings.embed(sentence)
print(sentence.embedding.size())

document_embeddings = DocumentLSTMEmbeddings([word_embeddings], hidden_size=100)
sentence = Sentence('hello world')
document_embeddings.embed(sentence)
print(sentence.embedding.size())
```

This prints: 

```console
Token embeddings:
torch.Size([100])
torch.Size([100])

Document embeddings:
torch.Size([1, 100])
torch.Size([1, 100])
```

Better would be if all four embeddings had the same shape, namely `torch.Size([100])`.

alanakbik:
closes #367 

- `Sentence` embeddings are now vectors, same as `Token`. 
- Unit tests test for embedding length
- Make printing prettier for `Sentence` and `TaggedCorpus`

### DIFF
--- a/flair/data.py
+++ b/flair/data.py
@@ -588,7 +588,12 @@ class Sentence:
         return s
 
     def __str__(self) -> str:
-        return 'Sentence: "{}" - {} Tokens'.format(' '.join([t.text for t in self.tokens]), len(self))
+
+        if len(self.labels) == 0:
+            return f'Sentence: "{self.to_tokenized_string()}" - {len(self)} Tokens'
+        else:
+            return f'Sentence: "{self.to_tokenized_string()}" - {len(self)} Tokens - Labels: {self.labels} '
+
 
     def __len__(self) -> int:
         return len(self.tokens)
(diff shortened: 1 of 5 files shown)


flair/data.py
```diff
@@ -588,7 +588,12 @@ def __copy__(self):
         return s
 
     def __str__(self) -> str:
-        return 'Sentence: "{}" - {} Tokens'.format(' '.join([t.text for t in self.tokens]), len(self))
+
+        if len(self.labels) == 0:
```
bluesheeptoken: Don't you want to avoid calculating length and use implicit boolean conversion? `if not self.labels` ?

flair/data.py
```diff
@@ -588,7 +588,12 @@ def __copy__(self):
...
+            return f'Sentence: "{self.to_tokenized_string()}" - {len(self)} Tokens'
+        else:
+            return f'Sentence: "{self.to_tokenized_string()}" - {len(self)} Tokens - Labels: {self.labels} '
```
bluesheeptoken: Maybe factorize the first part of the string? And add ` - Labels: {self.labels}` only `if self.labels` ?

flair/data.py
tabergma: To simplify the check you could also do:
```
if self.labels:
  return f'Sentence: "{self.to_tokenized_string()}" - {len(self)} Tokens - Labels: {self.labels}'
else:
  return f'Sentence: "{self.to_tokenized_string()}" - {len(self)} Tokens'
```

tabergma:
:+1:

flair/data.py
alanakbik: Thanks - jusu added this!

alanakbik: Thanks - just added this!

flair/data.py
alanakbik: thanks! for readability purposes though I'd prefer an explicit if-else!

tabergma:
:+1:

PR #369, opened 2019-01-07
Use is to compare None

kinow:
Hi,

Found one comparison of `None` done with `==` while it should be `is` instead.

Cheers, and thanks for `flair`!
Bruno


tabergma:
:+1: Thanks!

Issue #325, opened 2018-12-21
torch errors when I try Bert embeddings


mvss80:
I'm on Ubuntu 18.04 with Intel® Core™ i5-8400 CPU @ 2.80GHz × 6  and GeForce GTX 1080/PCIe/SSE2.
I have pytorch 1.0 and installed flair. When I try the following code snippet
```
from flair.embeddings import BertEmbeddings, FlairEmbeddings
from flair.data import Sentence

BertEmbedding = BertEmbeddings()
sent = Sentence('I love Austin')
BertEmbedding.embed(sent)

# fe = FlairEmbeddings('news-forward')
# fe.embed(sent)
```
I get this error. My pytorch installation seems to be fine. Any ideas on how to fix this?
```
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
<ipython-input-1-47fbfcb3fd4f> in <module>
      5 BertEmbedding = BertEmbeddings()
      6 sent = Sentence('I love Austin')
----> 7 BertEmbedding.embed(sent)
      8 # fe.embed(sent)

~/anaconda3/envs/gpu-dev/lib/python3.6/site-packages/flair/embeddings.py in embed(self, sentences)
     54 
     55         if not everything_embedded or not self.static_embeddings:
---> 56             self._add_embeddings_internal(sentences)
     57 
     58         return sentences

~/anaconda3/envs/gpu-dev/lib/python3.6/site-packages/flair/embeddings.py in _add_embeddings_internal(self, sentences)
    957         # put encoded batch through BERT model to get all hidden states of all encoder layers
    958         self.model.eval()
--> 959         all_encoder_layers, _ = self.model(all_input_ids, token_type_ids=None, attention_mask=all_input_masks)
    960 
    961         with torch.no_grad():

~/anaconda3/envs/gpu-dev/lib/python3.6/site-packages/torch/nn/modules/module.py in __call__(self, *input, **kwargs)
    487             result = self._slow_forward(*input, **kwargs)
    488         else:
--> 489             result = self.forward(*input, **kwargs)
    490         for hook in self._forward_hooks.values():
    491             hook_result = hook(self, input, result)

~/anaconda3/envs/gpu-dev/lib/python3.6/site-packages/pytorch_pretrained_bert/modeling.py in forward(self, input_ids, token_type_ids, attention_mask, output_all_encoded_layers)
    607         extended_attention_mask = (1.0 - extended_attention_mask) * -10000.0
    608 
[...]

PR #377, opened 2019-01-09
GH-325: Fix BERT embeddings when running using CUDA

peteriz:
BertEmbeddings fails when running on cuda backend on mixing GPU/CPU tensors.
The fix is simple - copy the model onto the GPU.

Should fix issue #325 

Thanks

### DIFF
--- a/flair/embeddings.py
+++ b/flair/embeddings.py
@@ -895,6 +895,8 @@ class BertEmbeddings(TokenEmbeddings):
             all_input_masks = torch.tensor([f.input_mask for f in features], dtype=torch.long)
 
         # put encoded batch through BERT model to get all hidden states of all encoder layers
+        if torch.cuda.is_available():
+            self.model.cuda()
         self.model.eval()
         all_encoder_layers, _ = self.model(all_input_ids, token_type_ids=None, attention_mask=all_input_masks)
 


alanakbik:
Cool, thanks for looking into this. Have you tested whether it works? 

Theoretically, explicit putting it on CUDA should not be necessary since `PreTrainedBertModel` implements `torch.nn.module`. Since we use it as a field of `BertEmbeddings` and this is also a  `torch.nn.module` putting the SequenceTagger on CUDA results in all embeddings (including BERT) to be put on CUDA. This is inbuilt into PyTorch that if you move a module to CUDA, all submodules get moved automatically as well. 

You can see this if you instantiate a SequenceTagger with BertEmbeddings and do a printout on the tagger object - all BERT classes are registered, meaning that CPU/GPU handling should already be taken care of. 

@blmoistawinde @mvss80 @ghaddarAbs  - could you test whether this fix works for you?

mvss80:
> Cool, thanks for looking into this. Have you tested whether it works?
> 
> Theoretically, explicit putting it on CUDA should not be necessary since `PreTrainedBertModel` implements `torch.nn.module`. Since we use it as a field of `BertEmbeddings` and this is also a `torch.nn.module` putting the SequenceTagger on CUDA results in all embeddings (including BERT) to be put on CUDA. This is inbuilt into PyTorch that if you move a module to CUDA, all submodules get moved automatically as well.
> 
> You can see this if you instantiate a SequenceTagger with BertEmbeddings and do a printout on the tagger object - all BERT classes are registered, meaning that CPU/GPU handling should already be taken care of.
> 
> @blmoistawinde @mvss80 @ghaddarAbs - could you test whether this fix works for you?

This worked... thanks @peteriz @alanakbik

alanakbik:
Wow cool! Great, thanks for fixing this @peteriz !

peteriz:
@alanakbik glad it helped, thanks for the fast review!

Issue #381, opened 2019-01-10
Make case randomization in TextCorpus optional


snovotney:
#219 explained the benefit for randomly flipping the case of a small percentage of sentences. For some use cases where the text is known to be all mono-case, it is useful to turn this feature off.

A proposed solution is to add a parameter in `TextCorpus` to turn off this behavior, defaulting to `True`.

tabergma:
closed by https://github.com/zalandoresearch/flair/pull/382

PR #382, opened 2019-01-10
Issue #GH-381: add param to turn off case flipping in TextCorpus

snovotney:
If a corpus is known to be mono-lingual, then it is useful to turn off
the default behavior of randomly swapping the case of a small
percentage of lines.

This fix adds a top level parameter that by default preserves the
original behavior. The user can set this parameter to False if he or
she does not want the data randomly swapped.


tabergma:
:+1: Thanks!

PR #402, opened 2019-01-17
GH-270: Use new cuda semantics

### LINKED ISSUE #270: [PyTorch] Use new cuda semantics
As described in the [CUDA semantics](https://pytorch.org/docs/stable/notes/cuda.html#best-practices) by PyTorch we should use 

```
device = None
if torch.cuda.is_available():
    device = torch.device('cuda')
else:
    device = torch.device('cpu')

x = torch.empty((8, 42), device=device)
net = Network().to(device=device)
```

instead of 

```
x = torch.empty((8,42))
if torch.cuda.is_available():
    x = x.cuda()
```

tabergma:
closes #270 

- Don't use `torch.autograd.Variable` anymore
- Use `tensor = tensor.to(flair.device)` instead of `if torch.cuda.is_available(): tensor = tensor.cuda()`

### DIFF
--- a/flair/models/language_model.py
+++ b/flair/models/language_model.py
@@ -5,6 +5,7 @@ import torch
 import math
 from torch.autograd import Variable
 from typing import List, Union, Tuple
+from typing import List
 
 from torch.optim import Optimizer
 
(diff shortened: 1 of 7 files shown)


flair/models/language_model.py
```diff
@@ -5,9 +5,11 @@
 import math
 from torch.autograd import Variable
 from typing import List, Union, Tuple
+from typing import List
```
alanakbik: duplicate import List

alanakbik:
This is much more concise, thanks!

PR #406, opened 2019-01-18
Asynchronously load LM training data

snovotney:
pytorch has Dataset and DataLoader classes to read data into memory in parallel. The changes restructures the TextCorpus and LanguageModelTrainer classes to use the DataLoader generators for the training data.  This new functionality cut the time per epoch in half on my dataset and the top level functionality is backwards compatible. 

The major changes are:
1) add a `TextDataset` class that inherits from pytorch's `Dataset`
2) Add a generator in the `train` function to asynchronously load training split files.
3) Change the overall training loop to be two nested for loops in order to use the training splits generator.
4) Remove the recent addition of `.to(flair.device)` in `charsplit` since it is redundantly called in `.batchify`
5) remove the `get_next_train_slice` in `TextCorpus` since this is now handled by the `DataLoader` generator. 

Any comments welcome, I hope you find this useful.


flair/trainers/language_model_trainer.py
```diff
@@ -161,10 +135,22 @@ def tokenize(self, path: Path):
...
+        
+    
+class TextCorpus(object):
+    def __init__(self, path: Path, dictionary: Dictionary, forward: bool = True, character_level: bool = True, random_case_flip : bool = True):
+        self.dictionary: Dictionary = dictionary
+        self.forward                = forward
+        self.split_on_char          = character_level
+        self.random_case_flip       = random_case_flip
+        
+        self.train = TextDataset(path / 'train',dictionary, True,self.forward,self.split_on_char,self.random_case_flip)
```
alanakbik: Should be `self.train = TextDataset(path / 'train', dictionary, False, self.forward, self.split_on_char,                                 self.random_case_flip)`, i.e. `expand_vocab` needs to be `False` to be consistent with previous behavior. Could throw errors otherwise for new characters that are not in the pre-defined dictionary.

alanakbik:
Hi @snovotney thanks very much for adding this - the `DataLoader` makes a lot of sense here and can really speed things up!

flair/trainers/language_model_trainer.py
snovotney: good catch. I will fix.

snovotney: I'm new to github's review functionality. I committed a change swapping the bool and see the change in 2c937f5. Let me know if there is a certain way you would like me to push the changes.

alanakbik:
This works great - I've tested on a GPU setup and like you wrote it is now almost **twice as fast**. Thanks!!

stefan-it:
Thanks for adding this :) I just want to note, that the `shm` size needs to be increased when running `flair` inside a docker container. I tried it an got an error message that is similar to [this](https://github.com/pytorch/pytorch/issues/5040) thread. Then I simply passed `--shm-size 64G` to the `nvidia-docker` command and it worked :)

PR #418, opened 2019-01-23
GH-417: Fix bug in macro avg f-score

### LINKED ISSUE #417: macro_avg_f_score bug
In the file flair/training_utils.py, when calculating macro_avg_f_score, I think 'recall' method should be called but 'precision' is called.

image

i think it needs to write like this:
`def macro_avg_f_score(self): 
         class_precisions = [self.precision(class_name) for class_name in self.get_classes()] 
         class_recalls = [self.recall(class_name) for class_name in self.get_classes()]`

tabergma:
closes #417


alanakbik:
Great!

PR #419, opened 2019-01-23
GH-415: Enable existing loggers.

### LINKED ISSUE #415: Logging overwrite less sweeping
To be removed, once it is done: Please add the appropriate label to this ticket, e.g. feature or enhancement.

When using flair in other applications, the fact that it disables existing logs in `__init__.py` can be detrimental. For instance when wrapping it up as a component in a tool like rasa_nlu, importing flair overrides all logging except its own, breaking functionality in rasa_nlu.

This is the [line that does so ](https://github.com/zalandoresearch/flair/blob/c2bb0d8776f25493a5b994dcd89a96f71ac175b8/flair/__init__.py#L13) and it was done on purpose to disable BERT logging in #282 .

Ideally, the problem of disabling logging from certain known dependencies should be much more limited in scope. Importing flair as a package shouldn't disable all the other loggers. At a minimum, perhaps the logging could only *optionally* disable all existing logs

tabergma:
closes #415

### DIFF
--- a/flair/__init__.py
+++ b/flair/__init__.py
@@ -10,6 +10,7 @@ import logging.config
 
 logging.config.dictConfig({
     'version': 1,
+    'disable_existing_loggers': False,
     'formatters': {
         'standard': {
             'format': '%(asctime)-15s %(message)s'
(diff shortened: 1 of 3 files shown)


jamesmf:
Is `disable_existing_handlers` the default? This link says it's defaulted to True for backwards compatibility

https://docs.python.org/3/howto/logging.html

alanakbik:
Great, thanks!

PR #422, opened 2019-01-24
Loosen version constraints of dependencies

### LINKED ISSUE #420: Don't pin package dependencies in setup.py
To be removed, once it is done: Please add the appropriate label to this ticket, e.g. feature or enhancement.

It is not considered good practice to pin package dependencies in setup.py (see additional context).

For instance, I'm forced to downgrade certain packages by installing flair.

Just list the abstract requirements in setup.py with less restrictive version bounds.

**Additional context**

See https://packaging.python.org/discussions/install-requires-vs-requirements/

fbergroth:
Closes #420

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,6 +1,5 @@
 torch==1.0.0
 gensim==3.4.0
-typing==3.6.4
 pytest==3.6.4
 tqdm==4.26.0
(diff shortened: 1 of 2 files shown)


alanakbik:
Great, thanks!

PR #432, opened 2019-01-31
No pinned version dependencies on requirements.txt

davidsbatista:
To keep in line with `setup.cfg`

I also removed the pinned version dependencies on `requirements.txt`


alanakbik:
Great, thanks!

PR #435, opened 2019-01-31
GH-429: relax matplotlib requirement to make compatible with allennlp

### LINKED ISSUE #429: Cannot install allennlp due to matplotlib dependency conflict
Hello, thanks for the great package. I want to play with ELMoEmbeddings, which requires package allennlp, not installed by default with Flair. However, installing latest allennlp fails because it requires matplotlib==2.2.3, while Flair requires >=3.0.0. When I install allennlp==0.7.0 (without the matplotlib requirement), it still fails because of PyTorch conflicts. Makes me wonder what versions I should use that are actually compatible?

alanakbik:
closes #429 

AllenNLP uses a lower version of matplotlib than the one we used in our requirements.txt. To make both compatible, we lower our matplotlib verison as well.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -3,7 +3,7 @@ gensim>=3.4.0
 pytest>=3.6.4
 tqdm>=4.26.0
 segtok>=1.5.7
-matplotlib>=3.0.0
+matplotlib>=2.2.3
 mpld3==0.3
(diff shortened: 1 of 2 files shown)


kashif:
👍

PR #437, opened 2019-01-31
add default class_name parameter to metric getters

kubapok:
add default class name parameter as same as in get_tp method


alanakbik:
Thanks for adding this!

PR #451, opened 2019-02-04
Output confidence score in test.tsv

gccome:
Sometimes it would be good to see confidence score for each token.


alanakbik:
Thanks for adding this!

PR #472, opened 2019-02-08
Trainer: check if Corpus.test actually contains any data before calling final_score

davidsbatista:
In some situations we just want to train a model without being forced to evaluate it, currently the trainer always calls the `final_score` function, which crashes if no test data is present, due to this statement: 

`eval_loss /= len(sentences)` present [here](https://github.com/zalandoresearch/flair/blob/28be5ba0b8f2142b349f5e2be166680b9f52dc41/flair/trainers/trainer.py#L397) and [here](https://github.com/zalandoresearch/flair/blob/28be5ba0b8f2142b349f5e2be166680b9f52dc41/flair/trainers/trainer.py#L446)

This PR checks if actually there are test data present before `final_score` is called


alanakbik:
Hey this is great, thanks for adding this!

Issue #438, opened 2019-01-31
Using Pre-trained ELMo Representations for Many Languages in Flair


mauryaland:
Hello,

First of all, thanks for the great work. This library is very useful and I follow with attention the many improvements!

I wonder if there is a possibility of implementing elmo embeddings from the repository [ELMoForManyLangs](https://github.com/HIT-SCIR/ELMoForManyLangs) ?

Thank you in advance for your answer and I am available to help if the answer is positive.

Amaury

alanakbik:
Hello @mauryaland this looks very interesting - multilingual ELMo would definitely be a great addition to Flair. Are you planning an installable pip package?

mauryaland:
@alanakbik I will ask them if they are planning to or if I can create it. I let you know about that.

alanakbik:
Cool, thanks!

mauryaland:
@alanakbik I had an answer from the author on the following [issue](https://github.com/HIT-SCIR/ELMoForManyLangs/issues/46) and the project is too much unstable so far, so wait and see. I will follow the topic.

On another side, I have discovered great embeddings recently, subwords embeddings for many languages in fact. It is called [BPEmb](https://nlp.h-its.org/bpemb/). Could be interesting to used and it is  available on pypi. Things are going really fast in NLP these days !

alanakbik:
Wow these look interesting - perhaps we can integrate them.

stefan-it:
Would be great if they had used the official ELMo training code 😂 From myexperience the Transformer ELMo model is a good alternative to the default ELMo model, and training is a lot faster.

BPE embeddings could be interesting, it would work easily for text classification, because you only have to bpe encode the sentences and train a baseline model; if you also want to use a language model, then you need to train on a BPE encoded corpus.

For sequence tagging: I made some experiments with SentencePiece (SentencePiece word embeddings + SentencePiece language model + Converting a CoNLL NER dataset also to SentencePiece) but the results weren't very promising (It's an open question how to tag the Pieces...).

But for text classification I think it is worth to try these BPEmb. E.g. there's a subword variant with SentencePiece in combination with the ULMfit model on Polish, see paper [here](https://arxiv.org/pdf/1810.10222.pdf).

mauryaland:
Thanks for the details.

Indeed, how to tag the pieces for sequence tagging seems tricky.

Appreciate the paper on SentencePiece in combination with the ULMfit model, really good results!

bheinzerling:
[...]

PR #473, opened 2019-02-08
GH-438: added byte pair embeddings

alanakbik:
Add [bype pair embeddings](http://www.lrec-conf.org/proceedings/lrec2018/pdf/1049.pdf) to Flair.

### DIFF
--- a/flair/embeddings.py
+++ b/flair/embeddings.py
@@ -7,6 +7,7 @@ from typing import List, Union, Dict
 import gensim
 import numpy as np
 import torch
+from bpemb import BPEmb
 from deprecated import deprecated
 
 from pytorch_pretrained_bert.tokenization import BertTokenizer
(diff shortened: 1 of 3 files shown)


gccome:
Thanks for adding this! Can we use BPEmb to do sequence tagging? If so, how to get the tag for each token that is separated as multiple subwords?

alanakbik:
Yes, you can use it in sequence tagging. In this current implementation, the word embedding is constructed as a concatenation of the first and last subword. Alternatively, we could try other ways such as pooling all subword embeddings, or using only the last subword. 

To use the embeddings when training a sequence labeler, do this: 

```python
corpus: TaggedCorpus = NLPTaskDataFetcher.load_corpus(NLPTask.UD_ENGLISH)

tag_type = 'upos'

# initialize embeddings
embeddings =  BytePairEmbeddings(language='en')

# initialize sequence tagger

                                        embeddings=embeddings,
                                        tag_type=tag_type)
print(tagger)

# train model
trainer.train('resources/taggers/ud_bpe')
```

stefan-it:
@alanakbik I'm currently running my UD Basque experiment with bpe embeddings and will report some results here (it's also a nice playground for testing different merge operations and their impact on the downstream task) - so e.g. here's a nice [paper](https://arxiv.org/abs/1706.09733) for machine translation that uses different merge operations (for a specific amount of parallel sentences) :)

alanakbik:
@stefan-it great, thanks - I'd be very interested to hear which parameters work best in your experience. We could then use them as default params in the BytePairEmbeddings class.

gccome:
@alanakbik Thanks Alan! This is great!

alanakbik:
@bheinzerling we noted that this workaround for the serialization has some problems, see #504

bheinzerling:
Ah, that's annoying. I cannot find a way to properly serialize SWIG objects (something like [this](https://stackoverflow.com/questions/9310053/how-to-make-my-swig-extension-module-work-with-pickle) doesn't work either).

As a further workaround, the sentence piece model is now downloaded to a temp directory if it isn't already in the cache during deserialization:

https://github.com/bheinzerling/bpemb/blob/0caeba2e2215b897a84d4f18ea55f1a3b5a320ab/bpemb/bpemb.py#L438

I'll push this change to pip later.

alanakbik:
Ah great - I think you would also need to update the `cache_dir` variable when de-serializing, otherwise it will use the `cache_dir` of the machine the model was serialized on and potentially try to download the model into a nonexisting `cache_dir`.

https://github.com/zalandoresearch/flair/blob/4d4a3a4d93fda6fc20c6bd30950891c02a2bd167/flair/embeddings.py#L268

bheinzerling:
You're right, thanks!

bheinzerling:
You're right, thanks! Hopefully the last fix for serialization now

PR #475, opened 2019-02-09
make sure evaluation is not performed over self.corpus.test if this doesn't contain any Sentence

davidsbatista:
@alanakbik regarding my latest PR there were two more small changes that needed to be made in order to make sure that the evaluation is not called on over `self.corpus.test` if this is an empty list.


alanakbik:
Great, thanks!

Issue #68, opened 2018-08-15
transformer models for language model training and tag prediction instead of LSTM's


mittalsuraj18:
I recently read the generative pretraining paper of openAI.
According to the benchmarks, fine-tuning the openAI model on a custom dataset takes a very less amount of time compared to a LSTM based approach.
Also the model has shown to improve SOTA in a lot of tasks.
So I was wondering if it is possible to replace the pipeline by a transformer based model implemented by OpenAI.

alanakbik:
Great idea - we've been discussing this internally and really want to try it out, and compare the two approaches! Any help / pointers are appreciated :)

mittalsuraj18:
```https://github.com/huggingface/pytorch-openai-transformer-lm``` has an implementation of transformer model in Pytorch and scripts to load openai transformer weights.
Will have a look at it this weekend and check out the complexity of the implementation.

alanakbik:
Great, thanks! Perhaps this code can be the basis of new transformer-based `LanguageModel` and `LanguageModelTrainer` classes!

stefan-it:
A deep Transformer model achieves state-of-the-art results also in language modeling now, see [this](https://arxiv.org/pdf/1808.04444.pdf) paper. So I think integrating such an architecture in `flair` would be awesome :heart: 

But don't look at the evaluation section in the paper mentioned above ;) I took more than 7 days on a single Cloud TPU :scream:

tabergma:
Small update: We are going to add the BERT embeddings (see https://github.com/zalandoresearch/flair/issues/251) in the next release to flair. They are based on transformers.

We are still thinking of adding our own transformer model at one point. But not in the near future.

mittalsuraj18:
alright 👍

stefan-it:
@alanakbik and @tabergma : Here's another great paper about a Transformer-based LM: 

[Transformer-XL: Attentive Language Models Beyond a Fixed-Length Context](https://arxiv.org/abs/1901.02860)

-> Yesterday they provided both a TensorFlow and PyTorch implementation of the model. I'm going to play with the implementation now, maybe I find a way to get embeddings for a sentence (like it is done with FlairEmbeddings).

alanakbik:
Wow this looks really interesting!

stefan-it:
Two PR's from the `pytorch-pretrained-BERT` repositoy are very interesting:

* [OpenAI GPT support](https://github.com/huggingface/pytorch-pretrained-BERT/pull/183)
[...]

PR #490, opened 2019-02-12
GH-68 Bert update

kashif:
for issue #68


stefan-it:
I think it should also be updated in `setup.py` :)

kashif:
@stefan-it are you working on adding transformer-xl and openai-gpt embedding classes or should i do it?

stefan-it:
@kashif I haven't started yet, so it would be great if you can add a kind of skeleton for both integrations, so that we can work together on it :)

alanakbik:
:+1:

stefan-it:
:+1:

stefan-it:
@kashif *0.6.1* is out today. New release includes the new GPT-2 model :)

Issue #483, opened 2019-02-12
Confusion Matrix for Multi-class prediction


frtacoa:
Hi guys,

I would like to share a concern regarding the prediction results output.
Currently, true negatives are being calculated for multi-class classification, which will produce duplicated counts and artificially increase the overall accuracy.
Please take a look at the following portion of log:

```
2019-02-07 22:45:48,581 Testing using best model ...
2019-02-07 22:45:48,664 MICRO_AVG: acc 0.9963 - f1-score 0.9448
2019-02-07 22:45:48,665 MACRO_AVG: acc 0.9963 - f1-score 0.9227
2019-02-07 22:45:48,665 label_1 tp: 5 - fp: 1 - fn: 0 - tn: 139 - precision: 0.8333 - recall: 1.0000 - accuracy: 0.9931 - f1-score: 0.9091
2019-02-07 22:45:48,665 label_2 tp: 5 - fp: 0 - fn: 0 - tn: 140 - precision: 1.0000 - recall: 1.0000 - accuracy: 1.0000 - f1-score: 1.0000
2019-02-07 22:45:48,665 label_3 tp: 5 - fp: 1 - fn: 0 - tn: 139 - precision: 0.8333 - recall: 1.0000 - accuracy: 0.9931 - f1-score: 0.9091
2019-02-07 22:45:48,665 label_4 tp: 5 - fp: 0 - fn: 0 - tn: 140 - precision: 1.0000 - recall: 1.0000 - accuracy: 1.0000 - f1-score: 1.0000
2019-02-07 22:45:48,665 label_5 tp: 5 - fp: 1 - fn: 0 - tn: 139 - precision: 0.8333 - recall: 1.0000 - accuracy: 0.9931 - f1-score: 0.9091
2019-02-07 22:45:48,665 label_6 tp: 5 - fp: 0 - fn: 0 - tn: 140 - precision: 1.0000 - recall: 1.0000 - accuracy: 1.0000 - f1-score: 1.0000
2019-02-07 22:45:48,665 label_7 tp: 5 - fp: 0 - fn: 0 - tn: 140 - precision: 1.0000 - recall: 1.0000 - accuracy: 1.0000 - f1-score: 1.0000
2019-02-07 22:45:48,666 label_8 tp: 5 - fp: 0 - fn: 0 - tn: 140 - precision: 1.0000 - recall: 1.0000 - accuracy: 1.0000 - f1-score: 1.0000
[...]

PR #495, opened 2019-02-13
GH-483: added unit test for multi label classification

alanakbik:
added a data set and a simple prediction problem for integration testing of multi label classification


alanakbik:
:+1:

kashif:
👍

PR #506, opened 2019-02-15
GH-504: workaround for serialization of BPEmb object

### LINKED ISSUE #504: BytePairEmbeddings serialization error
In the current master branch, the `BytePairEmbeddings` serialization creates problems when serialized on another machine than where they get deserialized.  

Specifically, this line causes the problem: 

https://github.com/zalandoresearch/flair/blob/4b3562eed534e4e305a36d2f22d12961bc6c25d5/flair/embeddings.py#L261 

Since this is only a path, this only works if the sentence piece model is already on your local machine under exactly the same path. If I train remotely and then deserialize locally, I get a `No such file or directory Error`.

alanakbik:
closes #504 

Workaround in which the sentence piece model is serialized as a binary that is written to the appropriate folder if missing.

### DIFF
--- a/flair/embeddings.py
+++ b/flair/embeddings.py
@@ -1,3 +1,4 @@
+import os
 import re
 import logging
 from abc import abstractmethod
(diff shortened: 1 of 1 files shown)


53X:
Thanks for this fix. I wanted to push the **Attentional BiLSTM** today but found this error after a recent git pull

alanakbik:
@53X hey that's great - really looking forward to the attentional BiLSTM!

alanakbik:
:+1:

kashif:
👍

Issue #511, opened 2019-02-16
DocumentLSTMEmbeddings use GRU instead of LSTM


kubapok:
In DocumentLSTMEmbeddings there is:

`        self.rnn = torch.nn.GRU(self.embeddings_dimension, hidden_size, num_layers=rnn_layers,
                                bidirectional=self.bidirectional)`

which is inconsistent naming. Both GRU and LSTM are type of RNN networks, but GRU is not the same as LSTM.

alanakbik:
fixed in #512 - thanks!!

PR #512, opened 2019-02-16
GH-511: change DocumentLSTMEmbeddings to use LSTM instead of GRU

kubapok:
add other available RNN types to choose from

### DIFF
--- a/flair/embeddings.py
+++ b/flair/embeddings.py
@@ -1497,7 +1497,7 @@ class DocumentPoolEmbeddings(DocumentEmbeddings):
         pass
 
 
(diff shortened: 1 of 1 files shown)


alanakbik:
Hey this is great, thanks for adding this. 

However, I am getting an error now ('`DocumentLSTMEmbeddings' object has no attribute 'embedding_length`') if I call this code: 

```python
classifier = TextClassifier.load('en-sentiment')
```

There may a problem with backwards compatibility with serialized models.

alanakbik:
@kubapok thanks for commiting this - I'll merge this into the release branch and take alook at the error with the serialized models!

PR #538, opened 2019-02-21
Adds Attentional BiLSTM in the Text Classification architecture

### LINKED ISSUE #460: Introducing Attentional BiLSTM for Text Classification
As per my understanding ,  Flair by default uses the **standard BiLSTM model** for text classification.

I guess introducing **Attentional BiLSTM** might improve this maybe?

An easy solution to this might be to just introduce a parameter named **attention** in the **TextClassifier** class that decides whether to use attention or not

53X:
This fixes #460 .    The attention used here is **Soft Attention**.


alanakbik:
Hi @53X thanks for adding this! Could you remove the files: 

- flair/Python-3.6.5.tar.xz
- train.py

from the PR? The embeddings.py and text_classification_model.py should be enough for the PR!

53X:
@alanakbik , Done

alanakbik:
Thanks - could you also remove `train.py`?

53X:
@alanakbik , is it LGTM?

53X:
@alanakbik , can you please help me out here a bit ?

53X:
`train.py` was originally there is Flair. I think that keeping it doesn't hurt. What do you say?

alanakbik:
oh yes, sorry. train.py can stay!

alanakbik:
Thanks! I'll merge to a branch and look into the failing unit test!

53X:
@alanakbik  when can this be merged to the master ?

alanakbik:
Hi @53X the code works well, but was only working in a very specific setup, i.e. if we use `DocumentLSTMEmbeddings` with `bidirectional=False`. I made some changes to get it to run for arbitrary document embeddings by using a hooking mechanism: 

https://github.com/zalandoresearch/flair/blob/6f10533ea293c5031e83acc9b45483bf9f6ec957/flair/models/text_classification_model.py#L94 

This allows us to preserve the original `.embed()` interface, so it maintains the same method signature as all other embeddings: 

https://github.com/zalandoresearch/flair/blob/6f10533ea293c5031e83acc9b45483bf9f6ec957/flair/embeddings.py#L1821

Could you take a look at this code to see if it works the way as intended?

53X:
Sure @alanakbik

alanakbik:
@53X just pushed some more fixes for GPU and gradient propagation. Seems to work now with DocumentLSTMEmbeddings and DocumentRNNEmbeddings with different parameters.

53X:
Great. Let me know if you need me to do anything else

alanakbik:
Could you verify that its working as intended? If so, I'll merge!

heukirne:
Hello @53X , 
How can I extract the feature importance from attention layer?
How to list the most informative word for the classifier model?
Thanks!

PR #553, opened 2019-02-21
GH-483: accuracy calculation

alanakbik:
changes the accuracy calculation as to be consistent with scikit-learn. See discussion in #483


alanakbik:
:+1:

kashif:
👍

PR #576, opened 2019-03-01
Adding support to portuguese pre-trained word embeddings from NILC lab

pvcastro:


flair/embeddings.py
```diff
@@ -1937,3 +1937,47 @@ def _add_embeddings_internal(self, sentences: List[Sentence]):
...
+        cache_dir = Path('embeddings')
+
+        # GLOVE embeddings
+        if embeddings.lower() == 'glove':
+            cached_path(f'{base_path}{embeddings}/{embeddings}_s{size}.zip', cache_dir=cache_dir)
+            embeddings = cached_path(f'{base_path}{embeddings}/{embeddings}_s{size}.zip', cache_dir=cache_dir)
+
+        elif embeddings.lower() in ['fasttext', 'wang2vec', 'word2vec']:
+            cached_path(f'{base_path}{embeddings}/{model}_s{size}.zip', cache_dir=cache_dir)
+            embeddings = cached_path(f'{base_path}{embeddings}/{model}_s{size}.zip', cache_dir=cache_dir)
```
pvcastro: I realized that there's a problem here. All the zip files at their repository have the same name pattern: {model}_s{size}.zip. So how can I cache these using different names for different embedding types? For an example: at the cache_dir I would have a word2vec_skip_s100.zip and wang2vec_skip_s100.zip.

stefan-it: Maybe you could add a model signature to the `cache_dir` path, e.g. `cache_dir = Path('embeddings') / embeddings.lower() / f'{model}_s{size}'?

pvcastro: OK, done. Thanks for the tip!

heukirne:
Excelent, @pvcastro !!

alanakbik:
@pvcastro looks good from my side, thanks!

Just one minor thing: When you initialize the embeddings, it can take while until they are loaded into memory. But there is no logging output, which could confuse some users who might wonder what is happening right now. Could you add some logging output like "Loading X embeddings" and "Loading X embeddings complete"?

pvcastro:
Hi @alanakbik , sure. But can you point me to an example showing exactly where this should be done? I'm asking because I saw no such logging in the WordEmbeddings class, and I tried keeping the same type of code. 
Thanks!

alanakbik:
Sure! The logger is already initialized in the embeddings class as a global variable, here: 

https://github.com/zalandoresearch/flair/blob/ccb2ffb4080550d25e1e067e2e82e661c8b81411/flair/embeddings.py#L27 

Then, you could add a `log.info` line, like we do in the ModelTrainer class, for instance in this line:

https://github.com/zalandoresearch/flair/blob/ccb2ffb4080550d25e1e067e2e82e661c8b81411/flair/trainers/trainer.py#L126 

If could be a log.info that gets printed right before loading the embeddings that just prints 'reading embeddings [name]'.

pvcastro:
Thanks @alanakbik . Done, please let me know if it's ok.

alanakbik:
Yes, looks good!

alanakbik:
:+1:

alanakbik:
@kashif @tabergma could one of you check, +1 and merge?

tabergma:
👍

PR #582, opened 2019-03-03
GH-538: attention

alanakbik:
This PR adds a layer of attention implemented by @53X to the `TextClassifier`. 

You can activate it by setting `attention=True` in the `TextClassifier` initialization.


alanakbik:
:+1:

alanakbik:
@MichaelHintz @stefan-it @kashif could you check, +1 and merge?

53X:
Hi, @alanakbik  @stefan-it  @kashif @MichaelHintz can this be merged? I am actually planning to use it in one of my problems

kashif:
@53X I fixed the conflict, is it possible to add a test?

53X:
@kashif , can you please elaborate on the kind of test?

53X:
@alanakbik , there is a bug in this code .

alanakbik:
@53X what is the bug?

53X:
While loading back the model, if we forget to pass the bidirectional parameter's value, then only it gives an error for bidirectional cases. Otherwise it's fine. Adding that solves the issue. See **load_from_file** function in text_classifcation_model.py

loopdigga98:
Are you planning to add attention to NER classifier?

alanakbik:
@loopdigga98 that's a good idea. We are currently refactoring the model interfaces and classes so adding attention as an optional layer in all tasks could be interesting!

stefan-it:
One question: could the newly introduced [MultiheadAttention](https://github.com/pytorch/pytorch/pull/18334) class be used here 🤔 It is available in PyTorch 1.1.0 :)

alanakbik:
Oh looks interesting! Might be worth a try :)

aptperson:
Is this still going to be merged?

djstrong:
Here are perplexities that can be achieved by adding single head attention to LSTM: https://github.com/Smerity/sha-rnn

lfdharo:
Hello everyone, is it possible to get the attention weights that the TextClassifier is assigning to each word/token in the sentence to provide the final score in a multiclass setup? I thought that this was the spirit of the proposed incorporation to the library, but it seems it is not currently available in the latest version.

PR #626, opened 2019-03-24
Small fix for training example script

stefan-it:
Hi,

this PR fixes the training example script. The current `train.py` example script trains a model on English Universal Dependencies using the `pos` tag type. But that tag type is not correct for Universal Dependencies, because `upos` (indicating universal part-of-speech tags) should be used instead.

See also [CoNLL-U](https://universaldependencies.org/format.html) format description.


alanakbik:
Thanks! Feel free to +1 and merge!

alanakbik:
:+1:

tabergma:
👍

Issue #627, opened 2019-03-25
How to add a token


nh111:
Hi All,
I have used below code to split the sentences and tokenize the sentence into words:
`sentences = [Sentence(sent, use_tokenizer=True) for sent in split_single(Text)]`

in each sentence i got tokens, but now i want to add some tokens to some sentences, how can i do that?
I tried with **add_token** function, but it is adding a string data type. please help

stefan-it:
You could use the following:

```python
x = Sentence("Munich is a nice")
print(x)
x.add_token(Token("city"))
x.add_token(Token("."))
print(x)
```

Should output:

```python
Sentence: "Munich is a nice" - 4 Tokens
Sentence: "Munich is a nice city ." - 6 Tokens
```

stefan-it:
I think it would be a good enhancement to also support strings here.

@alanakbik If you want I can extend the `add_token` to also support strings :)

nh111:
Thanks for your reply, I used similar to above. I came to know that we have to use "Token"
first we have to import Token as below:

`from flair.data import Token`

Then I used similar to stefan-it code. it works.

nh111:
Hi, are you expecting anything from me?

stefan-it:
@nh111 No, you just got a notification from GitHub, because I referenced this issue here in a pull request :)

PR #628, opened 2019-03-25
GH-627: Extend add_token method

stefan-it:
Hi,

with this PR both string and Token objects can be passed to the `add_token` method.

This was requested in #627.


alanakbik:
This is great - thanks for adding this!

alanakbik:
:+1:

tabergma:
👍

PR #640, opened 2019-04-02
Use a list of tokens obtained by any external tokenizer in Sentence class

mauryaland:
Related to this [discussion](https://github.com/zalandoresearch/flair/issues/563#issuecomment-476555364) and the possibility to use any tokenizer we want to use in the Sentence class. 

The current way to use an external tokenizer ( sentence = Sentence(' '.join(tokens_from_list)) ) is in my opinion not robust when we want to combine the SequenceTagger with other methods in a processing pipeline. Indeed, the indexes from original text and the text in the Sentence class will not be the same.

The ability to import tokens from a list in the Sentence class solves this problem and makes life easier for users 😃 .

### DIFF
--- a/flair/data.py
+++ b/flair/data.py
@@ -273,7 +273,7 @@ class Sentence:
     A Sentence is a list of Tokens and is used to represent a sentence or text fragment.
     """
 
-    def __init__(self, text: str = None, use_tokenizer: bool = False, labels: Union[List[Label], List[str]] = None):
+    def __init__(self, text: str = None, use_tokenizer: bool = False, tokens_from_list: List[str] = None, labels: Union[List[Label], List[str]] = None):
 
         super(Sentence, self).__init__()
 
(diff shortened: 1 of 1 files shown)


stefan-it:
Thanks for this PR :) I'll add a review now!

flair/data.py
```diff
@@ -290,12 +290,17 @@ def __init__(self, text: str = None, use_tokenizer: bool = False, labels: Union[
             # tokenize the text first if option selected
             if use_tokenizer:
```
stefan-it: By default `use_tokenizer` is False. In order to use `tokens_from_list` it must be manually set to `True` (which is a bit complicated I think).

stefan-it: Another problem: the code only works when you specify `text`:

```python
 # if text is passed, instantiate sentence with tokens (words)
        if text is not None:
```

 I think this is not necessary, because you already pass the tokenized text.

With the current implementation, it is possible to use it with:

```python

def test_sentence_external_tokenization():
    text: str = "Berlin and Munich are nice cities ."

    tokens: List[str] = text.split()

    sentence: Sentence = Sentence(text=text, use_tokenizer=True, tokens_from_list=tokens)

    assert ("Berlin and Munich are nice cities ." == sentence.to_tokenized_string())
```

which is too complicated. I think both `text` and `use_tokenizer` are not really needed here.

mauryaland: Thank you for your review.

For the first point, I thought that tokens_from_list should be a choice dependent to use_tokenizer=True, because we choose either segtok either any external tokenizer. But you right it is a bit complicated. It can totally be an independent argument.

For the second point, if the original text is needed, it is because I wanted to keep the same behavior and the same code that when we use the segtok tokenizer. It allows to use entity span indexes to work on the original text easily. Let me clarify this by an example using spacy french tokenizer:

```python
from spacy.lang.fr import French
nlp = French() # use only the tokenizer 
text = "J'aime vivre à Montréal lorsqu'il neige."
doc = nlp(text)
tokens = [token.text for token in doc] # ["J'", 'aime', 'vivre', 'à', 'Montréal', "lorsqu'", 'il', 'neige', '.']
sentence = Sentence(text, use_tokenizer=True, tokens_from_list=tokens)

assert(text == sentence.to_original_text())
assert("J' aime vivre à Montréal lorsqu' il neige ." == sentence.to_tokenized_string())
```

In my opinion, it is valuable to keep this behavior, but maybe it could be optional?

PR #642, opened 2019-04-04
Add the possibility to get the score for all tags when using CRF

mauryaland:
This PR makes possible to get the score of all tags for each token when using a CRF with the SequenceTagger. It could be an information that people want to have access.

However, I am not sure that accessing this information directly from the SequenceTagger class is the best solution. Let me know what do you think about that.


alanakbik:
Hello @mauryaland thanks for the PR and sorry for the late reply (I was out of office). A quick question: Do you think access to the scores is necessary for some use cases? If so we could add this, but if not I would hesitate to add a public field to the SequenceTagger.

mauryaland:
Hello @alanakbik, no worries and thank you for your answer. 

Currently, I am working on the anonymization of court decisions using NER. Therefore, we do not want to miss an important entity. The ability to get all the scores for each token permits to raise a doubt when we have a tag 'O' below a threshold and to propose the second choice. It is, in my opinion, a good feature to have.

 I guess we could just save the second highest score instead of all scores in order to save memory footprint.

davidsbatista:
Just a suggestion, would it be easy to also get the CRF transition matrix?

Philipduerholt:
This seems to be related https://github.com/zalandoresearch/flair/issues/745

Philipduerholt:
I created an updated PR based on this: https://github.com/zalandoresearch/flair/pull/782.

alanakbik:
I'll close this PR since its subsumed by #782. Thanks again @mauryaland!

PR #661, opened 2019-04-13
add option to support customized elmo embeddings

GiliGoldin:
I added two optional arguments to ELMoEmbeddings's init function: 
1) options_file
2) weights_file
This way a customized ELMO model can be provided by the user for embeddings.
This is useful for using a self-trained ELMO model or any known ELMO models that are not yet supported by Flair.


flair/embeddings.py
```diff
@@ -354,23 +354,23 @@ def __init__(self, model: str = 'original'):
...
-        if model == 'medium':
-            options_file = 'https://s3-us-west-2.amazonaws.com/allennlp/models/elmo/2x2048_256_2048cnn_1xhighway/elmo_2x2048_256_2048cnn_1xhighway_options.json'
-            weight_file = 'https://s3-us-west-2.amazonaws.com/allennlp/models/elmo/2x2048_256_2048cnn_1xhighway/elmo_2x2048_256_2048cnn_1xhighway_weights.hdf5'
-        if model == 'pt' or model == 'portuguese':
-            options_file = 'https://s3-us-west-2.amazonaws.com/allennlp/models/elmo/contributed/pt/elmo_pt_options.json'
-            weight_file = 'https://s3-us-west-2.amazonaws.com/allennlp/models/elmo/contributed/pt/elmo_pt_weights.hdf5'
-        if model == 'pubmed':
-            options_file = 'https://s3-us-west-2.amazonaws.com/allennlp/models/elmo/contributed/pubmed/elmo_2x4096_512_2048cnn_2xhighway_options.json'
-            weight_file = 'https://s3-us-west-2.amazonaws.com/allennlp/models/elmo/contributed/pubmed/elmo_2x4096_512_2048cnn_2xhighway_weights_PubMed_only.hdf5'
+        if not options_file or not weight_file:
```
stefan-it: I think we can just initialize both `options_file` and `weight_file` via constructor argument:

```python
def __init__(self, model: str = 'original', 
                    options_file: str = allennlp.commands.elmo.DEFAULT_OPTIONS_FILE,
                    weight_file: str = allennlp.commands.elmo.DEFAULT_WEIGHT_FILE):
```

Then it is semantically clear, that these default values are used + no `if not` checks are needed here :)

stefan-it:
Thanks for that PR :) I just left a review ☺️

flair/embeddings.py
GiliGoldin: @stefan-it 
Hi, 
thanks for the review.

This is what i initially wanted to do, but the code currently imports allennlp only at a later stage, in a try except block
```python
try:
      import allennlp.commands.elmo
except:
```
so the default options and weight files are unavailable in the constructor.
Any suggestions?

stefan-it: Oh, sorry. I forgot that we import `allennlp` later...

alanakbik:
:+1:

alanakbik:
@GiliGoldin thanks for adding this! 
@stefan-it I've merged to the new code formatting style - could you +1 and merge?

stefan-it:
:+1:

alanakbik:
Thanks!

PR #663, opened 2019-04-15
Fixed a method call and a missing import.

khituras:
I think the `find_learning_rate` call should be applied to the `trainer` object instead to the `ModelTrainer` class, the latter not working in `0.4.1` anyway.
Also the `Plotter` import was missing.


alanakbik:
:+1:

kashif:
👍

PR #664, opened 2019-04-16
GH-605: add softmax to single label classification for confidence score

### LINKED ISSUE #605: TextClassifier label predictions always have score 1.0
Hi guys,

I have trained my own TextClassifier following this tutorial: https://github.com/zalandoresearch/flair/blob/master/resources/docs/TUTORIAL_7_TRAINING_A_MODEL.md

I am using the option multi_label=False, as each sentence should be assigned only one label. In terms of embeddings, I use FlairEmbeddings mix-forward and mix-backward. 

The issue is that every time I predict the label of a new unseen sentence, I get a label score = 1.0. It seems that it never has any different value between 0.0 and 1.0. It is always 1.0. 

Is this the expected behavior? What am I doing wrong?

Thanks!

alanakbik:
closes #605 

Adds a softmax layer to get confidence scores for predictions in the `TextClassifier`.

### DIFF
--- a/flair/models/text_classification_model.py
+++ b/flair/models/text_classification_model.py
@@ -252,7 +252,8 @@ class TextClassifier(flair.nn.Model):
         return labels
 
     def _get_single_label(self, label_scores) -> List[Label]:
-        conf, idx = torch.max(label_scores, 0)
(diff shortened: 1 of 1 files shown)


kashif:
👍

alanakbik:
:+1:

PR #672, opened 2019-04-17
Relax `regex` version in dependencies

mattilyra:
I don't see any reason why `regex` is pinned to `2018.1.10`. This is causing some issues with our other dependencies so I removed the version pin.

All tests pass locally on python 3.6

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -11,4 +11,4 @@ deprecated>=1.2.4
 hyperopt>=0.1.1
 pytorch-pretrained-bert>=0.6.1
 bpemb>=0.2.9
-regex==2018.1.10
+regex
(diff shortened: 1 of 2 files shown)


alanakbik:
Hi @mattilyra makes sense - this was added in a brief window in which the most current regex package had an error causing problems for persons doing a fresh install, but this problem is resolved now. 

Could you also change this in setup.py?

mattilyra:
Didn't notice that it was in `setup.py` as well -- changed.

I don't know if there's any specific requirement for you to restrict `regex` to anything  so I removed any version spec to allow downstream packages flexibility on the version.

Since `sklearn` also doesn't have any version spec on it I'm assuming this is accepted practice for you?

alanakbik:
Thanks - looks good from my side!

alanakbik:
:+1:

kashif:
👍

PR #683, opened 2019-04-24
Log test_loss to loss.tsv

abishekk92:


BernierCR:
Why is this PR not approved? All it does is fix an obvious typo.

alanakbik:
Yes we should have merged this PR when it came in - very sorry about this @abishekk92! 

@BernierCR it looks like the issue was already fixed as part of a broader change of how we log in the trainer. So the test loss is currently being logged. I'll close the issue.

PR #687, opened 2019-04-25
Change default pooling operation for PooledFlairEmbeddings

stefan-it:
Hi,

as briefly discussed with @alanakbik, this PR changes the default pooling operation for the `PooledFlairEmbeddings` from `fade` to `min`.

`fade` was not even mentioned in the upcoming NAACL paper, and `min` leads to good results:

image


alanakbik:
Thanks for correcting this!

alanakbik:
:+1:

PR #707, opened 2019-05-02
capability to change threshold during multi label classification

ashutoshsingh0223:
Ability to define custom threshold level for multi-label classification problems while initialising the `TextClassifier` class. Default value in `0.5`

Eg
```  
text_classifier = TextClassifier(doc_embeddings,label_dictionary=label_dictionary, multi_label=True, multi_label_threshold=0.7 )
```

Any changes can be made later on just by changing this variable. For Eg.
```
classifier = TextClassifier.load_from_file(<model_path>)
classifier.multi_label_threshold = 0.65
```


ashutoshsingh0223:
@stefan-it . Can you take a look??

PR #727, opened 2019-05-15
remove __init__ from tests

MSusik:
`__init__.py` in `tests` directory creates a situation where package `tests` is installed within `site-packages`.

Obviously, that causes problems if one wants to write his own tests in `tests` directory within his project and uses flair.

I suggest to remove the file. Then, setup.py will omit `tests` directory as it is no longer a package.

Additionally I added an entry in .gitignore that might be useful.


alanakbik:
Thanks! Unfortunately, this seems to have broken the travis build. 

@tabergma any idea why the `__init__.py` is needed here?

MSusik:
@alanakbik I will check if I can fix the build

MSusik:
https://travis-ci.org/zalandoresearch/flair/builds/532794466 - this is the build for 6ad679c that passes

Why the succeeding builds don't pass is beyond my knowledge of the project. I can revert to 6ad679c if you want.

alanakbik:
Your code looks good to me - I'll take a look at the travis error.

tabergma:
`__init__.py` just marks `tests` as a package. If that is not wanted anymore, I guess we can remove it.

alanakbik:
Hm the unit test that suddenly fails now also fails in other previously healthy PRs, but still passes locally. The specific test that fails downloads models from AWS and uses them, but times out. It may be that somehow travis cannot connect to AWS at the moment and times out. I'll try again later.

alanakbik:
Hello @MSusik - I think the problem is caused by the new gitignore that includes the `regression_test` folder that the failing unit test requires. Could I ask you to comment out the test `test_trainer_results` in `test_text_regressor.py`? I'll fix this test later then.

alanakbik:
:+1:

MSusik:
OK, thanks for help!

kashif:
👍

alanakbik:
:+1:

alanakbik:
@MSusik thanks for adding this!

Issue #457, opened 2019-02-05
Unable to load corpus


romanovzky:
**Describe the bug**
Following [Tutorial 7](https://github.com/zalandoresearch/flair/blob/master/resources/docs/TUTORIAL_7_TRAINING_A_MODEL.md) I cannot load my own corpus using

```python
corpus = NLPTaskDataFetcher.load_classification_corpus(Path('./'),
                                                       test_file='test.txt',
                                                       dev_file='val.txt',
                                                       train_file='train.txt') 
```

The data in the `txt` follow the convention `__label__X\ttext`. The total amount of classes is two, the total amount of entries is:
- train: 20387
- val: 20387
- test: 10194

The used memory rises continuously, ending up having all memory used by the code above, leading to non-responsive system and then python crashes (running inside a jupyter notebook, so the kernel restarts).

Reducing the number of documents (say produce only 1000-1000-500 sets) seems to workaround this. However this is not intended behaviour for a Deep Learning solution.

**To Reproduce**
As far as I know, data with the above specifications (I cannot divulge or share the data).

**Expected behavior**
To load a corpus to be used in a `TextClassifier`.

**Screenshots**
N/A

**Environment (please complete the following information):**
 - OS [e.g. iOS, Linux]: Linux
 - Version [e.g. flair-0.3.2]: 0.4.0
 - RAM: 16GB

**Additional context**
N/A

alanakbik:
Hello @romanovzky thanks for reporting this. This is in fact related to #426 and something that has been on our list for a while. Essentially, to fix this we need to implement an iterating data fetcher that does not keep the entire data set in memory as is currently done. Since more and more people are training classifiers over large data sets we need to make this a priority.

I will add a 'feature' tag to this issue.

ShaohongBai:
@alanakbik, do we have any work around solution currently for large data (it is very standard in my industry)for loading issue like this? Additionally, whether is there a functionality does need reload the data from text after formating, just batching the data in the loader?

alanakbik:
The only current workaround is to reduce the training data set size, which is not great. We're making this a priority feature for version 0.5 (see #563)!

superzadeh:
Also looking forward to this feature in 0.5, currently we can't use Flair at all to train text classifiers.

[...]

Issue #426, opened 2019-01-27
Iterating data fetcher for very large datasets


KobeDeShow:
I had 100 million samples,  when i used  data_fetcher.load_column_corpus api,  it costed 100G memory!   how can i fix this problem?

alanakbik:
Hi @KobeDeShow thanks for raising this. The problem is that our current implementation loads the entire data set into memory, which will not work well for data sets of this size. This is something we have been meaning to find a better solution for, but don't yet know exactly how to best address this.

One idea would be to use some sort of iterator or the DataLoader API as was recently added to the `LanguageModelTrainer` (master branch).

I'll add a 'feature' tag to this issue - this is something we need to look into. As always, any suggestions/ideas from the community would be greatly appreciated :)

PR #735, opened 2019-05-17
GH-457: PyTorch DataLoader

### LINKED ISSUE #458: Iterating data fetcher for large training data sets
When training a text classifier or sequence labeler, the current implementation loads the entire training data set into memory. However, many community members have reported that this requires too much memory when the training data set is very large, as is the case often in text classification (see #457 and #426 for instance).

Task: Develop a solution that does not read everything into memory. The PyTorch DataLoader API might be a good candidate for a solution

alanakbik:
This PR refactors the way datasets are loaded in Flair. Instead of loading all data sets into memory as a List of `Sentence`, we now utilize PyTorch's `Dataset` and `DataLoader` methods and allow the user to choose whether or not to load a data set into memory. This allows us to scale training to very large datasets that do not fit into memory. 

This PR also changes the syntax of how to load datasets. 

Old way: 
```python
from flair.data_fetcher import NLPTaskDataFetcher, NLPTask
corpus = NLPTaskDataFetcher.load_corpus(NLPTask.UD_ENGLISH)
```

New way: 
```python
import flair.datasets
corpus = flair.datasets.UD_ENGLISH()
```

To use streaming loading, i.e. to not load into memory, you can pass the `in_memory` parameter:
```python
import flair.datasets
corpus = flair.datasets.UD_ENGLISH(in_memory=False)
```

closes #458 and #457 #426


kashif:
👍

PR #752, opened 2019-05-24
Actually exclude tests from packaging

MarcDufresne:
This mistake is causing issues in my own unit tests, when trying to import utils from my `tests` package Python gets confused and ends up trying to import from Flair's `tests`.


alanakbik:
Hello @MarcDufresne - thanks for spotting and fixing this!

alanakbik:
:+1:

kashif:
👍

PR #771, opened 2019-05-31
change the last test.tsv

valbarriere:
So that the test.tsv at the end correspond to the scores of the final_test's predictions.


alanakbik:
@valbarriere thanks for adding this - makes a lot of sense! 

And sorry for the delay in response (I was busy attending NAACL last week).

alanakbik:
:+1:

kashif:
👍

PR #782, opened 2019-06-05
GH-642: Added return probability distribution over all classes

Philipduerholt:
Hi,

this PR is about issue https://github.com/zalandoresearch/flair/issues/745. It:

- adds the property `tags_proba_dist` to `Token`, which contains a list of `Label`. The list of labels represents the probability distribution over all possible labels for this token.
- modifies `SequenceTagger._obtain_labels` to not only return the most probable label, but in addition a complete list of all possible labels with their respective scores.

The required changes in `SequenceTagger._viterbi_decode` are taken from https://github.com/zalandoresearch/flair/pull/642.

**Disclaimer:**  This is my first PR to the flair project. I'm happy for any feedback or ideas on how to improve this.

### DIFF
--- a/flair/models/sequence_tagger_model.py
+++ b/flair/models/sequence_tagger_model.py
@@ -237,7 +237,7 @@ class SequenceTagger(flair.nn.Model):
                 with torch.no_grad():
                     features = self.forward(batch)
                     loss = self._calculate_loss(features, batch)
-                    tags = self._obtain_labels(features, batch)
+                    tags, _ = self._obtain_labels(features, batch)
 
                 eval_loss += loss
 
@@ -351,11 +351,12 @@ class SequenceTagger(flair.nn.Model):
 
                 with torch.no_grad():
                     feature = self.forward(batch)
-                    tags = self._obtain_labels(feature, batch)
+                    tags, all_tags = self._obtain_labels(feature, batch)
 
-                for (sentence, sent_tags) in zip(batch, tags):
-                    for (token, tag) in zip(sentence.tokens, sent_tags):
+                for (sentence, sent_tags, sent_all_tags) in zip(batch, tags, all_tags):
+                    for (token, tag, token_all_tags) in zip(sentence.tokens, sent_tags, sent_all_tags):
                         token.add_tag_label(self.tag_type, tag)
+                        token.add_tags_proba_dist(self.tag_type, token_all_tags)
 
                 # clearing token embeddings to save memory
                 clear_embeddings(batch, also_clear_word_embeddings=True)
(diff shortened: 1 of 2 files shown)


flair/models/sequence_tagger_model.py
```diff
@@ -587,11 +601,21 @@ def _viterbi_decode(self, feats):
             _, idx = torch.max(backscore, 0)
             prediction = idx.item()
             best_scores.append(softmax[prediction].item())
+            scores.append([elem.item() for elem in softmax.flatten()])
+        # This has been taken from https://github.com/zalandoresearch/flair/pull/642
```
jantrienes: It is not entirely clear to me what this code actually does. I surrounded it with a simple assertion and it seems that scores remains unchanged. 

```py
 _scores_before = scores.copy()

swap_best_path, swap_max_score = (
    best_path[0],
    scores[-1].index(max(scores[-1])),
)

scores[-1][swap_best_path], scores[-1][swap_max_score] = (
    scores[-1][swap_max_score],
    scores[-1][swap_best_path],
)
assert _scores_before == scores
```

 @mauryaland Since you added this code in your original PR, can you perhaps shed some light on this? :)

flair/models/sequence_tagger_model.py
```diff
@@ -514,26 +515,28 @@ def _calculate_loss_old(self, features, lengths, tags) -> float:
             score /= len(features)
             return score
 
-    def _obtain_labels(self, feature, sentences) -> List[List[Label]]:
+    def _obtain_labels(self, feature, sentences) -> (List[List[Label]], List[List[List[Label]]]):
```
jantrienes: I think it's not very transparent what the return values correspond to. Maybe we can add some explanation along the lines of:

```py
"""
Returns a tuple of two lists: 
    - The first list corresponds to the most likely `Label` per token in each sentence.
    - The second list contains a probability distribution over all `Labels` for each token in a sentence for all sentences.
"""
```py

Philipduerholt:
Hi @jantrienes, thank you for your timely feedback.

I'm glad everything worked as expected. If there are any other suggestions, please let me know.

alanakbik:
@Philipduerholt thanks for adding this! A lot of people will find this useful.

alanakbik:
:+1:

kashif:
👍

alanakbik:
@Philipduerholt and @mauryaland: thank you for adding this!

flair/models/sequence_tagger_model.py
mauryaland: Sorry for the late reply! This is done when using CRF because sometimes the best path is not related to the max score.  However,I'm not sure that my implementation is optimal or always good.

Glad to see this feature added and thanks @Philipduerholt for having enhance and finish the few work I started.

Issue #788, opened 2019-06-10
training.log not closed before next model


adodge:
**Describe the bug**
Training multiple models in one script results in the `training.log` files of earlier models including the log output of all subsequent models.

**To Reproduce**
Do something like this (which might naturally happen during some sort of grid search):

```
# [create classifier and corpus somehow]
for outdir in ['a', 'b']:
    ModelTrainer(classifier, corpus).train(outdir)
```

Note that the contents of `b/training.log` are also appended to the end of `a/training.log`.

**Expected behavior**
Each training.log file should contain only the log output involved in training that model.

**Environment (please complete the following information):**
 - Ubuntu
 - Flair 0.4.1

alanakbik:
Hello @adodge I've just pushed a PR to master that should fix this error. Could you check if it works for you?

PR #799, opened 2019-06-13
GH-788: fix training log

alanakbik:
closes 788 

This PR fixes the open log file handler issue when training multiple models in a loop. We also add more log output such as the model and training parameters.


adizdari:
👍

alanakbik:
:+1:

Issue #784, opened 2019-06-10
Let SequenceTagger.predict() keep word embeddings


khituras:
Request: Let the word embeddings, which are computed in the `SequenceTagger.predict()` method, be set on the tokens of the sentence on request.

Background: In some cases it might be useful to obtain the word embeddings, especially when there are context-aware embeddings in the model like `FlairEmbeddings` or `ElmoEmbeddings`. Currently, all embeddings are cleared at the beginning and at the end of the `SequenceTagger.predict()` method. Those would need to be recomputed if required, despite being computed for sequence tagging anyway.

Solution: I would propose to add a parameter `clear_word_embedding=True` to the method. The parameter value could then be passed to the `clear_embeddings()` at the end of the `predict()` method of the `SequenceTagger`.

Would this be acceptable?

alanakbik:
Hello @khituras that would be a good feature! 

However there are some potential difficulties, namely what happens when you apply two taggers trained with different embeddings to the same text. 

To illustrate: 

```python
# a sentence that we want to both POS tag and NER
sentence = Sentence('I love Berlin')

# predict NER
ner: SequenceTagger = SequenceTagger.load('ner')
ner.predict(sentence)

# predict POS
pos: SequenceTagger = SequenceTagger.load('pos')
pos.predict(sentence)
```

This code would break if we did not delete the embeddings, since the second tagger was trained with different embeddings. But, if we make the argument you propose optional, maybe this is not a big problem for many users.

khituras:
Hey @alanakbik, thanks very muc for your response!

I actually don't think that this would be an issue because the `SequenceTagger.predict()` method actually cleans the word embeddings twice:

1. In line 333
  ```python
    # remove previous embeddings
    clear_embeddings(filtered_sentences, also_clear_word_embeddings=True)
```
2. In line 365
```python
    # clearing token embeddings to save memory
    clear_embeddings(batch, also_clear_word_embeddings=True)
```

So my proposal would only touch the second cleaning which would thus just result in the memory not being freed, which would be exactly my goal since I'd like to use the embeddings otherwise.
However, I might miss something.

alanakbik:
Ah yes, you're right! Then this would work - would you like to put in a PR for this?

khituras:
Will do!

alanakbik:
PR was merged a while back!

PR #809, opened 2019-06-17
Allow to keep token embeddings after SequenceTagger.predict()

khituras:
This PR addresses #784. A new parameter `clear_word_embeddings` was added to `SequenceTagger.predict()` which is just connected through to `clear_embeddings` method that already has this parameter.
Since the embeddings are also cleared at the beginning of the `predict` method, there should be no issue with embeddings bleeding through the usage of multiple models on the same sentence/tokens.

I added a small test to show that the new parameter works as intended.

I hopy everything is fine with the PR, let me know!


alanakbik:
@khituras the build was failing due to not passing the code formatting checks performed by black. I've corrected and hopefully it runs through now. Will merge then!

khituras:
Aw, I forgot about that part, I meant to check via black myself. Sorry about that!

alanakbik:
No worries - I'm actually thinking of perhaps removing the formatter again as it seems to be causing confusion.

alanakbik:
:+1:

yosipk:
:+1:

PR #818, opened 2019-06-19
GH-817: trainer modifications

### LINKED ISSUE #796: Disabling test set usage
It seems odd that during training the model is continuously evaluated on both the dev and test set. This is not really the purpose of a test set? Is there any easy way of disabling test set usage, either at the corpus input processing level, or during training. Looking at the [ColumnCorpus](https://github.com/zalandoresearch/flair/blob/master/flair/datasets.py#L20) class it doesn't seem like this is possible but I may have missed something.

### LINKED ISSUE #817: Bad epochs logging counterintuitive
The current logging of the `bad_epochs` parameter is potentially misleading since it always logs the state of bad epochs at the _start_ of each epoch. This is problematic since each line gives us the _results_ of each epoch (i.e. the state at the end of each epoch). 

**Goal**: log `bad_epochs` value as it is at the end of each epoch. 

So then a log line might begin like this: 

```console
33	10:47:51	1	0.0250	2.6336
```

and may be read as "epoch 33 is bad epoch 1 with a loss of 2.6336".

alanakbik:
closes #817 
closes #796 

This PR introduces small fixes to the ModelTrainer class. It introduces a new parameter `monitor_test` that by default is set to `False`, meaning that test set evaluation is no longer performed during training (#796). It also changes the logging behavior of the `bad_epochs` parameter to hopefully be more intuitive (#817).

### DIFF
--- a/flair/trainers/trainer.py
+++ b/flair/trainers/trainer.py
@@ -97,6 +97,7 @@ class ModelTrainer:
         log_test = True if (not param_selection_mode and self.corpus.test) else False
         log_dev = True if not train_with_dev else False
 
+        # prepare loss logging file and set up header
         loss_txt = init_output_file(base_path, "loss.tsv")
         with open(loss_txt, "a") as f:
             f.write(f"EPOCH\tTIMESTAMP\tBAD_EPOCHS\tLEARNING_RATE\tTRAIN_LOSS")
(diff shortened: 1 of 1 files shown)


alanakbik:
:+1:

yosipk:
:+1:

PR #826, opened 2019-06-21
Gh 798 add csv reader

### LINKED ISSUE #798: Corpus reader for CSV text classification files
Many text classification datasets are distributed as simple CSV files in which each row corresponds to a data point and columns correspond to text, labels and other metadata. 

Currently, we require users to first reformat such datasets into FastText format so that the ClassificationCorpus can be read. 

**Feature**: Add a new method for directly loading CSV files for text classification.

alanakbik:
This PR adds another way for loading classification datasets, namely datasets that are in CSV format. You need to pass a column format (like in ColumnCorpus), which indicates which column(s) in the CSV holds the text and which field(s) the label(s).

```python
corpus = CSVClassificationCorpus(
    # path to the data folder containing train / test / dev files
    data_folder='path/to/data',
    # indicates which columns are text and labels
    column_name_map={4: "text", 1: "label_topic", 2: "label_subtopic"},
    # if CSV has a header, you can skip it
    skip_header=True)
```

closes #798

### DIFF
--- a/flair/datasets.py
+++ b/flair/datasets.py
@@ -678,7 +678,7 @@ class TableDataset(FlairDataset):
                     if self.max_chars_per_doc > 0:
                         text = text[: self.max_chars_per_doc]
 
-                    sentence = Sentence(text)
+                    sentence = Sentence(text, use_tokenizer=self.use_tokenizer)
 
                     for column in self.column_name_map:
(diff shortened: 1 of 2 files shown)


alanakbik:
:+1:

yosipk:
:+1:

Issue #736, opened 2019-05-17
CUDA out of memory (again)


emrul:
I previously got CUDA out of memory issues but resolved them by
1. Ensuring no sentence is greater than 200 tokens
2. Updating PyTorch

This worked well with my small dataset (12.5k training sentences) but now I am trying with a bigger dataset (160k sentences) and my CUDA issue is back.

This is my code currently:
`
columns = {0: 'text', 1: 'ner'}
data_folder = 'training/sequences'
corpus: TaggedCorpus = NLPTaskDataFetcher.load_column_corpus(data_folder, columns) print(corpus)
tag_type = 'ner'
tag_dictionary = corpus.make_tag_dictionary(tag_type=tag_type)
print(tag_dictionary.idx2item)
embeddings = StackedEmbeddings(embeddings=[
    WordEmbeddings('glove'),
    PooledFlairEmbeddings('news-forward', pooling='mean', only_capitalized=True, use_cache=True, chars_per_chunk=32),
    PooledFlairEmbeddings('news-backward', pooling='mean', only_capitalized=True, use_cache=True, chars_per_chunk=32),
    WordEmbeddings(embeddings="embeddings/markup2vec.wv")
])

tagger: SequenceTagger = SequenceTagger(hidden_size=128,
                                        embeddings=embeddings,
                                        tag_dictionary=tag_dictionary,
                                        tag_type=tag_type)

trainer: ModelTrainer = ModelTrainer(tagger, corpus)

trainer.train('resources/taggers/v003',
              max_epochs=150,
              mini_batch_size=192,
              eval_mini_batch_size=16,
              checkpoint=True)
`

and this is the error

> RuntimeError: CUDA out of memory. Tried to allocate 156.00 MiB (GPU 0; 24.00 GiB total capacity; 15.79 GiB already allocated; 142.52 MiB free; 3.81 GiB cached)

I've tried using the latest Flair (GitHub Master) and running on Windows 10 with Python 3.7 any PyTorch 1.1.0.  Any ideas on what else I can try to get training to work with a large number of samples?

alanakbik:
Hello @emrul we just now pushed new changes to the master branch that enable streaming data loading using the `Dataset` and `DataLoader` apis. 

If you replace the lines: 

```python
columns = {0: 'text', 1: 'ner'}
data_folder = 'training/sequences'
corpus: TaggedCorpus = NLPTaskDataFetcher.load_column_corpus(data_folder, columns) 
```

with: 
```python
[...]

PR #832, opened 2019-06-24
GH-736: minor bug fixes

alanakbik:
This PR fixes two bugs: 

- CUDA out of memory errors were happening with `FlairEmbeddings` on large sentences due to a cat operation being performed on CUDA. Moved cat to normal memory. This may address #736 

- Error when loading sentences without labels with CSV classification dataset.


yosipk:
👍

alanakbik:
:+1:

Issue #833, opened 2019-06-24
Add option to change the length and hidden size of CharacterEmbedding RNN


yosipk:
For some problems it may be beneficial to increase the length and size of hidden layer of CharacterEmbedding model. It would be cool to be able to modify `char_embedding_dim` and `hidden_size_char` when constructing the model.

alanakbik:
Feature added in PR

PR #834, opened 2019-06-24
GH-833: added parameters to CharacterEmbedding constructor to modify embedding size and hidden state size

yosipk:
Sometimes it is beneficial to use larger character-based word embeddings. This PR changes the constructor of CharacterEmbeddings class to allow that:
```
embedding = CharacterEmbedding(char_embedding_dim=64, hidden_size_char=64)
```
By default both embedding dimension and hidden state size will be 25, as before.


yosipk:
👍

alanakbik:
:+1:

Issue #725, opened 2019-05-15
Flair with Multiple GPU


Dhanachandra:
A clear and concise description of what you want to know.

alanakbik:
Partially answered in #731 so closing this issue for now: For 0.4.2 Multi-GPU support is unlikely, since we have problems with GPU utilisation we want to look into first before adding more GPUs. The only part of the code that is ready for multi-GPU now is the language model training part since it already uses DataLoaders and has 100% GPU utilization. But 0.4.2 will add PyTorch DataLoaders for classification and sequence labeling datasets, so it will prepare multi-GPU support in the future.

PR #848, opened 2019-07-01
GH-725: Added support for multi-gpu in language model training

jasminesjlee:
Hello! I added some code to make training language models compatible with multiple GPUs if they are available. I spent some time doing this so that I could pretrain some models more quickly, but thought I would share it with you as well in case it will help. I saw that there were many issues addressing this but that it has not been implemented yet.

### DIFF
--- a/flair/models/__init__.py
+++ b/flair/models/__init__.py
@@ -1,3 +1,3 @@
 from .sequence_tagger_model import SequenceTagger
-from .language_model import LanguageModel
+from .language_model import LanguageModel, FlairLoss
 from .text_classification_model import TextClassifier
--- a/flair/models/language_model.py
+++ b/flair/models/language_model.py
@@ -60,6 +60,8 @@ class LanguageModel(nn.Module):
         # auto-spawn on GPU if available
         self.to(flair.device)
 
+        self.loss_func = torch.nn.CrossEntropyLoss()
+
     def init_weights(self):
         initrange = 0.1
         self.encoder.weight.detach().uniform_(-initrange, initrange)
@@ -91,6 +93,11 @@ class LanguageModel(nn.Module):
             output,
             hidden,
         )
+
+    def loss(self, data, hidden, targets, ntokens):
+        output, rnn_output, hidden = self.forward(data, hidden)
+        loss = self.loss_func(output.view(-1, ntokens), targets.view(-1)).unsqueeze(0).unsqueeze(0)
+        return loss, hidden
 
     def init_hidden(self, bsz):
         weight = next(self.parameters()).detach()
@@ -364,3 +371,18 @@ class LanguageModel(nn.Module):
         perplexity = math.exp(loss)
 
         return perplexity
+
+
+class FlairLoss(nn.Module):
+    """Container module with an encoder, a recurrent module, and a decoder."""
+
+    def __init__(
+        self,
+        model: LanguageModel
+    ):
+
+        super(FlairLoss, self).__init__()
+        self.model = model
+
+    def forward(self, data, hidden, targets, ntokens):
+        return self.model.loss(data, hidden, targets, ntokens)
--- a/flair/trainers/language_model_trainer.py
+++ b/flair/trainers/language_model_trainer.py
@@ -10,7 +10,7 @@ from torch.optim.sgd import SGD
 
 import flair
 from flair.data import Dictionary
-from flair.models import LanguageModel
+from flair.models import LanguageModel, FlairLoss
 from flair.optim import *
 from flair.training_utils import add_file_handler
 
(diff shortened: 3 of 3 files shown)


kashif:
dear @jasminesjlee thank you for the PR. It seems you are using the `DataParallel` which will only place the model on multiple gpus, however to train with multiple gpu you will need to use `DistributedDataParallel`.

jasminesjlee:
Hello! Can you clarify why? My code also calculates the predictions/loss in parallel.

kashif:
@jasminesjlee Yes you are right, `DataParallel` will chunk the data across the batch dimension and sum the gradients in the backward pass. However i would have thought you would only need to just wrap the `self.model` with it as the approach you have seems a bit too involved... is it beacuse there is no `forward` method in the `FlairModel` ?

jasminesjlee:
There is a `forward` method in the `LanguageModel` class, but it outputs predictions. I did not want to change this, but I still wanted to calculate loss in parallel, so I had to define a class `FlairLoss` that calculates loss in the `forward` method. It would also be possible for me to remove the `FlairLoss` class and instead use the `forward` method of the `LanguageModel` class to output loss, if that is what you prefer!

jbesomi:
Hey,

I believe supporting multi-GPU training is quite important for Flair as it would drastically reduce training time. I plan to use Flair quite extensively in the next few months. I have four GPU (RTX 2080) and I would like to use all of them during training. That said, I would like to help you support multi GPUs, starting from LMs and later on adding support for all task-specific training.

@alanakbik, in #936 you said that "we tend to favor horovod [rather than DataParallel] since this is less invasive to the rest of the code and has more functionality". From my understanding, and by looking at the official [pytorch multi-GPU tutorial](https://pytorch.org/tutorials/beginner/blitz/data_parallel_tutorial.html), for multi-gpu training it suffices to wrap the PyTorch model into a `nn. DataParallel` class. Assuming that works, I believe this is much less invasive than adding an external dependency such as `horovod` (considering many factors, for instance, code maintainability).

@jasminesjlee I don't get what you just wrote in your last comment. Can you please try to explain a bit better? Wouldn't be enough to do something like shown below? Thanks!

(`language_model_trainer.py `)

```
class LanguageModelTrainer:
    def __init__(..., model, ...):
    
        if torch.cuda.device_count() > 1:
           self.model: LanguageModel =  nn.DataParallel(model)
```
        
@alanakbik it would be great to receive support from someone more expert that had to deal with multi-processing in the past. May you please reopen this issue and add a `help wanted` label?

Thanks,
Jonathan

djstrong:
We tried DataParallel but it was not so straightforward. Finally, we only parallelized LSTM: `self.rnn = torch.nn.DataParallel(nn.LSTM(embedding_size, hidden_size, nlayers), dim=1)`.

jbesomi:
Hey @djstrong, thank you for your comment. Can you please add more information? I would love to know more. For instance, who is "we" and why is it not so `straightforward `? Which kind of issues did you get? How big was the gain with `nn.DataParallel`? Moreover, why are you specifying `dim=1` (default is 0)?

alanakbik:
Hello @jbaiter - as it happens, @kishaloyhalder has very recently created a (I believe) working version of the DataParallel solution that you describe. If it turns out to be stable, it would be great to add this to Flair.

djstrong:
@jbesomi It was 3 years ago and I don't remember: https://github.com/djstrong/flair/commit/9683254943e019369d5082d90c8923eb801974fb
The speed gain was not so good, about 50% for two cards.

alanakbik:
I believe @kishaloyhalder got more of an improvement, but it was a very GPU-heady task.

jbesomi:
@alanakbik do you know where we can find @kishaloyhalder multi-GPU code approach?

kishaloyhalder:
Hi @jbesomi , @djstrong 

I have been able to train a `TextClassifier` model with multiple GPUs with Flair using pytorch's DataParallel. However, it is not is a shape to create a PR yet. If your use case overlaps with this, I would be happy to share the code (via a tarball or a new branch maybe) with you.

If you are interested in how it works: Flair uses a custom object called `Sentence` to pass around between trainer and the model. Pytorch's DataParallel assumes all the data would be in [tensor format](https://github.com/pytorch/pytorch/blob/04e7891aab09bda348964c4ae62db66265141936/torch/nn/parallel/scatter_gather.py#L19). It distributes them evenly across all the GPU cards and gathers the losses back. However, in Flair since the data object is not tensor it does not work well with the off-the-shelf DataParallel. I had overridden the `scatter` method in the DataParallel to be compatible with `Sentence` objects.

with regards,
Kishaloy

jbesomi:
Hi @kishaloyhalder,

Thanks, I see. I would be more than interested to have a look at the code and see if it might come in handy to help me implement multi-GPU support for LMs training. For sharing, both tarballs or new branch works. Out of curiosity, how big was the speed gain? Thanks!

PR #853, opened 2019-07-01
GH-852: Add crosslingual MUSE embeddings

### LINKED ISSUE #852: Crosslingual MUSE embeddings
**Multiple languages in one word embedding space**

We often have multilingual data that we want to tag, but training data only for one language (typically English). One interesting solution to this problem is to use word embeddings for multiple languages that have been aligned into the same embedding space. In such _crosslingual embeddings_, the English word 'cat' would have a similar word embedding as the German translation 'Katze'. If we train a classifier or sequence labeler using such crosslingual embeddings, we can train a classifier with English training data and then also use it for German since the word embeddings correspond. 

MUSE embeddings

Facebook's [MUSE](https://github.com/facebookresearch/MUSE) embeddings are exactly this: Embeddings for 30 languages aligned into one embedding space. 

**Goal**: Add MUSE embeddings to Flair with a new `CrosslingualMuseEmbeddings` class. When calling `.embed()` on a sentence, this class needs to (1) perform language identification of the sentence that is being embedded, (2) select the MUSE embedding for the detected language and (3) embed the sentence with this embedding. This will allow us to embed sentences in different languages into the same embedding space. 

We've done some initial testing and this indeed allows us to train models that work for 30 languages with training data for a single languages.

alanakbik:
Closes #852 

Use the new `MuseCrosslingualEmbeddings()` class to embed any sentence in one of 30 languages into the same embedding space. Behind the scenes the class first does language detection of the sentence to be embedded, and then embeds it with the appropriate language embeddings. If you train a classifier or sequence labeler with (only) this class, it will automatically work across all 30 languages, though quality may widely vary. 

Here's how to embed: 
```python
# initialize embeddings
embeddings = MuseCrosslingualEmbeddings()

# two sentences in different languages
sentence_1 = Sentence("This red shoe is new .")
sentence_2 = Sentence("Dieser rote Schuh ist rot .")

# language code is auto-detected
print(sentence_1.get_language_code())
print(sentence_2.get_language_code())

# embed sentences
embeddings.embed([sentence_1, sentence_2])

# print similarities
cos = torch.nn.CosineSimilarity(dim=0, eps=1e-6)
for token_1, token_2 in zip (sentence_1, sentence_2):
    print(f"'{token_1.text}' and '{token_2.text}' similarity: {cos(token_1.embedding, token_2.embedding)}")

```

### DIFF
--- a/flair/data.py
+++ b/flair/data.py
@@ -318,6 +318,7 @@ class Sentence:
         text: str = None,
         use_tokenizer: bool = False,
         labels: Union[List[Label], List[str]] = None,
+        language_code: str = None,
     ):
 
         super(Sentence, self).__init__()
@@ -329,6 +330,8 @@ class Sentence:
             self.add_labels(labels)
 
         self._embeddings: Dict = {}
+
+        self.language_code: str = language_code
 
         # if text is passed, instantiate sentence with tokens (words)
         if text is not None:
@@ -673,6 +676,17 @@ class Sentence:
 
     def __len__(self) -> int:
         return len(self.tokens)
+
+    def get_language_code(self) -> str:
+        if self.language_code is None:
+            import langdetect
+
+            try:
+                self.language_code = langdetect.detect(self.to_plain_string())
+            except:
+                self.language_code = "en"
+
+        return self.language_code
 
 
 class Corpus:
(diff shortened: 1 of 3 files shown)


yosipk:
👍

alanakbik:
:+1:

Issue #856, opened 2019-07-02
Simple TaggedCorpus Deprecation Update


aychang95:
The experiments markdown file _resources/docs/EXPERIMENTS.md_ includes the deprecated "TaggedCorpus" objects, which should be replaced with the more robust "Corpus" object.

 _flair/data_fetcher_ also has most documentation referring to the deprecated "TaggedCorpus" object.

Running any of the experiments will fail with a "cannot import..." error obviously.

Will initialize a pr if no objections

aychang95:
p.s. Experiment breaking commit was at GH-232: Add new Corpus interface over TaggedCorpus and MultiCorpus

alanakbik:
Sure, we'd appreciate a PR!

Issue #232, opened 2018-11-21
Multi-Dataset Training


alanakbik:
Some tasks have multiple datasets available over which we might want to train at the same time. 

For instance, there are currently [6 English treebanks](http://universaldependencies.org/treebanks/en-comparison.html) listed on the UD web page over which we might want to train a PoS tagger for English. Or we might want to train a multilingual PoS tagger for several languages, in which case we would want to simultaneously train over UD corpora for English, German etc.  

**Task**: Add data and training methods to allow for training of models over multiple datasets.

PR #860, opened 2019-07-03
GH-856: Fixed deprecated references to TaggedCorpus in experiments an…

aychang95:
…d data_fetcher.py docs accordingly to the Corpus update that addressed GH-232

Straight forward update so experiment scripts don't break and docs are up to date when referencing the Corpus object.


alanakbik:
Great, thanks for fixing this!

alanakbik:
:+1:

yosipk:
👍

PR #866, opened 2019-07-04
GH-865: keep the same order of sentences in the batch in DocumentRNNEmbedding

### LINKED ISSUE #865: Keeping the order of sentences in the batch in DocumentRNNEmbedding
In DocumentRNNEmbedding the sentences in the batch are permuted (sorted) to take advantage of PyTorch's `pack_padded_sequence` and `pad_packed_sequence` require sentences ordered by length. In the rest of the code sentences are kept in the permuted order, so the outputs will also be permuted. This is fine when the order of the sentences in the batch doesn't matter or when labels are assigned to Sentence objects (like in Flair). However, for some tasks, where association between sentence and label (or an object in general) is not 1-1, it is assumed that order of sentences in the batch does not change.

Therefore, it would be good to keep the order of sentences in the batch. For that we'd need to keep sorting indices so we can undo the permutation due to sorting, and we can do that just after the RNN embeddings are calculated.

yosipk:
1) a fix for keeping the same order of sentences in the batch in DocumentRNNEmbedding, 
2) initializing the DocumentRNNEmbedding in evaluation mode, 
3) a test for 1

closes #865

### DIFF
--- a/flair/embeddings.py
+++ b/flair/embeddings.py
@@ -2122,6 +2122,8 @@ class DocumentRNNEmbeddings(DocumentEmbeddings):
 
         self.to(flair.device)
 
+        self.eval()
+
     @property
     def embedding_length(self) -> int:
         return self.__embedding_length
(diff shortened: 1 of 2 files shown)


alanakbik:
:+1:

yosipk:
👍

Issue #857, opened 2019-07-03
Extremely low GPU ultilization


YangFei1990:
Hello, I was testing the example provided in [TUTORIAL_7_TRAINING_A_MODEL.md](https://github.com/zalandoresearch/flair/blob/v0.4.1/resources/docs/TUTORIAL_7_TRAINING_A_MODEL.md#training-a-sequence-labeling-model) with some small modification. However, I received very low GPU ultilization (around 20%). Also, for the cpu, there are only 1 or 2 cores fully ultilized. 
Here is my code:

```
from flair.data import Corpus 
from flair.data_fetcher import NLPTaskDataFetcher, NLPTask 
from flair.embeddings import TokenEmbeddings, WordEmbeddings, StackedEmbeddings 
from flair.embeddings import CharacterEmbeddings, FlairEmbeddings 
from typing import List

# 1. get the corpus
#corpus: TaggedCorpus = NLPTaskDataFetcher.load_corpus(NLPTask.WNUT).downsample(0.1)
corpus: Corpus = NLPTaskDataFetcher.load_corpus(NLPTask.WNUT_17)
print(corpus)

# 2. what tag do we want to predict?
tag_type = 'ner'

# 3. make the tag dictionary from the corpus
tag_dictionary = corpus.make_tag_dictionary(tag_type=tag_type)
print(tag_dictionary.idx2item)

# 4. initialize embeddings
embedding_types: List[TokenEmbeddings] = [

    WordEmbeddings('glove'),

    # comment in this line to use character embeddings
    CharacterEmbeddings(),

    # comment in these lines to use flair embeddings
    # FlairEmbeddings('news-forward'),
    # FlairEmbeddings('news-backward'),
]

embeddings: StackedEmbeddings = StackedEmbeddings(embeddings=embedding_types)

# 5. initialize sequence tagger
from flair.models import SequenceTagger

tagger: SequenceTagger = SequenceTagger(hidden_size=256,
                                        embeddings=embeddings,
                                        tag_dictionary=tag_dictionary,
                                        tag_type=tag_type,
                                        use_crf=True)

# 6. initialize trainer
from flair.trainers import ModelTrainer

trainer: ModelTrainer = ModelTrainer(tagger, corpus)

# 7. start training
trainer.train('resources/taggers/example-ner',
              learning_rate=0.1,
              mini_batch_size=32,
              max_epochs=150)

# 8. plot training curves (optional)
from flair.visual.training_curves import Plotter
plotter = Plotter()
[...]

PR #891, opened 2019-07-16
GH-857: Memory management and interfaces

alanakbik:
This PR introduces a number of changes to address memory management and slim down some interfaces. 

For **memory management**, it introduces the `embedding_storage_mode` parameter to the `ModelTrainer` class and `predict()` methods. This parameter can be one of 'none', 'cpu' and 'gpu':
- If 'none' all embeddings are deleted after usage - this has lowest memory requirements but means that embeddings need to be recomputed at each epoch of training potentially causing a slowdown. 
- If 'cpu' all embeddings are moved to CPU memory after usage. During training, this means that they only need to be moved back to GPU for the forward pass, and not recomputed so in many cases this is faster, but requires memory. 
- If 'gpu' all embeddings stay on GPU memory after computation. This eliminates memory shuffling during training, causing a speedup. However this option requires enough GPU memory to be available for all embeddings of the dataset. 

To use this option during training, simply set the parameter: 

```python
        # initialize trainer
        trainer.train(
            "path/to/your/model",
            embedding_storage_mode='gpu',
        )
```

In terms of **interfaces**, this PR introduces the `DataPoint` interface that is currently implemented by the `Token` and `Sentence` classes. The interface currently only has a `embedding` property, meaning that each `DataPoint` in Flair needs to be able to get an embedding. We will utilize this interface in the near future to extend Flair to other data types besides words and sentences.

It also slims down the `evaluate()` method in the `flair.nn.Model` interface to take a `DataLoader` instead of a group of parameters.

### DIFF
--- a/flair/data.py
+++ b/flair/data.py
@@ -1,7 +1,7 @@
 from abc import abstractmethod
 from typing import List, Dict, Union
 
-import torch
+import torch, flair
 import logging
 
 from collections import Counter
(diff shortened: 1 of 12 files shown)


yosipk:
👍

alanakbik:
:+1:

Issue #678, opened 2019-04-20
Multi-label classification not working?


gwohlgen:

gwohlgen:
Hi, 
I am trying to make **multi-label classification** work with the **dataset used in this fasttext tutorial:** [https://fasttext.cc/docs/en/supervised-tutorial.html](https://fasttext.cc/docs/en/supervised-tutorial.html).

The problem is, that no matter what embedding used, and what hyperparameter, training always quickly goes towards 0.000 F1 / acc:
**2019-04-20 23:31:15,587 EPOCH 4 done: loss 0.0009 - lr 0.1000 - bad epochs 0
2019-04-20 23:31:30,078 DEV  : loss 0.00073111 - f-score 0.0000 - acc 0.0000
2019-04-20 23:31:44,417 TEST : loss 0.00073590 - f-score 0.0000 - acc 0.000**0

Maybe the problem is that it has a high number of labels, some with low frequency (1)?

Full code and logs here: [https://github.com/gwohlgen/misc/blob/master/classifier__multi-label.ipynb](https://github.com/gwohlgen/misc/blob/master/classifier__multi-label.ipynb)

I split it into train/dev/test, eg 
```
bash$ head cooking.train
wohlg@wohlg-XPS:~/itmo/misc/cooking_classification/preprocessed$ head cooking.train 
__label__sauce __label__cheese how much does potato starch affect a cheese sauce recipe ? 
__label__food-safety __label__acidity dangerous pathogens capable of growing in acidic environments
__label__cast-iron __label__stove how do i cover up the white spots on my cast iron stove ? 
__label__restaurant michelin three star restaurant; but if the chef is not there
__label__knife-skills __label__dicing without knife skills ,  how can i quickly and accurately dice vegetables ? 
__label__storage-method __label__equipment __label__bread what ' s the purpose of a bread box ? 
.....
```

Looks fine.

Then created corpus etc:
```
from flair.data_fetcher import NLPTaskDataFetcher
from flair.embeddings import WordEmbeddings, FlairEmbeddings, DocumentLSTMEmbeddings, CharacterEmbeddings
from flair.models import TextClassifier
from flair.trainers import ModelTrainer
from pathlib import Path

data_path = '/home/wohlg/itmo/misc/cooking_classification/preprocessed'
corpus = NLPTaskDataFetcher.load_classification_corpus(Path(data_path), 
                                                       test_file='cooking.test', 
                                                       dev_file='cooking.valid', 
[...]

PR #908, opened 2019-07-18
GH-678: Data Samplers

alanakbik:
This PR adds a set of Samplers that can be used in the ModelTrainer: 

The `ImbalancedClassificationDatasetSampler` will upsample rare classes and downsample common classes in a classification dataset. It may potentially help with imbalanced datasets. 

The two chunk samplers split a dataset into chunks and shuffle them. This preserves some ordering of the original data while also randomizing the data.

### DIFF
--- a/flair/samplers.py
+++ b/flair/samplers.py
@@ -1,4 +1,5 @@
 import logging
+from collections import defaultdict
 
 from torch.utils.data.sampler import Sampler
 import random, torch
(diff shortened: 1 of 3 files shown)


flair/samplers.py
```diff
@@ -0,0 +1,134 @@
...
+
+        self.block_size = 1
+        self.plus_window = 0
+        self.epoch_count = 0
+
+    def __iter__(self):
+
+        self.epoch_count += 1
+
+        data = [i for i in range(len(self.data_source))]
```
bluesheeptoken: Same as above

alanakbik:
@bluesheeptoken thanks! I've made the changes.

alanakbik:
:+1:

yosipk:
👍

Issue #873, opened 2019-07-11
pytorch-pretrained-bert to pytorch-transformers upgrade


stefan-it:
Hi,

the upcoming 1.0 version of `pytorch-pretrained-bert` will introduce several API changes, new models and even a name change to `pytorch-transformers`.

After the final 1.0 release, `flair` could support 7 different Transformer-based architectures:

* [x] BERT -> `BertEmbeddings`
* [x] OpenAI GPT -> `OpenAIGPTEmbeddings`
* [x] OpenAI GPT-2 -> `OpenAIGPT2Embeddings` 🛡️
* [x] Transformer-XL -> `TransformerXLEmbeddings`
* [x] XLNet -> `XLNetEmbeddings` 🛡️
* [x] XLM -> `XLMEmbeddings` 🛡️
* [x] RoBERTa -> `RoBERTaEmbeddings` 🛡️ (currently not covered by `pytorch-transformers`)

🛡️ indicates a new embedding class for `flair`.

It also introduces an universal API for all models, so quite a few changes in `flair` are necessary so support both old and new embedding classes.

This issue tracks the implementation status for all 6 embedding classes 😊

alanakbik:
Awesome - really look forward to supporting this in Flair!

aychang95:
pytorch-transformers 1.0 was released today: at https://github.com/huggingface/pytorch-transformers 

Migrations summary can be found on the readme [here](https://github.com/huggingface/pytorch-transformers#Migrating-from-pytorch-pretrained-bert-to-pytorch-transformers)
___________
Main takeaways from the migration process are:

- **Models now output tuples**:  Seems like the quickest fix is to retrieve the first element of our model outputs with a safe 0 index
- **Serialization**: Serialization methods have been standardized with the `saved_pretrained(save_directory)`, and models are in evaluation mode by default and must be set to training mode during training
- **Optimizers**: BertAdam and OpenAIAdam are now to be replaced with AdamW, and schedulers are not part of the optimizer anymore.  Scheduler step must be specified per batch and are now standard PyTorch learning schedulers

Otherwise, pytorch-transformers 1.0 looks great and looking forward to use them on flair as well as standalone

alanakbik:
Looks great!

stefan-it:
[...]

PR #914, opened 2019-07-19
GH-873: Add new type of embeddings: XLMEmbeddings

nullphantom:
Added XLMEmbeddings class based on #873


stefan-it:
Thanks for that - I'm currently running per-layer analysis for the (default) English model, but I think it's fine to get merged into my `pytorch-transformers` branch!

stefan-it:
:+1:

Issue #923, opened 2019-07-23
Adding tensorboardX for realtime visualization


felicitywang:
Would you consider adding support for TensorBoard? [tensorboardX](https://github.com/lanpa/tensorboardX) is very easy to use. Or APIs from the trainer to add values to tensorboardX. This would allow visualization for real time or for multiple experiments together.

alanakbik:
This looks interesting. Would you be interested in doing a PR for this?

stefan-it:
Hi,

I made a first draft, implementation is very easy. Since PyTorch 1.1.0 the TensorBoardX implementation is included, so it can be imported with `from torch.utils.tensorboard import SummaryWriter`.

Here's a screenshot of for training a model with Flair:

![image](https://user-images.githubusercontent.com/20651387/61735923-4caa5e00-ad85-11e9-917c-a65f3d152e56.png)

felicitywang:
Thanks for the commit and quick response. Helped with my modifications. 

Not going to create a PR as my needs quite specific and some hard-coded. Pasting my modifications here though in case someone's interested in some reference(tensorboard and config.ini): https://github.com/felicitywang/flair/commit/c2876af28f24a58944f49af155a5468d4910b783


Closing?

alanakbik:
@felicitywang thanks for sharing your modifications. @stefan-it added a first PR for this and we hope this with more tensorboardX features in the future! Closing for now, but feel free to reopen if you have more comments / questions.

PR #924, opened 2019-07-23
GH-923: add support for PyTorch's TensorBoard (ModelTrainer)

stefan-it:
Hi,

PyTorch 1.1.0 has native support for TensorBoard 🎉

This PR adds support for writing a TensorBoard compatible output, that looks like:

image

In order to use TensorBoard during training a model, just use:

```python

trainer: ModelTrainer = ModelTrainer(tagger, corpus, use_tensorboard=True)
```

All logs will be written to a local `./runs` folder.

Just install the latest `tensorboard` version (required to workaround [this refreshing bug](https://github.com/pytorch/pytorch/issues/20477)):

```bash
$ pip3 install tb-nightly
```

Then you can start TensorBoard with:

```bash
$ tensorboard --logdir runs
```

Finally, open `http://localhost:6006` in your webbrowser.

### DIFF
--- a/flair/trainers/trainer.py
+++ b/flair/trainers/trainer.py
@@ -36,6 +36,7 @@ class ModelTrainer:
         loss: float = 10000.0,
         optimizer_state: dict = None,
         scheduler_state: dict = None,
+        use_tensorboard: bool = False,
     ):
         self.model: flair.nn.Model = model
         self.corpus: Corpus = corpus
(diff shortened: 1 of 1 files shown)


alanakbik:
Hey this is great, thanks! We can probably add more structured output in the `ModelTrainer` so that more features of tensorboardX can be used. But we could already merge this PR and add more features in the future?

stefan-it:
I think we can add other features later :)

alanakbik:
:+1:

yosipk:
👍

PR #929, opened 2019-07-25
GH-891: memory storage

alanakbik:
This PR continues refactoring work from #891 to make interfaces more general and allow for 
- additional atomic types of data points (currently only `Token` and `Sentence`) 
- more diverse downstream tasks (currently only `SequenceTagger`, `TextClassifier` and `TextRegressor`)

The PR also includes `embeddings_storage_mode` as a default parameter in the `.evaluate()` routine, and adds the `step` paramter to the `ExpandingChunkSampler` to control the number of epochs before chunk size is increased.

### DIFF
--- a/flair/embeddings.py
+++ b/flair/embeddings.py
@@ -115,7 +115,7 @@ class DocumentEmbeddings(Embeddings):
 class StackedEmbeddings(TokenEmbeddings):
     """A stack of embeddings, used if you need to combine several different embedding types."""
 
-    def __init__(self, embeddings: List[TokenEmbeddings], detach: bool = True):
+    def __init__(self, embeddings: List[TokenEmbeddings]):
         """The constructor takes a list of embeddings to be combined."""
         super().__init__()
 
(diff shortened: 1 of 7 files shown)


yosipk:
👍

alanakbik:
:+1:

BernierCR:
Thanks. However, I'm confused about the effect of embeddings_storage_mode. How does it effect performance in time and memory usage? Would one of those modes use less GPU memory? If I can reduce GPU memory usage, I may be able to train two neural networks on each GPU.

alanakbik:
@BernierCR this is still somewhat work in progress but the general idea is to select between different optimization 'modes' to control the tradeoff between CPU memory usage, CUDA memory usage and execution speed. It all affects if and how embeddings are stored for reuse at the next epoch. 

Currently, we have three storage modes: 

1. '**none**': In this mode, all embeddings are immediately deleted after each mini-batch. This means nothing is held in memory (neither CPU or CUDA). This also means that all word / sentence embeddings need to be generated over and over again at each epoch. So 'none' means: Minimal CPU/CUDA memory, but probably longer compute times. 

2. '**cpu**': In this mode, all embeddings are stored in normal memory. This means more memory usage, but also means that embeddings do not need to be generated anew for each epoch. So runtime may be much faster, but memory requirements will be much higher. For very large datasets, or if you use large embeddings, memory requirements may even me prohibitive. 

3. '**gpu**': In this mode, all embeddings are stored on GPU memory. This only works on either small datasets or if your embeddings are not too large. This is however even faster than 'cpu' because this way, you need not shuffle data between CPU and CUDA - everything is on CUDA.

PR #941, opened 2019-08-01
GH-873: PyTorch-Transformers update

stefan-it:
Hi,

this PR updates the old `pytorch-pretrained-BERT` library to the latest version of `pytorch-transformers` to support various new Transformer-based architectures for embeddings.

A total of 7 (new/updated) embeddings can be used in Flair now:

```python
    BertEmbeddings,
    OpenAIGPTEmbeddings,
    OpenAIGPT2Embeddings,
    TransformerXLEmbeddings,
    XLNetEmbeddings,
    XLMEmbeddings,
    RoBERTaEmbeddings,
)

bert_embeddings = BertEmbeddings()
gpt1_embeddings = OpenAIGPTEmbeddings()
gpt2_embeddings = OpenAIGPT2Embeddings()
txl_embeddings = TransformerXLEmbeddings()
xlnet_embeddings = XLNetEmbeddings()
xlm_embeddings = XLMEmbeddings()
roberta_embeddings = RoBERTaEmbeddings()
```

Detailed benchmarks on the downsampled CoNLL-2003 NER dataset for English can be found in #873 . This PR is the first working attempt to include various new Transformer-based embeddings.

Unit tests can be executed with `pytest --runslow tests`. These unit tests for Transformer embeddings will take ~ 4 minutes using GPU.

### DIFF
--- a/flair/embeddings.py
+++ b/flair/embeddings.py
@@ -11,23 +11,46 @@ import numpy as np
 import torch
 from bpemb import BPEmb
 from deprecated import deprecated
+from torch.nn import ParameterList, Parameter
 
-from pytorch_pretrained_bert import (
+from pytorch_transformers import (
     BertTokenizer,
     BertModel,
     TransfoXLTokenizer,
     TransfoXLModel,
     OpenAIGPTModel,
     OpenAIGPTTokenizer,
+    GPT2Model,
+    GPT2Tokenizer,
+    XLNetTokenizer,
+    XLMTokenizer,
+    XLNetModel,
+    XLMModel,
+    PreTrainedTokenizer,
+    PreTrainedModel,
 )
 
-from pytorch_pretrained_bert.modeling_openai import (
-    PRETRAINED_MODEL_ARCHIVE_MAP as OPENAI_GPT_PRETRAINED_MODEL_ARCHIVE_MAP,
+from pytorch_transformers.modeling_openai import (
+    OPENAI_GPT_PRETRAINED_MODEL_ARCHIVE_MAP as OPENAI_GPT_PRETRAINED_MODEL_ARCHIVE_MAP,
 )
 
-from pytorch_pretrained_bert.modeling_transfo_xl import (
-    PRETRAINED_MODEL_ARCHIVE_MAP as TRANSFORMER_XL_PRETRAINED_MODEL_ARCHIVE_MAP,
+from pytorch_transformers.modeling_gpt2 import (
+    GPT2_PRETRAINED_CONFIG_ARCHIVE_MAP as OPENAI_GPT2_PRETRAINED_MODEL_ARCHIVE_MAP,
 )
+
+
+from pytorch_transformers.modeling_transfo_xl import (
+    TRANSFO_XL_PRETRAINED_MODEL_ARCHIVE_MAP as TRANSFORMER_XL_PRETRAINED_MODEL_ARCHIVE_MAP,
+)
+
+from pytorch_transformers.modeling_xlnet import (
+    XLNET_PRETRAINED_MODEL_ARCHIVE_MAP as XLNET_PRETRAINED_MODEL_ARCHIVE_MAP,
+)
+
+from pytorch_transformers.modeling_xlm import (
+    XLM_PRETRAINED_MODEL_ARCHIVE_MAP as XLM_PRETRAINED_MODEL_ARCHIVE_MAP,
+)
+
 from torch.nn.utils.rnn import pack_padded_sequence, pad_packed_sequence
 
 import flair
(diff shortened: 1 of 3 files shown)


flair/embeddings.py
```diff
@@ -994,6 +1193,149 @@ def __str__(self):
...
+
+    def _add_embeddings_internal(self, sentences: List[Sentence]) -> List[Sentence]:
+        self.model.to(flair.device)
+        self.model.eval()
+
+        with torch.no_grad():
+            for sentence in sentences:
+                for token in sentence.tokens:
+                    token_text = token.text
+                    subwords = self.model.encode(token_text)
```
myleott: One subtle detail here is that the GPT-2 BPE used here generally has leading spaces. Assuming `token_text` does not have a leading space then that could negatively affect the results.

For example:
```python
>>> roberta.encode('Hello world').tolist()
[0, 31414, 232, 2]
>>> roberta.encode(' world').tolist()
[0, 232, 2]
>>> roberta.encode('world').tolist()
[0, 8331, 2]
```

stefan-it: @myleott thanks so much for that hint! I will fix that tokenization bug here :)

Additionally, I found out that this also affects the tokenization routine that is used for the GPT-1 and GPT-2 embeddings. This explains the bad results for all GPT models that I reported in #873.

Update is coming soon :)

flair/embeddings.py
```diff
@@ -994,6 +1193,149 @@ def __str__(self):
...
+
+                    hidden_states = self.model.extract_features(
+                        subwords, return_all_hiddens=True
+                    )
```
myleott: One other issue is that extracting features per token will probably not produce good representations. Ideally one would encode the entire sentence and realign the representations, similar to what's described here: https://explosion.ai/blog/spacy-pytorch-transformers#alignment.

I will be merging some code very soon that will make this very easy with RoBERTa -- you can provide your own tokenized version of the sentence and it will produced aligned features automatically.

stefan-it: @myleott Thanks :) I did a complete re-write of all embedding layers over the last two days. The new strategy is to pass the complete sentence into all models and re-align the embeddings later :)

I'll add some exciting results for NER on CoNLL very soon!

flair/embeddings.py
```diff
@@ -994,6 +1418,140 @@ def __str__(self):
...
+        :param layers: comma-separated list of layers
+        :param pooling_operation: defines pooling operation for subwords
+        :param use_scalar_mix: defines the usage of scalar mix for specified layer(s)
+        """
+        super().__init__()
+
+        try:
+            self.model = torch.hub.load("pytorch/fairseq", model)
+        except:
+            log_line(log)
```
alanakbik: log_line needs to be imported, otherwise this fails.
```python
from flair.training_utils import log_line
```

flair/embeddings.py
```diff
@@ -994,6 +1418,140 @@ def __str__(self):
...
+    def embedding_length(self) -> int:
+        return self.__embedding_length
+
+    def __getstate__(self):
+        # Copy the object's state from self.__dict__ which contains
+        # all our instance attributes. Always use the dict.copy()
+        # method to avoid modifying the original state.
+        state = self.__dict__.copy()
+        # Remove the unpicklable entries.
+        state["model"] = None
```
alanakbik: I think this line does nothing, since "model" is part of "_modules". However, the saved model is still huge, which is strange because in `__setstate__` the RoBERTa model is re-loaded from torch.hub.

flair/embeddings.py
stefan-it: Fixed :) I think I have to change my PyCharm default theme...

flair/embeddings.py
stefan-it: @alanakbik I compared the model sizes for BERT and RoBERTa:

BERT (base): 429MB
RoBERTa (base): 487M

I will check the `state["model"]` now.

However, there's an upcoming PR in the PyTorch-Transformers repo that adds RoBERTa 🔥 So in near future it won't be necessary to use the `torch.hub` wrapper here :)

alanakbik: Ah ok, in this case don't spend too much time on this. It works already, so no need to fix something that will be fixed upstream :)

stefan-it: Great, I'll just leave it as it is now and will update the `RoBERTaEmbeddings` implementation whenever it is available in `pytorch-transformers`!

alanakbik:
:+1:

yosipk:
👍

alanakbik:
Awesome - thank you @stefan-it!!

Hellisotherpeople:
Yes I love this!!!!!

berfubuyukoz:
You are great! @stefan-it Thank you for your generosity!!!

PR #943, opened 2019-08-01
Enhancement to Plotter() visualization

SiewLinYap:
1 enable plots to be displayed when running in both .py and .ipynb
2 display text to notify user the path of the saved plots

Screenshot 2019-08-01 at 2 31 24 PM

### DIFF
--- a/flair/visual/training_curves.py
+++ b/flair/visual/training_curves.py
@@ -9,9 +9,18 @@ import csv
 import matplotlib
 import math
 
-matplotlib.use("Agg")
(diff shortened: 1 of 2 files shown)


alanakbik:
@SiewLinYap thanks for adding this. Looks like the build is failing because `IPython` is not found. Could you add the from `IPython` dependency to `requirements.txt` and push an update?

alanakbik:
Thanks!

alanakbik:
:+1:

yosipk:
👍

PR #956, opened 2019-08-04
GH-940: Hashtags in corpus

### LINKED ISSUE #940: HashTag ignored for Column Corpus
**Describe the bug**
When reading from a ColumnCorpus, lines with hash tags are ignored. This may bring problems to tasks such as Wnut17(there are hash tags at start of words.)
See https://github.com/zalandoresearch/flair/blob/b0e4cc1a92c4bd56536ed4e09c8ff6cb1a89f08c/flair/datasets.py#L424

**To Reproduce**
Run NER experiment with Wnut17.

**Expected behavior**
Hash-tag-starting words should be kept.

 - OS [e.g. iOS, Linux]: linux
 - Version [e.g. flair-0.3.2]: flair-0.4.2

**Additional context**

alanakbik:
Closes #940 

Fixes the problem that in certain column formats, notably WNUT-17, lines that begin with a hash tag are not in fact comments, but words that begin with a hash tag.

### DIFF
--- a/flair/datasets.py
+++ b/flair/datasets.py
@@ -1070,7 +1070,11 @@ class GERMEVAL(ColumnCorpus):
             )
             log.warning("-" * 100)
         super(GERMEVAL, self).__init__(
-            data_folder, columns, tag_to_bioes=tag_to_bioes, in_memory=in_memory
+            data_folder,
+            columns,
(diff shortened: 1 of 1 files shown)


alanakbik:
:+1:

yosipk:
👍

PR #962, opened 2019-08-05
GH-960: make num workers a parameter

### LINKED ISSUE #960: the num_worker parameterLanguageModelTrainer
the num_worker parameters in LanguageModelTrainer seems to be hard coded.

it would be great to set it as a parameter of the LanguageModelTrainer class.

**Additional context**
I am currently fine tuning  FlairEmbeddings on a relatively small corpus ~30 M words which fits on memory (so I use a single split). 
For some reason the datasets seems to be preprocessed between each epoch. As a results training takes more time between each Epoch than the actual training time of a single epoch.  Any advice on this particular scenario would be helpful.

Btw great work with this package it is very helpful 👍 If you need any help, I would be happy to contribute.

alanakbik:
Closes #960 

You can now set the number of workers in the `LanguageModelTrainer`, like so: 

```python
# train your language model

trainer.train(
    "resources/LMs/example-lm",
    # sequence_length should normally be around 200-300
    sequence_length=256,
    # set the number of workers
    num_workers=12,
)
```

### DIFF
--- a/flair/trainers/language_model_trainer.py
+++ b/flair/trainers/language_model_trainer.py
@@ -236,7 +236,6 @@ class LanguageModelTrainer:
 
         self.loss_function = torch.nn.CrossEntropyLoss()
         self.log_interval = 100
-        self.num_workers = 2
         self.epoch = epoch
         self.split = split
         self.loss = loss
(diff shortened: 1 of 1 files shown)


yosipk:
👍

alanakbik:
:+1:

PR #991, opened 2019-08-12
GH-981: fix max_tokens_per_doc in ClassificationCorpus

### LINKED ISSUE #981: Truncating sentences with max_tokens_per_doc when in_memory=False
It appears that argument `max_tokens_per_doc` has no effect when creating a `ClassificationCorpus` if `in_memory=False` (which is the default for `ClassificationCorpus`).

Looking at `ClassificationDataset` in `datasets.py`, I see that `max_tokens_per_doc` is not used when `self.in_memory` is False.

I also get an error due to long lines with the following code:

```python
from pathlib import Path 
from flair.datasets import ClassificationCorpus
from flair.embeddings import DocumentRNNEmbeddings, BertEmbeddings

data_folder = Path('data_fasttext_format') # some of the sentences have long lines
corpus: ClassificationCorpus = ClassificationCorpus(data_folder,train_file='train.txt',dev_file='val.txt',test_file='val.txt', max_tokens_per_doc=20)
label_dict = corpus.make_label_dictionary()                                     
embs = [BertEmbeddings()]
document_embeddings: DocumentRNNEmbeddings = DocumentRNNEmbeddings(embs, rnn_type='LSTM')
classifier = TextClassifier(document_embeddings, label_dictionary=label_dict, multi_label=False)
trainer.train('data_fasttext_format')
```

This results in error ``RuntimeError: index out of range at /pytorch/aten/src/TH/generic/THTensorEvenMoreMath.cpp:191``. However, when I use `in_memory=True` when creating the `ClassificationDataset`, the sentences are truncated as expected and there are no errors.

Enviroment:
- Linux
- Python 3.6.2
- flair-0.4.2

alanakbik:
Closes #981

### DIFF
--- a/flair/datasets.py
+++ b/flair/datasets.py
@@ -810,6 +810,7 @@ class ClassificationDataset(FlairDataset):
 
         self.total_sentence_count: int = 0
         self.max_chars_per_doc = max_chars_per_doc
+        self.max_tokens_per_doc = max_tokens_per_doc
 
         self.path_to_file = path_to_file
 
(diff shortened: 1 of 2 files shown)


yosipk:
👍

alanakbik:
:+1:

PR #993, opened 2019-08-13
PyTorch-Transformers: add possibility to pass model path to embeddings

stefan-it:
Hi,

this PR adds the possibility to pass a model path to the PyTorch-Transformer based embeddings.

This is highly needed when Flair users want to use own trained models, like e.g. an own trained XLNet model.

The argument name is changed from `model` to `pretrained_model_name_or_path` (as it is named in `pytorch-transformers`).

### DIFF
--- a/flair/embeddings.py
+++ b/flair/embeddings.py
@@ -1170,25 +1170,22 @@ def _get_transformer_sentence_embeddings(
 class TransformerXLEmbeddings(TokenEmbeddings):
     def __init__(
         self,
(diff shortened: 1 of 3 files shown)


stefan-it:
Btw: an extra section that shows how to use an own trained model is coming very soon (it shows how to use sciBERT with `flair`) 😅

alanakbik:
Awesome, thanks for adding this!

alanakbik:
:+1:

yosipk:
👍

PR #996, opened 2019-08-14
Simplified sequence tagger

alanakbik:
Adds a number of simplifications and bug fixes for the `SequenceTagger`: 

- Fixes `SequenceTagger` with `use_rnn=False` which previously threw shape error
- `SequenceTagger` no longer sorts sentences by length during forward pass since as of torch 1.1. `pack_padded_sequence` no longer requires sorting
- bumpts min torch version to 1.1 so that `pack_padded_sequence` with `enforce_sorted=False` works
-  removes unneeded code which somehow was still in there
- minor cleanup in logging


alanakbik:
:+1:

yosipk:
👍

PR #1020, opened 2019-08-22
Fixed logging unicode encoding error.

eurekaqq:
I fixed [issue#718](https://github.com/zalandoresearch/flair/issues/718), I think the problem is that the log file is not opened with 'utf-8'.


alanakbik:
@eurekaqq thanks for fixing this!

alanakbik:
:+1:

yosipk:
👍

PR #1022, opened 2019-08-22
Make tokenisation modular (easy integration of 3rd party lib)

pommedeterresautee:
replace the `use_tokenizer` parameter in `Sentence` class by a `tokenizer` parameter where a custom tokenizer can be provided.
The `tokenizer` function signature is simple: take a string and return a list of `Token`.

The idea is to let users decide what they want (specialized tokenization, etc.) and still provide them some basic options.
Space tokenizer (the space split when `use_tokenizer` is set to False) and `segtok` are provided as basic options. 
More may come in the future without having to make modifications.

I hope the API is clear. The only drawback I see is that several functions have their signatures modified. Hopefully, the new API is general enough to not require another change in the future regarding tokenisation, it is still version 0.x so it seems to be acceptable to perform the modification now. Moreover, I don't think a modular approach is possible with the current signature.

Related to https://github.com/zalandoresearch/flair/pull/640 , https://github.com/zalandoresearch/flair/issues/876 , and https://github.com/zalandoresearch/flair/issues/563

FWIW, the approach is inspired from https://github.com/dselivanov/text2vec which is the main NLP lib in `R` world. I use it a lot since quite some time and never found myself limited by this approach.


flair/data.py
```diff
@@ -337,19 +337,139 @@ def __repr__(self) -> str:
         )
 
 
+def space_tokenizer(text: str) -> List[Token]:
+    """
+    Tokenizer based on space character only.
+    """
+    tokens: List[Token] = list()
```
bluesheeptoken: In the code, the convention `[]` is used, maybe we should keep it that way for consistency ?

flair/data.py
```diff
@@ -337,19 +337,139 @@ def __repr__(self) -> str:
...
+    word = ""
+    index = -1
+    for index, char in enumerate(text):
+        if char == " ":
+            if len(word) > 0:
```
bluesheeptoken: `if word` instead ?

flair/data.py
```diff
@@ -337,19 +337,139 @@ def __repr__(self) -> str:
...
+                        text=word, start_position=start_position, whitespace_after=True
+                    )
+                )
+
+            word = ""
+        else:
+            word += char
+    # increment for last token in sentence if not followed by whitespace
+    index += 1
+    if len(word) > 0:
```
bluesheeptoken: `if word`

flair/data.py
```diff
@@ -337,19 +337,139 @@ def __repr__(self) -> str:
...
+        )
+    return tokens
+
+
+def segtok_tokenizer(text: str) -> List[Token]:
+    """
+    Tokenizer using segtok, a third party library dedicated to rules-based Indo-European languages.
+    https://github.com/fnl/segtok
+    """
+    tokens: List[Token] = list()
```
bluesheeptoken: `[]`

flair/data.py
```diff
@@ -337,19 +337,139 @@ def __repr__(self) -> str:
...
+
+    words: List[str] = list()
```
bluesheeptoken: `[]`

flair/data.py
```diff
@@ -337,19 +337,139 @@ def __repr__(self) -> str:
...
+                current_offset + 1 if current_offset > 0 else current_offset
+            )
+
+        token = Token(text=word, start_position=start_position, whitespace_after=True)
+        tokens.append(token)
+
+        if (previous_token is not None) and word_offset - 1 == previous_word_offset:
+            previous_token.whitespace_after = False
+
+        word_len = len(word)
```
bluesheeptoken: No need to assign a new variable for `len(word)`

flair/data.py
```diff
@@ -337,19 +337,139 @@ def __repr__(self) -> str:
...
+        raise ImportError(
+            "Please install Spacy v2.0 or better before using the Spacy tokenizer, otherwise you can use segtok_tokenizer as advanced tokenizer."
+        )
+
+    model: Language = model
+
+    def tokenizer(text: str) -> List[Token]:
+        doc: Doc = model.make_doc(text)
+        previous_token = None
+        tokens: List[Token] = list()
```
bluesheeptoken: `[]`

pommedeterresautee:
@bluesheeptoken Tks a lot for this review. 

Which lines should be deduplicated / refactored? 
For the tokenizer factories, there are some duplicated lines to make each of them independent and more easy to read / reproduce for other tokenizers not yet supported without having to understand anything else but the function signature, so it's a choice.  If the duplication is inside the Sentence class, let me know where it is.

bluesheeptoken:
@pommedeterresautee Thanks for replying ! I was spoken about this, if it is done on purpose, that is fine to me :)

flair/data.py
pommedeterresautee: Is it more readable? Shorter but it requires to know some Python convention. I have not found any other occurrence of this pattern but I may have missed something.

pommedeterresautee:
@bluesheeptoken have switched to Go a few months ago, out there readiness is above everything else :-)
*"[A little copying is better than a little dependency.](https://go-proverbs.github.io/)"*

alanakbik:
@pommedeterresautee this looks great, thanks! Will do some testing and merge soon!

bluesheeptoken:
@pommedeterresautee, For the check of empty lists, the "pythonic way" to check is `if l:`. This is preferred due to the fact this is more readable (relatable I guess, but `if word` is not misleading, we expect a string) and faster (does not need to calculate the full length of the list).

Anyway, this is not really important here I guess, since I have seen more occurences of `if len(l) == 0:` than `if l:`

Thanks for the Pull Request ! :)

resources/docs/TUTORIAL_1_BASICS.md
```diff
@@ -83,6 +84,18 @@ This should print:
 Sentence: "The grass is green ." - 5 Tokens
 ```
 
+You can write and provide your own wrapper around the tokenizer you want to use.
```
alanakbik: After the PR is merged, this text will be on the web page, but most users install flair through pip, i.e. they will not have access to this feature. Perhaps point out in the text that this is a feature only available on master branch currently.

flair/data.py
```diff
@@ -357,19 +357,138 @@ def __repr__(self) -> str:
...
     """
-    A Sentence is a list of Tokens and is used to represent a sentence or text fragment.
+       A Sentence is a list of Tokens and is used to represent a sentence or text fragment.
     """
 
     def __init__(
         self,
         text: str = None,
-        use_tokenizer: bool = False,
+        tokenizer: Callable[[str], List[Token]] = space_tokenizer,
```
alanakbik: This breaks backwards compatibility to downstream code that uses Flair with earlier versions. Perhaps the signature could be changed to: 
```python
    def __init__(
        self,
        text: str = None,
        use_tokenizer: Union[bool, Callable[[str], List[Token]]] = space_tokenizer,
        labels: Union[List[Label], List[str]] = None,
        language_code: str = None,
    ):
```

i.e. call it `use_tokenizer` instead of `tokenizer` and allow passing a bool in addition to callable. Then, a few lines down one could add: 

```python
        tokenizer = use_tokenizer
        if type(use_tokenizer) == bool:
            tokenizer = segtok_tokenizer if use_tokenizer else space_tokenizer
```

i.e. by default if a callable is passed `tokenizer = use_tokenizer` but if a bool is passed the tokenizer is instead initialized with the earlier behavior, i.e. `tokenizer = segtok_tokenizer` if `use_tokenizer=True`.

flair/data.py
```diff
@@ -384,62 +503,7 @@ def __init__(
...
-
-                        word = ""
-                    else:
-                        word += char
-                # increment for last token in sentence if not followed by whtespace
-                index += 1
-                if len(word) > 0:
-                    token = Token(word, start_position=index - len(word))
-                    self.add_token(token)
+            [self.add_token(token) for token in tokenizer(text)]
```
alanakbik: This is great :)

alanakbik:
@pommedeterresautee thanks again for this great PR! I've put some annotation inline - I wonder if it can be adapted to preserve backwards compatibility so that the original tokenization instructions in the online tutorial still remain valid and using a different tokenizer becomes a case of "opt-in complexity". Otherwise this would cause master branch (and visible documentation online) to diverge from the last Flair release.

pommedeterresautee:
You are right about the retro compatibility, it would be disturbing for many users to break API. 
However the API would have strange signature and difficult to clean in the future. What about adding back `tokenizer` as bool only, and if someone set it to True, segtok is used plus a warning indicating it s deprecated? So in the future it can be safely removed?

alanakbik:
Do you mean adding back `use_tokenizer` as an additional variable and having the two side by side? Yes, that would also be an option. 

For now I think a deprecation warning may not be necessary since I think many users are ok with the default tokenization, so simply having a boolean `use_tokenizer` option might make it easier for many users to get started (this is also one import statement less). So this is something we might want to keep, while also offering those users who want to switch out tokenizers a good way to do so. Either having two separate variables `use_tokenizer` and `tokenizer`, or one `use_tokenizer` variable that is type-overloaded would achieve this, so either way is good for me. What do you think?

pommedeterresautee:
Tks for your ideas.
I get your point, I will stick to your original original proposition of having use_tokenizer being both bool and a function.

alanakbik:
:+1:

adizdari:
👍

kashif:
👍

alanakbik:
Thanks a lot @pommedeterresautee for this PR!!

PR #1024, opened 2019-08-22
embeddings: deprecate ELMoTransformerEmbeddings

stefan-it:
Hi,

the loading/saving logic for Flair models changed quite a while ago, so it is no longer possible to use another pickle framework. The `ELMoTransformerEmbeddings` could only be loaded and saved using `dill`.

This PR temporarily deprecates the `ELMoTransformerEmbeddings` implementation until a better solution is found!


alanakbik:
:+1:

yosipk:
👍

Issue #718, opened 2019-05-12
Language model training: Logging error for non-latin language


ali-h:
**Describe the bug**
During the language model training, flair regularly logs sentences generated from the model. In my case, this raises the following error after every epoch:
```
--- Logging error ---
Traceback (most recent call last):
  File "...\torch\lib\logging\__init__.py", line 1037, in emit
    stream.write(msg + self.terminator)
  File "...\torch\lib\encodings\cp1252.py", line 19, in encode
    return codecs.charmap_encode(input,self.errors,encoding_table)[0]
UnicodeEncodeError: 'charmap' codec can't encode characters in position 28-31: character maps to <undefined>
Call stack:
  File "...flair_lm.py", line 33, in <module>
  File "...\torch\lib\site-packages\flair\trainers\language_model_trainer.py", line 336, in train
    log.info(self.model.generate_text())
Message: (...)
Arguments: ()
```

Note that this doesn't interrupt the training.

**To Reproduce**
Roughly, here's my code:
```python
if __name__ == '__main__':
    # are you training a forward or backward LM?
    is_forward_lm = True

    # load the default character dictionary
    dictionary: Dictionary = Dictionary.load_from_file('chars.pickle')

    # get your corpus, process forward and at the character level
    corpus = TextCorpus(Path('corpus'),
                        dictionary,
                        is_forward_lm,
                        random_case_flip=False,
                        character_level=True)

    # instantiate your language model, set hidden size and number of layers
    language_model = LanguageModel(dictionary,
                                   is_forward_lm,
                                   hidden_size=512,
                                   nlayers=1)

    # train your language model
    trainer = LanguageModelTrainer(language_model, corpus)

    trainer.train(Path('lm'),
                  sequence_length=250,
                  mini_batch_size=10,
                  max_epochs=10)
```

**Environment (please complete the following information):**
 - OS : Windows 10
 - Version : 0.4.1
 - Python: 3.7

**Additional context**
My corpus files are UTF-8 encoded Arabic text files,

alanakbik:
@ali-h thanks for reporting this and for providing the code snippet to reproduce the error! Could you also give us your `chars.pickle` file and a small text file that throws this error? If we can reproduce we can probably fix this.

khoaipx:
I have same issue when building Vietnamese language model. 

```UnicodeEncodeError: 'charmap' codec can't encode character '\u1ed1' in position 34: character maps to <undefined>```

[...]

PR #1025, opened 2019-08-23
Fixed unicode error when terminal encoding is not utf-8

eurekaqq:
Hello, about [issue #718](https://github.com/zalandoresearch/flair/issues/718), I did more tests.

On my Ubuntu computer, if terminal is not working with utf-8, it may get same problem.

Therefore, I make function `generate_text` output text encode to utf-8 in `language_model.py`.
And then, it will work well at runtime, but the log file will been written a lot of unicodes,
like below,
`(b'\n \xca\x95\xe6\x99\x82  y \xe7\xaa\xa0  \xe5\x85\xb1\xe5\x9c\x8b \xe9\x80\xa0 \xe5\x9b\x9b \xe8\xaa\x9e \xf0\x90\xad\xae \xe9\xab\x98\xe5\xad\xb8\xe6\x96\xbc \xe7\xb4\x80 \xe7\xbd\xbe \xe5\x8c\x97 \xe5\x85\x8b \xe9\x80\x8d\xe5\x85\x8b ah\xe5\x8f\x83 ..., 7.52521240234375)`

If you want to write correct word in log file.
You should do below,

```
export LC_ALL=C.UTF-8
export LANG=C.UTF-8
```

If I find a better solution, I will fix it.

### DIFF
--- a/flair/models/language_model.py
+++ b/flair/models/language_model.py
@@ -331,6 +331,8 @@ class LanguageModel(nn.Module):
             if not self.is_forward_lm:
                 text = text[::-1]
 
+            text = text.encode('utf-8')
+
             return text, log_prob
 
     def calculate_perplexity(self, text: str) -> float:


alanakbik:
Great, thanks!

alanakbik:
:+1:

yosipk:
👍

PR #1016, opened 2019-08-21
GH-1012: make all tag prediction optional to increase speed

alanakbik:
Closes #1012 

Increases `SequenceTagger` prediction speed by making it optional (and by default not activated) to predict all tags instead of only the most likely one. Adds the new boolean argument `all_tag_prob` to the `predict()` method to control this behavior (if set to `True`, predict all).

PR #1026, opened 2019-08-23
[NER] fix inference bug introduced by #1016 (tags are not saved in token object)

pommedeterresautee:
PR #1016 has introduced a new parameter to avoid computing scores for each tag and accelerating inference.
Now, when `all_tag_prob=False` during NER inference, predictions are not saved in `Token.tags`... so it is like nothing happened.
It is because of the `zip()` function taking multiple lists but if they have different sizes, the length of the shortest is taken as reference. When `all_tag_prob` is `False`, `all_tags` list is empty, so `zip` function acts like all of its parameters/lists are empty, and no iteration happens in the for loop saving results in `Token` objects.

It is a tricky bug as when you test the same sentence with `all_tag_prob=True` and then `all_tag_prob=False` the inference is working... I have not understood why.

The fix is to simply divide operations in 2 parts without any call to `zip`.

### DIFF
--- a/flair/models/sequence_tagger_model.py
+++ b/flair/models/sequence_tagger_model.py
@@ -371,13 +371,18 @@ class SequenceTagger(flair.nn.Model):
                         feature, batch, get_all_tags=all_tag_prob
                     )
 
-                for (sentence, sent_tags, sent_all_tags) in zip(batch, tags, all_tags):
-                    for (token, tag, token_all_tags) in zip(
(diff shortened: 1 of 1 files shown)


alanakbik:
@pommedeterresautee thanks for spotting and fixing this!

alanakbik:
:+1:

yosipk:
👍

PR #1030, opened 2019-08-24
[Sentence] Fix infinite loop bug

pommedeterresautee:
Sometimes, `start_pos` in `Token` is set to None.
This is for instance the case when loading a `Corpus` in columnar format.

In those cases `while t.start_pos != pos` from `to_original_text` method is always True.
This fix manages this case.

Another solution would be to just throw an Exception. Let me know what you prefer. In this case it may be good to forbid to have `start_pos` being `None` in any case and to rewrite `Corpus` class. May be to put in another PR

### DIFF
--- a/flair/data.py
+++ b/flair/data.py
@@ -299,7 +299,7 @@ class Span:
     def to_original_text(self) -> str:
         pos = self.tokens[0].start_pos
         if pos is None:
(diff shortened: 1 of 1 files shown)


alanakbik:
:+1:

alanakbik:
thanks for fixing this!

PR #1031, opened 2019-08-25
Improve NER Viz

pommedeterresautee:
* each sentence is put in its own paragraph making the HTML easier to read (before all sentences were put one after the other, part of a large single block of text)
* simplify API by removing unnecessary parameter (now color dict does both list of tags and colors)
* clean code:
  * put all HTML viz related things in the same file
  * replace hard coded tags and colour codes inside the function body by default parameters

Before (one block)
image

After (each sentence is in its own paragraph for a more pleasing manual check)
image

(I have manually redacted some parts in red)


alanakbik:
@pommedeterresautee thanks for refactoring this - looks great!

alanakbik:
:+1:

yosipk:
👍

PR #1038, opened 2019-08-26
[NER] improve inference perf by > 40% on GPU when CRF is used

pommedeterresautee:
Viterbi seems to be the bottle neck of the inference part when GPU is used with CRF.

Here we simply move the computation of this part from GPU to CPU as the operations are not GPU optimized.
On my 2080TI with I7 (6 real cores), on the same data / same model, I went from 1mn 33s to 63 seconds inference time.
I have not made any test on other hardware but I expect perf improvement to be similar.

Moreover, this open the door to more optimizations as it removes the "lock" on GPU (we can easily do a better job at multithreading the process than Pytorch is currently doing), plus the same performance boost can probably be reproduced on the training part.

### DIFF
--- a/flair/models/language_model.py
+++ b/flair/models/language_model.py
@@ -331,7 +331,7 @@ class LanguageModel(nn.Module):
             if not self.is_forward_lm:
                 text = text[::-1]
 
-            text = text.encode('utf-8')
+            text = text.encode("utf-8")
 
             return text, log_prob
 
--- a/flair/models/sequence_tagger_model.py
+++ b/flair/models/sequence_tagger_model.py
@@ -535,7 +535,10 @@ class SequenceTagger(flair.nn.Model):
             return score
 
     def _obtain_labels(
-        self, feature: torch.Tensor, sentences: List[Sentence], get_all_tags: bool = False
+        self,
+        feature: torch.Tensor,
+        sentences: List[Sentence],
+        get_all_tags: bool = False,
     ) -> (List[List[Label]], List[List[List[Label]]]):
         """
         Returns a tuple of two lists:
--- a/flair/models/sequence_tagger_model.py
+++ b/flair/models/sequence_tagger_model.py
@@ -548,10 +548,12 @@ class SequenceTagger(flair.nn.Model):
 
         tags = []
         all_tags = []
-        for feats, length in zip(feature, lengths):
+        feature_cpu = feature.detach().to('cpu')
+        transitions_cpu = self.transitions.detach().to('cpu')
+        for feats, length in zip(feature_cpu, lengths):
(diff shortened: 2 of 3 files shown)


pommedeterresautee:
I have made some tests to move the whole CRF part on the CPU (including for the training), keeping the featurisation on GPU, and it brings no improvement at all for the training part. 

I still think there are performance opportunities remaining as the GPU use is low and the inference is still not multithreaded at its best. 

One strategy I want to try now that the Viterbi part is on CPU is multithreading the process at the batch level. Right now the Viterbi algo is applied serially to each sentence, however the features are extracted by the GPU in batch. Before this PR there was a GPU lock, but now Viterbi is CPU only it should be doable to distribute the computation on each CPU core.

alanakbik:
Hello @pommedeterresautee thanks a lot for this PR - we've unhappy with the performance of the CRF decoder for a while now so this really helps! I've run a test on CoNLL-03 predicting with the current 'ner' model on a v100 and prediction time decreases **from ~20 to ~15 seconds** on the evaluation dataset and **from ~95 to ~60 seconds** on all sentences in CoNLL-03. This is great! 

I'll run a few more tests on some other datasets to make sure that we get the same eval numbers and will merge into master!

stefan-it:
Not directly related to this issue: but I've seen some F-Score mismatches between Flair output and the CoNLL script (maybe we can switch back to the conll script, or at least with a configurable parameter 🤔) But I'll collect some examples and open a new issue for that!

alanakbik:
@stefan-it yes good point - Flair uses a slightly different logic to get from tokens to spans which should account for the difference. But this is something we've been meaning to add back in so it's good to put this back on the agenda.

alanakbik:
@pommedeterresautee the following code is throwing an error: 

```python
tagger = SequenceTagger.load("frame-fast")
```

This likely affects pre-trained models that were trained without a CRF. I'll check into this a bit more.

alanakbik:
I think the problem is that `self.transitions` is now always assumed to be there, but sometimes models are trained with `use_crf` set to `False`, which then throws an error. To reprodue: 

```python
tagger = SequenceTagger.load("frame-fast")
tagger.predict(Sentence("I love Berlin"))
```

@pommedeterresautee could you take a look?

pommedeterresautee:
Tks, will check and fix.

pommedeterresautee:
You were right about the error, self.transitions is the culprit, to remove the error it would require to retrain every model. The commit avoid such behaviour, now the matrix is created only when required.

alanakbik:
Thanks! Looks good!

alanakbik:
:+1:

yosipk:
👍

PR #1047, opened 2019-08-30
Fix transpose_() to transpose()

roy29fuku:
Hi, 

transpose_() is in-place version of transpose()
https://pytorch.org/docs/stable/tensors.html#torch.Tensor.transpose
You don't need asign althogh it makes no errors.

I reccomend use transpose() as like other files in flair.


alanakbik:
@roy29fuku thanks for correcting this!

alanakbik:
:+1:

yosipk:
👍

PR #1066, opened 2019-09-04
Simplified similarity learner

yosipk:
Added two new types of `DataPoint`:
* `DataPair` type encapsulates a pair of `DataPoint` objects.
* `Image` type represents an image. 

Added three new corpora: 
* `FeideggerCorpus` is a corpus of image-sentence pairs, 
* `ParallelTextCorpus` is a general corpus of sentence-sentence pairs,
* `OpusParallelTextCorpus` is a small parallel English-German parallel sentence corpus.

Added `ImageEmbedding` and a couple of derived classes:
* `IdentityImageEmbeddings` embeds the image into a image tensor (3xHxW),
* `PrecomputedImageEmbeddings` embeds the image into a precomputed embedding tensor,
* `NetworkImageEmbeddings` embeds an image into a tensor via neural net; 
currently supported `resnet50` and `mobilenet_v2` with weights pre-trained on ImageNet via `torchvision`.

Added `SimilarityLearner` class. The constructor takes:
* two input embeddings (`source_embeddings` and `target_embeddings`)
* two input adaptors (`source_mapping` and `target_mapping`)
* similarity measure (`similarity_measure` of type `SimilarityMeasure`)
* similarity loss (`similarity_loss` of type `SimilarityLoss`)
The class takes two input embeddings, adapts them if necessary, computes the similarity between all pairs in the batch, and applies the loss. 
 
Two `SimilarityMeasure` types are implemented:
* `ModelSimilarity` takes a (possibly adapted) source embedding and outputs a model which is used to produce a similarity score given (possibly adapted) target embedding
* `CosineSimilarity` computes the cosine similarity between two (possibly adapted) input embeddings

Two `SimilarityLoss` types are implemented, both take similarity scores between all input pairs and corresponding labels:
* `PairwiseBCELoss` computes the binary cross entropy loss for each pair, with option to balance the positives and negatives (`balance` option in the constructor)
* `RankingLoss` computes the triplet ranking loss with a margin for each triplet (pivot, positive, negative), where triplets are formed from all pairs. Margin is given by `margin` option in the constructor. 

Training is performed on all pairs _from the current batch_, but the validation is performed on _all validation pairs_. 

An example of learning to align English and German sentences from Opus parallel corpus:
```python
import torch

import flair
from flair.datasets import OpusParallelCorpus
from flair.embeddings import DocumentRNNEmbeddings, BytePairEmbeddings
from flair.models.similarity_learning_model import SimilarityLearner, CosineSimilarity, RankingLoss

# instantiate parallel corpus
corpus = OpusParallelCorpus('tatoeba', 'de', 'en', in_memory=False)

print(corpus)
print(corpus.train[0])

multilingual_embedding = DocumentRNNEmbeddings(
    [
        BytePairEmbeddings('multi'),
    ],
    bidirectional=True,
    dropout=0.25,
    hidden_size=256,
)

source_embedding = multilingual_embedding
target_embedding = multilingual_embedding

similarity_measure = CosineSimilarity()

similarity_loss = RankingLoss(margin=0.15)

similarity_model = SimilarityLearner(source_embeddings=source_embedding,
                                     target_embeddings=target_embedding,
                                     similarity_measure=similarity_measure,
                                     similarity_loss=similarity_loss)

print(similarity_model)

trainer: ModelTrainer = ModelTrainer(similarity_model, corpus, optimizer=torch.optim.SGD)

trainer.train(
    f'opus-similarity-cosine-ranking',
    learning_rate=2,
    mini_batch_size=128,
    max_epochs=1000,
    min_learning_rate=1e-6,
    shuffle=True,
    anneal_factor=0.5,
    patience=4,
    num_workers=0,
    embeddings_storage_mode='none'
)
```


alanakbik:
:+1:

yosipk:
👍

djstrong:
There is a problem with `RankingLoss` if in the mini-batch all pairs are positive with each other (i.e. there is one DataPoint in mini-batch). Then loss is `nan`.

yosipk:
Hey @djstrong, you're right: the ranking loss is not defined in case when all labels in the batch are the same. It's true that in that case we could just set it to 0 rather than `nan`, but I think that it would be maybe better to take care of this in the batch loader and raise an exception if all labels in the batch are the same.

djstrong:
Sure, but the exception should not interrupt the training - this mini-batch should be omitted (for simplicity).

yosipk:
Since such batch is useless (in the sense that the forward pass is done, but the backward pass doesn't modify the model), care should be taken in the batch loader. So yes, agreed: the batch loader could take care of this in the form of warning message, to alert the user that (s)he should make sure that batches are formed in a proper way.

djstrong:
So `train` method needs to accept custom `DataLoader`?

pommedeterresautee:
why not do the check in forward method? Therfore it can be applied to all kinds of dataloader.

djstrong:
That would be simpler in implementation. What should do and return `forward` method of `RankingLoss`? Simply `return 0.0`?

djstrong:
@yosipk I saw in the code that repeated sentences (as first or second point) are supported, e.g. in the machine translation one sentence can have multiple correct translations. Is the `RankingLoss` supporting this situation correctly? In the code positive examples gets target=1.0, and negatives 0.0. Is this range obligatory for `RankingLoss`? E.g. `CosineSimilarity` returns from range [-1;1] - can negatives be -1.0? Or in general, can I assign any score for a `DataPair` and the loss will be correct?

djstrong:
For the `nan` problem I have edited `RankingLoss` by adding 1 to divisors:
```
        neg_targets_01_sum = torch.sum(neg_targets, dim=1) + 1
        neg_targets_10_sum = torch.sum(neg_targets, dim=0) + 1
```

djstrong:
@yosipk How to use `ModelSimilarity`? E.g. linear layer on concatenated inputs.

PR #1074, opened 2019-09-05
Reduce inference time by 30% when using Flair embeddings

pommedeterresautee:
It appeared on profiler (screenshots on https://github.com/zalandoresearch/flair/issues/1070) that `Tensor.clone()` was one of the slowest single operations during inference.
Looking at the code, it's always called even when gradient is disabled.
My understanding is that when gradient is disabled, embeddings can't change and so there is no need to clone the Tensor to avoid it to be updated. That's what does this commit.

I have made tests on inference, but not on LM training, I don't think it changes anything but let me know if I misses something.

Time on 2080 TI  with storage set to `None` on my French dataset.
- before 36 secs
- after 25 secs

FYI, before I was doing measures on storage set to `"cpu"`, that's why *before* time decreased a lot compared to PR https://github.com/zalandoresearch/flair/pull/1068 (43 secs after optimization). Details on https://github.com/zalandoresearch/flair/issues/1070

Add a warning when cpu option is used with a GPU, plus update the documentation.
fix https://github.com/zalandoresearch/flair/issues/1070

All evaluation where still having embedding storage set to cpu, I moved them to none

### DIFF
--- a/flair/trainers/trainer.py
+++ b/flair/trainers/trainer.py
@@ -75,7 +75,7 @@ class ModelTrainer:
         train_with_dev: bool = False,
         monitor_train: bool = False,
         monitor_test: bool = False,
-        embeddings_storage_mode: str = "cpu",
+        embeddings_storage_mode: str = "none",
         checkpoint: bool = False,
         save_final_model: bool = True,
         anneal_with_restarts: bool = False,
--- a/flair/trainers/trainer.py
+++ b/flair/trainers/trainer.py
@@ -75,7 +75,7 @@ class ModelTrainer:
         train_with_dev: bool = False,
         monitor_train: bool = False,
         monitor_test: bool = False,
-        embeddings_storage_mode: str = "none",
+        embeddings_storage_mode: str = "cpu",
         checkpoint: bool = False,
         save_final_model: bool = True,
         anneal_with_restarts: bool = False,
--- a/resources/docs/TUTORIAL_7_TRAINING_A_MODEL.md
+++ b/resources/docs/TUTORIAL_7_TRAINING_A_MODEL.md
@@ -1,16 +1,17 @@
 # Tutorial 7: Training a Model
 
-This part of the tutorial shows how you can train your own sequence labelling and text
-classification models using state-of-the-art word embeddings.
+This part of the tutorial shows how you can train your own sequence labelling and text classification models using state-of-the-art word embeddings.
 
 For this tutorial, we assume that you're familiar with the [base types](/resources/docs/TUTORIAL_1_BASICS.md) of this
-library and how [word embeddings](/resources/docs/TUTORIAL_3_WORD_EMBEDDING.md) work (ideally, you also know how [flair embeddings](/resources/docs/TUTORIAL_4_ELMO_BERT_FLAIR_EMBEDDING.md) work). You should also know how to [load
+library and how [word embeddings](/resources/docs/TUTORIAL_3_WORD_EMBEDDING.md) work (ideally, you also know how [flair embeddings](/resources/docs/TUTORIAL_4_ELMO_BERT_FLAIR_EMBEDDING.md) work). 
+You should also know how to [load
 a corpus](/resources/docs/TUTORIAL_6_CORPUS.md).
 
 
 ## Training a Sequence Labeling Model
 
-Here is example code for a small NER model trained over WikiNER data, using simple GloVe embeddings. To run this code, you first need to obtain the CoNLL-03 English dataset (alternatively, use `NLPTaskDataFetcher.load_corpus(NLPTask.WNUT_17)` instead for a task with freely available data).
+Here is example code for a small NER model trained over WikiNER data, using simple GloVe embeddings. 
+To run this code, you first need to obtain the CoNLL-03 English dataset (alternatively, use `NLPTaskDataFetcher.load_corpus(NLPTask.WNUT_17)` instead for a task with freely available data).
 
 In this example, we downsample the data to 10% of the original data because the WikiNER dataset is huge:
 
@@ -74,11 +75,9 @@ plotter.plot_weights('resources/taggers/example-ner/weights.txt')
 ```
 
 Alternatively, try using a stacked embedding with FlairEmbeddings and GloVe, over the full data, for 150 epochs.
-This will give you the state-of-the-art accuracy we report in the paper. To see the full code to reproduce experiments,
-check [here](/resources/docs/EXPERIMENTS.md).
+This will give you the state-of-the-art accuracy we report in the paper. To see the full code to reproduce experiments, check [here](/resources/docs/EXPERIMENTS.md).
 
-Once the model is trained you can use it to predict the class of new sentences. Just call the `predict` method of the
-model.
+Once the model is trained you can use it to predict the class of new sentences. Just call the `predict` method of the model.
 
 ```python
 # load the model you trained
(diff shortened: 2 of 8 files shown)


flair/trainers/trainer.py
```diff
@@ -75,7 +75,7 @@ def train(
         train_with_dev: bool = False,
         monitor_train: bool = False,
         monitor_test: bool = False,
-        embeddings_storage_mode: str = "cpu",
+        embeddings_storage_mode: str = "none",
```
alanakbik: During training, it does make sense to have 'cpu' be the default option since many embeddings like FlairEmbeddings and the various transformer embeddings are expensive to produce. If the default is 'none' this will slow down training since embeddings have to be generated over and over again at each epoch.

resources/docs/TUTORIAL_7_TRAINING_A_MODEL.md
```diff
@@ -316,18 +315,25 @@ trainer.train('resources/taggers/example-ner',
...
-also means that embeddings get generated over and over again at each epoch. If you're using computationally costly embeddings such as BERT and you don't have a strong GPU, this may make training very slow. 
+1. **'none'**: If you set `embeddings_storage_mode='none'`, embeddings do not get stored in memory. 
+Instead they are generated on-the-fly in each training mini-batch. 
+The main advantage is that this keeps your memory requirements low.
 
-2. **'cpu'**: If you set `embeddings_storage_mode='cpu'`, embeddings will get stored in regular memory. This in many cases speeds things up significantly since embeddings only need to be computed in the first epoch, after which they are just retrieved from memory. A disadvantage is that this increases memory requirements. Depending on the size of your dataset and your memory setup, this option may not be possible.
+2. **'cpu'**: If you set `embeddings_storage_mode='cpu'`, embeddings will get stored in regular memory.
```
alanakbik: Same as above: During training, it does make sense to have 'cpu' be the default option since this can speed up things a lot. Only during inference (i.e. predict() method) it makes no sense unless you want to use the word embeddings later.

flair/trainers/trainer.py
pommedeterresautee: ok, I have added questions in my last issue :-)
Will restore the default option

resources/docs/TUTORIAL_7_TRAINING_A_MODEL.md
pommedeterresautee: ok thanks makes things more clear for me !!!!

pommedeterresautee:
@alanakbik for 
```
                    if self.fine_tune:
                        embedding = embedding.clone()
                    else:
                        embedding = embedding.detach()
```
Does `detach` makes sense? (because there is no gradient)

pommedeterresautee:
@alanakbik I have another optimization coming (mainly LRU cache on CUDA tensors for WordEmbeddings, small increase of memory foot print for significant speed gain when using WordEmbeddings by avoiding transfer between CPU and GPU and because of zipf law). 

Do you prefer to review first this PR before seeing the new one or both at the same time?

alanakbik:
@pommedeterresautee sorry for the delay - I'm planning to look into #1078 and review this PR today!

alanakbik:
Hello @pommedeterresautee thanks for the PR! Prediction speed is well increased when setting storage mode to 'cpu'. 

However, when setting storage mode to 'gpu', I now get OOM errors. For instance: 

```python
from flair.datasets import WNUT_17
from flair.embeddings import WordEmbeddings, StackedEmbeddings, FlairEmbeddings

corpus = WNUT_17(in_memory=True)

tag_type = 'ner'

# embeddings
embeddings = StackedEmbeddings([
    FlairEmbeddings('news-forward'),
    FlairEmbeddings('news-backward'),
    WordEmbeddings('glove'),
])

                                        embeddings=embeddings,
                                        tag_type=tag_type,
                                        )

# 6. initialize trainer

# # 7. start training
trainer.train(
    f'resources/taggers/local-test',
    max_epochs=5,
    train_with_dev=True,
    embeddings_storage_mode='gpu',
)
```

This previously trained well with GPU storage mode, but now is throwing memory errors. It may be that without the `.clone()` it somehow does not release the memory from the language model. That's strange because a `.detach()` seems like it should be enough to move the hidden state tensor out of scope, but somehow this isn't happening.

alanakbik:
@pommedeterresautee quick update: I think the problem lies in the line https://github.com/zalandoresearch/flair/blob/e0437d577be28cf7d099df588ad3a997e542df60/flair/embeddings.py#L1856 

where we select from all hidden states of all characters in the sentence only the states at one position to become the word embedding. Essentially, what we need to do here is copy these states over to a new tensor so that we can discard all the rest of the huge tensor `all_hidden_states_in_lm`. As discussed [here](https://stackoverflow.com/questions/55266154/pytorch-preferred-way-to-copy-a-tensor) and elsewhere, the preferred way in PyTorch is to do this via `clone()`. After the copy, no references to the tensor exist and it gets cleaned up.

With your PR, we don't do the `clone()` and it still works if the embedding storage mode is either 'cpu' or 'none' because these storage modes cause either the embedding to be deleted ('none') or a new tensor to be created on cpu (essentially another copy operation), thus freeing up `all_hidden_states_in_lm`. This means we save the `clone()` operation giving us a significant speed-up.

But if the storage mode is 'gpu', the `embedding` tensor is preserved and so also `all_hidden_states_in_lm` which quickly blows up the memory. So the PR as it is speeds up 'none' and 'cpu' but breaks 'gpu'.

pommedeterresautee:
Came to the same conclusion than you.
It s like the new tensor is a kind of view over a part of the large matrix.
It would make sense to proceed that way to avoid copy and share data layer between tensors and the large matrix (tensor keeping a simple reference to the data layer of the matrix).
clone() would perform a real copy on a brand new data layer.
The problem is that if this was the case, each token of the batch would have a reference toward the same matrix, therefore the memory should not raise exception.

I agree that this has no effect on none, but it increases memory consumption on CPU.
Moreover, inside Language Model I don't know which way have been chosen to store embeddings, and can't perform a simple if not "none" then tensor.clone...

Complicated...

alanakbik:
One potential solution would be to define a global variable. We currently only have two: the `flair.cache_root` and the `flair.device`, both defined in the `__init__.py`: 
https://github.com/zalandoresearch/flair/blob/e0437d577be28cf7d099df588ad3a997e542df60/flair/__init__.py#L8 

We could add an `optimization_settings` variable here that initializes as "default". When someone selects mbedding storage mode to be 'cpu', we change this variable. In the FlairEmbeddings class we do the check: if `optimization_settings` is set to default, we always do the `.clone()` otherwise we don't. 

Drawback is that this adds another global variable but if it is only for non-essential optimization instructions maybe it would be ok in this case. What do you think? I could add this in a separate PR.

pommedeterresautee:
tks @alanakbik.
For now I have reversed the optimization so you can merge this PR and https://github.com/zalandoresearch/flair/pull/1022 if you have time.
After that, I will push some other optimizations I have worked on (hoping that they don't come with their own problem :-) ), it will give me some time to work on clone() in parallel.
What do you think @alanakbik ?
We can choose the global variable path when we will be sure there is no other solution to take the benefit of removing clone?

pommedeterresautee:
I think I start to understand, the view theory may be the right one, they share `storage` until clone() is called. What I still don't understand is why it raises OOM, it should not consume more memory if they share the storage...

image

alanakbik:
Hi @pommedeterresautee yes that sounds good - I'll merge shortly! 

Thanks for sharing your findings! I guess because they share the storage, the `all_hidden_states_in_lm` tensor is prevented from going out of scope and thus garbage collected.

alanakbik:
:+1:

adizdari:
👍

PR #1077, opened 2019-09-06
GH-1073: batch first

### LINKED ISSUE #1073: Change SequenceTagger and DocumentRNNEmbeddings to batch_first mode
We are currently not consistent in Flair with whether we use `batch_first=True` or not. The `LockedDropout` and `WordDropout` classes for instance do not currently support a `batch_first` mode, so we do awkward transpose operations in the `SequenceTagger` and `DocumentRNNEmbeddings`. 

**Task**: Make everything `batch_first`. Add support for this in `LockedDropout` and `WordDropout` and remove transpose operations from `SequenceTagger` and `DocumentRNNEmbeddings`.

alanakbik:
Closes #1073 

This PR changes `WordDropout` and `LockedDropout` to be robust to `batch_first` mode. For `LockedDropout` this means adding a boolean parameter `batch_first` to the constructor that by default is set to `True`. 

The `WordDropout` is now implemented differently as the previous version was a 'lcoked' word dropout that would always drop the same index words across a mini-batch. Now, words are dropped fully at random. Some initial experiments show that this improves downstream task training. 

The PR also removes the transpose operations in the `SequenceTagger` and the `DocumentRNNEmbeddings` and removes the sorting logic from the `DocumentRNNEmbeddings`.

### DIFF
--- a/flair/embeddings.py
+++ b/flair/embeddings.py
@@ -2681,6 +2681,7 @@ class DocumentRNNEmbeddings(DocumentEmbeddings):
                 hidden_size,
                 num_layers=rnn_layers,
                 bidirectional=self.bidirectional,
+                batch_first=True,
             )
         else:
             self.rnn = torch.nn.GRU(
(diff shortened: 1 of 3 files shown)


kashif:
👍

alanakbik:
:+1:

PR #1082, opened 2019-09-10
GH-1079: Increase CPU training speed by pinning tensors

### LINKED ISSUE #1079: Keep word embedding tensors on pinned cpu memory during training
If `embeddings_storage_mode = 'cpu'` is selected during training, all word embedding tensors are moved to CPU memory and are at each epoch copied over to GPU memory when processing a mini-batch. It's possible to pin tensors so that the GPU copy operation becomes faster. The drawback is that pinning tensors is costly, so this would increase the time needed for the first epoch (which creates and pins the tensors) but would decrease the time needed for all subsequent epochs (since copy operations are now faster). 

Task: Add tensor pinning to training routine and investigate how this impacts training speed.

alanakbik:
This PR makes minor modifications to the `.to()` method of the `DataPoint` base class and all implementing classes, namely adding the option of moving a data point tensor to pinned memory. Pinning a tensor is a one-time cost but then allows all subsequents GPU tensor copy operations to work faster. When training a model with `embeddings_storage_mode = 'cpu'`, we at each epoch move tensors from CPU to GPU, so this PR increases overall training speed (closes #1079).  

This PR also adds a check if the `.to()` operation is even necessary (not needed if a tensor is already on the relevant device), leading to a small increase in training speed.

### DIFF
--- a/flair/data.py
+++ b/flair/data.py
@@ -153,7 +153,7 @@ class DataPoint:
         pass
 
     @abstractmethod
-    def to(self, device: str):
+    def to(self, device: str, pin_memory: bool = False):
         pass
 
     @abstractmethod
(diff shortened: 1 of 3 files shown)


kashif:
👍

alanakbik:
:+1:

PR #1084, opened 2019-09-10
Set a LRU cache for word embeddings for a decrease of 20% of inference time

pommedeterresautee:
Set an embedding LRU cache of Tensor of word embeddings.
This approach avoid to convert most used Gensim embeddings to Pytorch Tensor and even more important to avoid to transfer from computer RAM to GPU Ram (this is a slow operation).
Because of zipf law, the effect of such caching approach are magnified.
LRU cache is set to 10000 because it s very small and still provide most of the performance boost compared to loading all embeddings in GPU Ram. A 1000 embedding LRU cache is slightly less performant on my own dataset.

A second little optimization is to replace a call of `unsqueeze` on each token followed by a `cat` by a single call of `stack`.

Time to process 100 French documents decreased from 33s to 26s with this PR.
For what it worths, Spacy takes exactly the same time (26s) on my dataset with much lower accuracy on some tricky entities (and same accuracy on easiest to recognize entities).
Another change is on Ner HTML viewer with the introduction of a HTML title parameter.

### DIFF
--- a/flair/data.py
+++ b/flair/data.py
@@ -237,11 +237,11 @@ class Token(DataPoint):
         device = flair.device
         if len(self._embeddings.keys()) > 0:
             device = next(iter(self._embeddings.values())).device
-        self._embeddings[name] = vector.to(device, non_blocking=True)
+        self._embeddings[name] = vector.to(device)
 
     def to(self, device: str):
         for name, vector in self._embeddings.items():
-            self._embeddings[name] = vector.to(device, non_blocking=True)
+            self._embeddings[name] = vector.to(device)
 
     def clear_embeddings(self, embedding_names: List[str] = None):
         if embedding_names is None:
--- a/flair/data.py
+++ b/flair/data.py
@@ -563,7 +563,7 @@ class Sentence(DataPoint):
     def embedding(self):
         return self.get_embedding()
 
-    def set_embedding(self, name: str, vector):
+    def set_embedding(self, name: str, vector: torch.tensor):
         device = flair.device
         if len(self._embeddings.keys()) > 0:
             device = next(iter(self._embeddings.values())).device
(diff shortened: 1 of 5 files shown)


flair/data.py
```diff
@@ -237,7 +237,7 @@ def set_embedding(self, name: str, vector: torch.tensor):
         device = flair.device
         if len(self._embeddings.keys()) > 0:
             device = next(iter(self._embeddings.values())).device
-        self._embeddings[name] = vector.to(device, non_blocking=True)
+        self._embeddings[name] = vector.to(device)
```
alanakbik: I'm generally not yet sure if sending embeddings to devices with `non_blocking=True` makes sense or not, but the current version does set `non_blocking`, which would be removed with this PR. Do you think it makes no sense here?

pommedeterresautee: In my measures, without this LRU cache, non_blocking set to True was slightly slower (on inference) than when it was set to True. I was wondering why it was set to True plus why it is slower because https://discuss.pytorch.org/t/should-we-set-non-blocking-to-true/38234 says that it should be approximately the same when not required. Therefore I thought, it s not the default (kind of recommended) option, it s slightly slower, I remove. However, I may have missed something. 
To finish, my understanding of the description of the option and from when this line is called, I would have bet for a slight performance improvement when the option is set to True as the data are not used just after, so Pytorch can be lazy here. Anyway, with this cache, loading time are reduced a lot so not sure if it changes anything.

alanakbik: ok great - I haven't measured the difference so if its faster without `non_blocking` then lets remove it :)

alanakbik:
This is really great, including this PR the combined improvements of your recent PRs take inference time **from ~95 seconds** on CoNLL-03 **down to ~33 seconds** - Thanks for all your great work!

alanakbik:
:+1:

kashif:
👍

PR #1089, opened 2019-09-11
Star Wars: Episode II – Attack of the clone()

pommedeterresautee:
In PR #1074 it has been noticed that removing call to `clone()` get us some nice improvement, however, when training with embedding storage set to GPU, we rapidly got OOM exceptions. Finally the optimization have been reverted.

Pytorch Tensor have a a data layer called `storage`. When we subset an existing tensor, the new one has just a pointer toward the original Tensor storage, plus some offset, etc., in this case they are just a new View over existing data. It means that the old Tensor will never be garbage collected if the new Tensor is alive. On the other side, we avoid data copy that can be very costly.

calling clone() realizes storage copy in the new tensor, so the old one have no more pointer towards it and goes out of scope -> gc -> ciao

The last remaining mystery (at least for me) was why is there some OOM? Because all embeddings are pointing to the same large Tensor. What I didn't realized is that most of the original big Tensor is garbage: it includes padding plus dimensions related to character inside a Token, however we don't care about those dimensions, we only use the dimensions related to the first / last character of each token.

So the optimization is quite simple: we limit the number of clone() call, from one per token to one per sentence. Basically we first extract the useful part of the Tensor for each sentence, then we detach / clone, and we point towards the new Tensor. All tokens from the same sentence will share the same medium large Tensor. Memory footprint is exactly the same as before but by reducing the number of calls we get some nice perf improvement (which is slightly lower than not calling clone at all but benefit to cpu storage training).

Processing time of my French data goes from 26s to 21s, a 19% decrease for NER inference.

### DIFF
--- a/flair/embeddings.py
+++ b/flair/embeddings.py
@@ -1847,6 +1847,8 @@ class FlairEmbeddings(TokenEmbeddings):
                     append_padded_sentence(padded)
 
             # get hidden states from language model
+            # all_hidden_states_in_lm contains dimensions related to padding or character inside a token
+            # however only dimensions related to the first or last character are used
             all_hidden_states_in_lm = self.lm.get_representation(
                 sentences_padded, self.chars_per_chunk
             )
@@ -1857,17 +1859,15 @@ class FlairEmbeddings(TokenEmbeddings):
 
                 offset_forward: int = extra_offset
                 offset_backward: int = len(sentence_text) + extra_offset
+                dimensions_to_extract = list()
 
                 for token in sentence.tokens:
-
                     offset_forward += len(token.text)
 
                     if self.is_forward_lm:
-                        offset = offset_forward
+                        dimensions_to_extract.append(offset_forward)
                     else:
-                        offset = offset_backward
-
-                    embedding = all_hidden_states_in_lm[offset, i, :]
+                        dimensions_to_extract.append(offset_backward)
 
                     # if self.tokenized_lm or token.whitespace_after:
                     offset_forward += 1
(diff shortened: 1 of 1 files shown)


pommedeterresautee:
Ok, some good / bad news:
I have tried Connl 2003 and get same behaviour than the one you see (slower with this PR).
After comparison, my French dataset (100 legal cases) is made of very long sentences, much longer than what you find in Connl. I refactored the large `_add_embeddings_internal` function in smaller ones to see where is the slowness, and basically it costs something to create first the sentence (detach clone) then extract the token, compared to cloning each token individually, without having to create the sentence tensor in between.

To make it short, both dataset have different profiles and what benefit to one don't to the other. Which means this is not a good optimization. I will try something else, and if it doesn't work, may be we can think of something to avoid the clone() when used with embedding storage set to none.

FYI, on Conll 2003, my GPU usage goes from 60 to 70% (on French dataset it's 40-50%). Still some margin are remaining outside of model refactoring, but less an less.

alanakbik:
@pommedeterresautee thanks for investigating :) as a potential alternative to discuss, I've just pushed a branch in which we add one more global variable for optimization that is set in `store_embeddings()` and checked in `FlairEmbeddings._add_embeddings_internal()`. This gives us a nice speed boost (inference down from 28.8 to 26 seconds), but of course it is not exactly pretty from an object-oriented perspective. However, if we keep use of global variables like this to a minimum, it might be a solution. What do you think?

pommedeterresautee:
Yeah, seems reasonable to stop here and be pragmatic. Regarding the global variable, why not call it "embedding_storage" or something like that, and store the exact option inside. For this optimization we only want to know if it s gpu or not, but may be for future stuff, detailed information would be useful. The other thing is that it s more readable (name describe what is it). The drawback is that it would require to store the storage mode for both classification and similarity tasks too, so the variable is always a description of what it is. I imagine that the same kind of optimization are doable for classification? (I have not checked yet) What do you think? Does it worth it ?

alanakbik:
Good point - will rename. Many of these optimizations also benefit classification since they use the same embedding and embedding storage mechanism, so I imagine there's already a speed improvement there as well. But more optimization is probably always possible :) Big thanks for all your help in improving the speed, thanks to you we've already got over 3x speed increase (almost 4x) which is huge!

Issue #1067, opened 2019-09-04
Scalable hyperparameter selection with Ray Tune


alanakbik:
[Ray Tune](https://ray.readthedocs.io/en/latest/tune.html) is a framework for hyperparameter selection that supports running processes in parallel across multiple CPUs/GPUs and has connectors to a number of hyperparameter selection frameworks such as Hyperopt and Bayesopt. 

As discussed in #563 with @myaldiz and @richardliaw it would be interesting to add support for this framework and potentially replace the current hyperparameter selection classes based on Hyperopt with this more general framework. For this, we need to slightly refactor the model trainer and the checkpointing mechanism.

richardliaw:
I can help review if you want! Just let me know when you open the PR.

alanakbik:
@richardliaw thanks a lot - I'll let you know when it's ready!

PR #1091, opened 2019-09-12
GH-1067: ray tune

alanakbik:
This PR adds a first version of scalable hyperparameter tuning to Flair (see #1067), using the `tune` module of `ray`. It refactors the training interfaces to be more compatible: 

- The `ModelTrainer` gets refactored such that parameters that influence training become part of the constructor and only parameters that influence logging stay in the `.train()` method. Training parameters thus become variables of the class. Also, the method `_train_one_epoch()` is added that executes training of one epoch.
- Checkpointing is refactored such that it now belongs to the `ModelTrainer` and no longer to the respective downstream task classes.

The PR adds a `FlairTune` base class that needs to be extended for a specific experiment involving a corpus and embeddings, as well as the hyperparameters one wishes to tune. For instance, to set up a simple `TextClassifier` tuning experiment for the hyperparameters `dropout` and `mini_batch_size`, the base class can be extended as follows: 

```python
class TuneTextClassifier(FlairTune):

    def _setup(self, config):
        # A. set the corpus you are using
        self.corpus = TREC_6(in_memory=True)

        # B. define your embeddings and hyperparameters
        embeddings = DocumentRNNEmbeddings(
            [
                WordEmbeddings('en'),
            ],
            dropout=config["dropout"] if "dropout" in config else 0.5,
        )

        # define model
        model: SequenceTagger = TextClassifier(
            document_embeddings=embeddings,
            label_dictionary=self.corpus.make_label_dictionary(),
        )

        # C. define training parameters
        self.trainer = ModelTrainer(
            model,
            self.corpus,
            mini_batch_size=config["mini_batch_size"] if "mini_batch_size" in config else 8,
        )
```

In this example, we allow two variables for hyperparameter tuning: the model parameter of `dropout` and the training parameter of `mini_batch_size`.

Now, to run an experiment, execute tune for instance like this: 

```python
    # 1. set the number of parameter combinations to try
    number_of_experiments: int = 50

    # 2. set the maximum number of epochs per experiment
    max_epochs: int = 100

    # 3. set how many experiments at the same time and what resources does each get
    max_concurrent: int = 5
    resources_per_experiment = {
        "cpu": 1,
        "gpu": 0.2,
    }

    # 4. define your search space (the two parameters we wish to tune)
    search = HyperOptSearch(
        {
            "dropout": hp.uniform("dropout", 0.00, 0.8),
            "mini_batch_size": hp.choice("mini_batch_size", [2, 4, 8, 16, 32]),
        },
        max_concurrent=max_concurrent,
        reward_attr="mean_accuracy")

    experiment_base_folder = os.path.expanduser('~/sweep/')
    experiment_name = "experiment-" + datetime.today().strftime("%Y-%m-%d_%H-%M-%S")

    # Define scheduler (optional)
    scheduler = ray.tune.schedulers.AsyncHyperBandScheduler(
        time_attr='training_iteration',
        metric='mean_accuracy',
        mode='max',
        grace_period=10,
        max_t=max_epochs)

    # from ray.tune.util import validate_save_restore
    ray.init(ignore_reinit_error=True, temp_dir='tune')

    # run experiment
    trials: List[Trial] = tune.run(
        TuneTextClassifier,
        local_dir=experiment_base_folder,
        name=experiment_name,

        scheduler=scheduler,

        stop={
            "1-lr": 0.9999,
            "training_iteration": max_epochs
        },

        resources_per_trial=resources_per_experiment,
        num_samples=number_of_experiments,

        search_alg=search,

        return_trials=True,
    )

    # summarize results
    path = experiment_base_folder + experiment_name + '/results.tsv'
    write_experiment_summary(trials, path)
```

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ sudo: false
 env:
   - BOTO_CONFIG=/dev/null
 python:
-  - "3.6"
+  - "3.7"
 install:
   - python setup.py develop -q
 before_script: cd tests
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,7 +5,7 @@ env:
 python:
   - "3.6"
 install:
-  - python setup.py develop -q
+  - pip install -q -e .
 before_script: cd tests
 script:
   - pip freeze
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,7 +3,7 @@ sudo: false
 env:
   - BOTO_CONFIG=/dev/null
 python:
-  - "3.7"
+  - "3.6"
 install:
   - python setup.py develop -q
 before_script: cd tests
--- a/flair/embeddings.py
+++ b/flair/embeddings.py
@@ -2695,14 +2695,11 @@ class DocumentRNNEmbeddings(DocumentEmbeddings):
         self.name = "document_" + self.rnn._get_name()
 
         # dropouts
-        if locked_dropout > 0.0:
-            self.dropout: torch.nn.Module = LockedDropout(locked_dropout)
-        else:
-            self.dropout = torch.nn.Dropout(dropout)
-
-        self.use_word_dropout: bool = word_dropout > 0.0
-        if self.use_word_dropout:
-            self.word_dropout = WordDropout(word_dropout)
+        self.dropout = torch.nn.Dropout(dropout) if dropout > 0.0 else None
+        self.locked_dropout = (
+            LockedDropout(locked_dropout) if locked_dropout > 0.0 else None
+        )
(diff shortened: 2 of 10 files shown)


alanakbik:
@richardliaw thanks very much - will try out the fix! Do you have a rough idea when the next ray release with the modin fix will come out?

richardliaw:
The next release should be in 3 weeks or so.

ujvl:
@alanakbik just a heads up the issue installing ray w/ `setup.py` should be fixed now, in case you intended to re-open this.

PR #1095, opened 2019-09-13
GH-1089: Optimize FlairEmbeddings depending on storage mode

alanakbik:
As discussed in #1089, the `clone()` operation in `FlairEmbeddings` is expensive and only required if the `embedding_storage_mode` is 'gpu'. However, information of which storage mode is selected is not visible in `FlairEmbeddings`. To address this, this PR adds a new global variable `embedding_storage_mode` that is set in the `store_embeddings()` method and checked in `FlairEmbeddings`. This gives us a roughly 10% improvement in inference speed. 

The PR also renames the `embeddings_storage_mode` parameter of the `evaluate()` functions to `embedding_storage_mode` to make it consistent.

### DIFF
--- a/flair/__init__.py
+++ b/flair/__init__.py
@@ -11,8 +11,8 @@ if torch.cuda.is_available():
 else:
     device = torch.device("cpu")
 
-# global variable: optimization_settings
-optimization_mode = "default"
+# global variable: embedding_storage_mode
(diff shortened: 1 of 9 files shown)


MichaelHintz:
:+1:

adizdari:
👍

PR #1101, opened 2019-09-15
GH-1087: refactor checkpointing mechanism

### LINKED ISSUE #1087: When restarting from checkpoint, the model is trained more than max_epochs
**Describe the bug**
As in the title, when loading a model that was already trained for some epochs, setting max_epochs parameter trains for *further* max_epochs.

**To Reproduce**
I saved a checkpoint after 3 epochs. When I load this checkpoint and run with max_epochs=3, I see epoch 4, 5, 6 get executed.

**Expected behavior**
The above behavior could make sense, but I'd really like to set the **total** max epochs (or my jobs can potentially never end if they keep getting interrupted and restored).

So, in the example above, I would expect that no further epochs are trained.

Alternatively, how can I read how many epochs were already trained when I load the checkpoint?

alanakbik:
This PR refactors the model checkpointing mechanism and slims down interfaces / code required to load checkpoints. Closes #1087 

In detail: 

-  The methods `save_checkpoint` and `load_checkpoint` are no longer part of the `flair.nn.Model` interface. Instead, saving and restoring checkpoints is now (fully) performed by the `ModelTrainer`.
- The optimizer state and scheduler state are removed from the `ModelTrainer` constructor since they are no longer required here. 
- Loading a checkpoint is now one line of code (previously two lines).

```python
# 1. initialize trainer as always with a model and a corpus
trainer: ModelTrainer = ModelTrainer(model, corpus)

# 2. train your model for 2 epochs
trainer.train(
    'experiment/folder',
    max_epochs=2,
    # example checkpointing
    checkpoint=True,
)

# 3. load last checkpoint with one line of code
trainer = ModelTrainer.load_checkpoint('experiment/folder/checkpoint.pt', corpus)

# 4. continue training for 2 extra epochs
trainer.train('experiment/folder_2',  max_epochs=4) 
```

### DIFF
--- a/flair/models/sequence_tagger_model.py
+++ b/flair/models/sequence_tagger_model.py
@@ -254,6 +254,9 @@ class SequenceTagger(flair.nn.Model):
         embedding_storage_mode: str = "none",
     ) -> (Result, float):
 
+        if type(out_path) == str:
+            out_path = Path(out_path)
+
         with torch.no_grad():
             eval_loss = 0
 
--- a/flair/nn.py
+++ b/flair/nn.py
@@ -66,24 +66,6 @@ class Model(torch.nn.Module):
 
         torch.save(model_state, str(model_file), pickle_protocol=4)
 
-    def save_checkpoint(
-        self,
-        model_file: Union[str, Path],
-        optimizer_state: dict,
-        scheduler_state: dict,
-        epoch: int,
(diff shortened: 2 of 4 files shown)


yosipk:
👍

alanakbik:
:+1:

PR #1107, opened 2019-09-16
[refactoring] move padding to LM class

pommedeterresautee:
Some cleaning to prepare a large refactoring
Simply share padding code between Flair embeddings and character embeddings


alanakbik:
Cool, looking forward to the refactoring :) The `CharLMEmbeddings` BTW are marked for deprecation and will be removed with the next release, i.e. that's why there's lots of duplicated code with the newer `FlairEmbeddings` that replaced them a while ago.

alanakbik:
:+1:

yosipk:
👍

PR #1112, opened 2019-09-16
Little optimization on device check

pommedeterresautee:
close https://github.com/zalandoresearch/flair/issues/1104
Provide a constant 1s improvement on conll 2003. (26s -> 25s, I am now using batch of 256, because since I don't know when, it matters on inference speed, AFAIK on the last released version it didn't on my GPU)
The best thing is that it makes the global variable more useful :-)

### DIFF
--- a/flair/data.py
+++ b/flair/data.py
@@ -235,7 +235,7 @@ class Token(DataPoint):
 
     def set_embedding(self, name: str, vector: torch.tensor):
         device = flair.device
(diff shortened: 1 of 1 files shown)


alanakbik:
Cool, thanks for adding this :)

alanakbik:
:+1:

yosipk:
👍

PR #1119, opened 2019-09-17
GH-986: fix batch step in learning rate finder

### LINKED ISSUE #982: find_learning_rate() Issues
Hello. So I've been building a TextClassifier. I am up to the optimize learning rate step. However, I don't think the find_learning_rate function is working correctly, and I don't know quite enough about it to debug it very well. 

~1) The model is supposed to take awhile to train. It completes really fast when using this function.~

2) It's supposed to try 100 different learning rates. I get 87 iterations, not 100, and they are all using the same starting learning rate. 

3) I can't see what it's doing at all. There is no output of the train step. It just all of a sudden says its done, but it hasn't produced anything of value. 

I don't know if this is a small bug, a user error, or something bigger.

It almost feels like it's running 87 minibatches at the starting learning rate, and then thinking it's done already when it's barely started. 

I mean worst case scenario, I can use hyperopt to check a few learning rates over night. But it would be really cool if this method could be made to work. I need a little guidance to help debug this. 

Thanks so much.

### LINKED ISSUE #986: find_learning_rate step problem
**Describe the bug**

find_learning_rate does not work. It will use the learning rate from iteration 2 for all iterations after 2. 

**To Reproduce**

This will happen with any application of find_learning_rate. You can reproduce with the tutorial code if you like or any code. 

**Expected behavior**
I expect iteration 3 to use the learning rate for iteration 3, and 4 for 4, and so on. It should end at the maximum learning rate set. 

**Screenshots**

 - OS: Linux of course
 - Version MASTER

**Additional context**

scheduler = ExpAnnealLR(optimizer, end_learning_rate, iterations)
scheduler.step(1)
learning_rate = scheduler.get_lr()[0]

def get_lr(self):
        iteration = self.last_epoch + 1
        pct = iteration / self.iterations
        return [base_lr * (self.end_lr / base_lr) ** pct for base_lr in self.base_lrs]

So get_lr is iterating based on last_epoch. Though it's really supposed to be iterating based on mini-batches. Anyways, it gets worse. 

ExpAnnealLR does not have it's own step method. So it uses the base class one. 

def step(self, epoch=None):
        if epoch is None:
            epoch = self.last_epoch + 1
        self.last_epoch = epoch
        for param_group, lr in zip(self.optimizer.param_groups, self.get_lr()):
            param_group['lr'] = lr

This function expects the correct epoch number to be passed to it.  Fine. 

But it's used like this in this case. 

scheduler.step(1)

This makes the epoch always come out as two. 

so either 
1) find_learning_rate needs to pass the correct iteration number to step
or,
2) the ExpAnnealLR  class needs it's own step function to overide the default, that will correctly increment itself. 

Considering nothing else uses ExpAnnealLR, I'm going with solution 2 in my copy. You do what you like in your package.

### LINKED ISSUE #1108: find_learning_rate does not work anymore
**Describe the bug**
Before, I've run several experiments with which I could plot the optimal learning rate as in the tutorial. Now, after an upgrade of the codebase, I cannot anymore reproduce the same experiments.
What I find is similar to issue #986 , but in my case the learning rate is never updated (while in the above issue is updated at least at the first step).

**To Reproduce**
I've run the experiments with commit 24b72a6dfd1d0c2519794e1183cacc42c932c4f4  , and set a start lr of 1e-08 and an end lr of 1000, with 200 iterations. I get a tsv with 200 lines where the lr is always the same (slightly above the start lr).

**Expected behavior**
I previously run the same experiments with commit 6c27431ecc731446f12bc3205bf68d0a859b9a8d and they seemed to work as in the tutorial. The lr in the tsv spanned all the interval specified.

alanakbik:
Fixes the step problem in the learning rate finder.

closes #986 
closes #982
closes #1108 

```python

# find learning rate
learning_rate_tsv = trainer.find_learning_rate(
    'resources/taggers/example-ner',
    'learning_rate.tsv',
)
```

### DIFF
--- a/flair/trainers/trainer.py
+++ b/flair/trainers/trainer.py
@@ -564,52 +564,63 @@ class ModelTrainer:
 
         train_data = self.corpus.train
 
-        batch_loader = DataLoader(train_data, batch_size=mini_batch_size, shuffle=True)
-
         scheduler = ExpAnnealLR(optimizer, end_learning_rate, iterations)
 
         model_state = self.model.state_dict()
-        model_device = next(self.model.parameters()).device
         self.model.train()
 
-        for itr, batch in enumerate(batch_loader):
-            loss = self.model.forward_loss(batch)
+        step = 0
+        while step < iterations:
+            batch_loader = DataLoader(
+                train_data, batch_size=mini_batch_size, shuffle=True
+            )
+            for batch in batch_loader:
+                step += 1
 
-            optimizer.zero_grad()
-            loss.backward()
-            torch.nn.utils.clip_grad_norm_(self.model.parameters(), 5.0)
-            optimizer.step()
-            scheduler.step(1)
-            learning_rate = scheduler.get_lr()[0]
+                # forward pass
+                loss = self.model.forward_loss(batch)
 
-            loss_item = loss.item()
-            if itr == 0:
-                best_loss = loss_item
-            else:
-                if smoothing_factor > 0:
-                    moving_avg_loss = (
-                        smoothing_factor * moving_avg_loss
-                        + (1 - smoothing_factor) * loss_item
+                # update optimizer and scheduler
+                optimizer.zero_grad()
+                loss.backward()
+                torch.nn.utils.clip_grad_norm_(self.model.parameters(), 5.0)
+                optimizer.step()
+                scheduler.step(step)
+
+                print(scheduler.get_lr())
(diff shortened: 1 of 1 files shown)


kashif:
:+1:

alanakbik:
:+1:

yosipk:
👍

PR #1130, opened 2019-09-19
Another strong reduce of concatenation for a small optimization (-4% inference time)

pommedeterresautee:
Reorganize embeddings + add some padding such a way we have only one call to concat (and no stack operation). Padding is preallocated to limit memory allocation. Get a constant 1 second inference time reduction on CONLL 2003 on 2080TI (25 -> 24s).
Had no measurable effect on my French dataset.

Good thing, the code is easier to read :-)

Nb: not related but I made a mistake in my measures... unfortunately memory transfer is not the main remaining bottleneck, I measured some synchronization operation which was happening before memory transfer. now I am using `;CUDA_LAUNCH_BLOCKING=1 ` before using cProfile instead of trying to be smart and calling synchronize manually here and there... Another thing, little functions called million of times appears slower than they are with the profiler...

### DIFF
--- a/flair/models/language_model.py
+++ b/flair/models/language_model.py
@@ -144,7 +144,9 @@ class LanguageModel(nn.Module):
                 ]
 
                 sequences_as_char_indices.append(char_indices)
-            t = torch.tensor(sequences_as_char_indices, dtype=torch.long).to(device=flair.device, non_blocking=True)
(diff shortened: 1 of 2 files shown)


alanakbik:
Thanks - looks good!

alanakbik:
:+1:

yosipk:
👍

PR #1138, opened 2019-09-20
GH-1072: Batch-growth annealing

### LINKED ISSUE #1072: Accumulating gradients to enable larger mini-batches
This [blog post](https://medium.com/huggingface/training-larger-batches-practical-tips-on-1-gpu-multi-gpu-distributed-setups-ec88c3e51255) by @thomwolf explains an easy way to enable larger mini-batches with limited memory: Chop a mini-batch down into smaller units and accumulate their gradients before doing the backward step.
 
We've had a number of users that only got training to run with their memory setups by choosing really small mini-batch sizes (for instance #903 #685 #736), so this approach could help in such cases.

**Task**: Add option for accumulating gradients to the `ModelTrainer`

alanakbik:
The paper [Don't Decay the Learning Rate, Increase the Batch Size](https://arxiv.org/abs/1711.00489) makes the case for increasing the batch size over time instead of annealing the learning rate. 

This PR adds the possibility to have arbitrarily large mini-batch sizes with an accumulating gradient strategy (closes #1072). It introduces the parameter `mini_batch_chunk_size` that you can set to break down large mini-batches into smaller chunks for processing purposes. 

So let's say you want to have a mini-batch size of 128, but your memory cannot handle more than 32 samples at a time. Then you can train like this: 

```python
trainer.train(
    "path/to/experiment/folder",
    # set large mini-batch size
    mini_batch_size=128,
    # set chunk size to lower memory requirements
    mini_batch_chunk_size=32,
)
```

Because we now can arbitrarly raise mini-batch size, we can now execute the annealing strategy in the above paper. Do it like this: 

```python
trainer.train(
    "path/to/experiment/folder",
    # set initial mini-batch size
    mini_batch_size=32,
    # choose batch growth annealing 
    batch_growth_annealing=True,
)
```

This will double the mini-batch size each time the learning rate anneals. You can also combine this with "annealing with restarts" in which the last best model state is restored each time the learning rate anneals. 

```python
trainer.train(
    "path/to/experiment/folder",
    # set initial mini-batch size
    mini_batch_size=32,
    # choose batch growth annealing 
    batch_growth_annealing=True,
    # reset model state to best on each anneal
    anneal_with_restarts=True,
)
```

### DIFF
--- a/flair/models/sequence_tagger_model.py
+++ b/flair/models/sequence_tagger_model.py
@@ -445,7 +445,6 @@ class SequenceTagger(flair.nn.Model):
             return sentences
 
     def forward(self, sentences: List[Sentence]):
-        self.zero_grad()
 
         self.embeddings.embed(sentences)
 
--- a/flair/trainers/trainer.py
+++ b/flair/trainers/trainer.py
@@ -61,7 +61,7 @@ class ModelTrainer:
         base_path: Union[Path, str],
         learning_rate: float = 0.1,
         mini_batch_size: int = 32,
-        eval_mini_batch_size: int = None,
+        mini_batch_chunk_size: int = 32,
         max_epochs: int = 100,
         anneal_factor: float = 0.5,
         patience: int = 3,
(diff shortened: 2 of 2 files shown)


yosipk:
👍

alanakbik:
:+1:

lucaventurini:
Hi @alanakbik , how do we fix the lr with this method? As I understood, we shouldn't reduce anymore the lr if we already double the batch size, am I correct?

I tried setting the anneal factor to 1.0, but I got an error. Also, should there be a max_batch_size equivalent to the min_learning_rate to stop early?

PR #1142, opened 2019-09-22
Accept string as input in NER predict

pommedeterresautee:
Add a new kind of dataset for strings and a dataloader to load them. That way new user wanted to make test doesn't need to understand how to build a Sentence. Plus it's more convenient to use in general for advanced users.
Has no impact on speed, however multiprocessing (to tokenize during GPU is working) makes things slower so it's kept disabled.

### DIFF
--- a/flair/datasets.py
+++ b/flair/datasets.py
@@ -570,7 +570,11 @@ class StringDataset(FlairDataset):
     A Dataset taking string as input and returning Sentence during iteration
     """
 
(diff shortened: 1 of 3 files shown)


alanakbik:
Very cool, thanks @pommedeterresautee! Example usage: 

```python
# init tagger
tagger= SequenceTagger.load('ner')

# predict over list of strings
sentences = tagger.predict(
    [
        'George Washington went to Berlin .', 
        'George Berlin lived in Washington .'
    ]
)

# output predictions
for sentence in sentences:
    print(sentence.to_tagged_string())
```

alanakbik:
:+1:

yosipk:
👍

PR #1145, opened 2019-09-23
Optimize character to index conversion for a 4% decrease on CONNL 2003

pommedeterresautee:
Optimize the way padding is done by reducing the number of dictionary call.
Remove the conversion to UTF-8, and take care of old models in old format.
In the future, the conversion may be removed safely.

close https://github.com/zalandoresearch/flair/issues/1133
@alanakbik finally I did both, change in dictionary deserialization and change inside the class itself.

CONNL 2003 from 24s to 23s (-4%)
French dataset from 12s to 11s (-8%)

### DIFF
--- a/flair/data.py
+++ b/flair/data.py
@@ -78,6 +78,9 @@ class Dictionary:
         self.item2idx_not_encoded = dict([(key.decode("UTF-8"), value) for key, value in self.item2idx.items()])
 
     def save(self, savefile):
+        if not hasattr(self, "item2idx_not_encoded"):
+            self.__decode_utf_internal()
+
         import pickle
 
         with open(file=savefile, mode="wb") as f:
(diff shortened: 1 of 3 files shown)


pommedeterresautee:
There was strange issues because of dictionary conversion, so I finally made a small version of the optimization, keeping things like they are :-( May be to investigate in the future

alanakbik:
Ah ok I was just measuring speed differences and finding they are roughly the same. Is this because of the last changes?

pommedeterresautee:
Here I still have 23s and 11s for CONNL and French dataset. May be it's a less than 1 sec change but with rounding it appears bigger?

alanakbik:
Ok great - looks good! Thanks for all your help!

alanakbik:
:+1:

yosipk:
👍

PR #1148, opened 2019-09-23
fix get_each_embedding when storage embedding is set to CPU

pommedeterresautee:
When run with CPU storage on a computer with CUDA / GPU, embeddings may not be on the right storage during training.

I don't know why, but it was not crashing on my own train task (French dataset)

`get_embedding` uses `get_each_embedding` now to get the same security.
`get_subembedding` has been deleted because it is not used in the project and would require to generalize get_embedding, implying more complexity.

The real issue I just realized is that Travis was not catching this bug because it doesn't run integration tests on GPU, and in this case it was a device storage issue. I don't know how to improve that.

### DIFF
--- a/flair/data.py
+++ b/flair/data.py
@@ -260,31 +260,21 @@ class Token(DataPoint):
                     del self._embeddings[name]
 
     def get_each_embedding(self) -> torch.tensor:
-        result = []
+        embeddings = []
         for embed in sorted(self._embeddings.keys()):
             embed = self._embeddings[embed].to(flair.device)
             if (flair.embedding_storage_mode == "cpu") and embed.device != flair.device:
                 embed = embed.to(flair.device)
-            result.append(embed)
-        return result
+            embeddings.append(embed)
+        return embeddings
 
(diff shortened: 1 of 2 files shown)


alanakbik:
@pommedeterresautee yes that's a good point on travis test not running on GPU. Before each release, I run manually on a GPU setup but of course that's not a great solution.

alanakbik:
BTW this fix is now already part of #1145 right?

pommedeterresautee:
Yes it is. I wanted to separate issues, but can't let a PR with a bug... so you got 2 times the same thing :-)

pommedeterresautee:
Oh, do you have an issue with hyperparamter tests? I always had them and they don't appear on Travis. Never tried to dig.

alanakbik:
For me the hyperparameter tests run through locally, though the hyperparameter selection module has not really been updated in a while and is more or less marked for replacement. We hope to find a better solution in the not too distant future to help users find good hyperparameters but are still looking into options.

PR #1150, opened 2019-09-23
GH-804: add two multilingual character dictionaries

### LINKED ISSUE #804: CharacterEmbeddings: missing characters in vocab
Hi,

there's one major issue with the `CharacterEmbeddings` class: by default it uses a "common chars" dictionary (fetched from [here](https://s3.eu-central-1.amazonaws.com/alan-nlp/resources/models/common_characters)).

I manually loaded these common characters with:

```python
d = Dictionary.load(name="chars")

len(d.get_items()) # Outputs 275
```

Then I used a Bulgarian word to check if all characters appear in the common characters dictionary:

```python
for char in "оглуша":
  if char not in d.get_items():
    print(char)
```

Unfortunately, the `ш` character is not found.

I think we should update these common characters dictionary :)

alanakbik:
This PR adds the two multilingual character dictionaries computed by @stefan-it. 

Load with 

```python
dictionary = Dictionary.load('chars-large')
print(len(dictionary.idx2item))

dictionary = Dictionary.load('chars-xl')
print(len(dictionary.idx2item))
```

Closes #804 - big thanks to @stefan-it!

### DIFF
--- a/flair/data.py
+++ b/flair/data.py
@@ -118,6 +118,16 @@ class Dictionary:
             char_dict = cached_path(base_path, cache_dir="datasets")
             return Dictionary.load_from_file(char_dict)
 
+        if name == "chars-large" or name == "common-chars-large":
+            base_path = "https://s3.eu-central-1.amazonaws.com/alan-nlp/resources/models/common_characters_large"
+            char_dict = cached_path(base_path, cache_dir="datasets")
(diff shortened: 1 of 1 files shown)


yosipk:
👍

alanakbik:
:+1:

PR #1160, opened 2019-09-26
Fix broken link in README

fte10kso:
The current link ([Contextual String Embeddings for Sequence Labeling](https://aclanthology.coli.uni-saarland.de/papers/C18-1139/c18-1139)) has an expired SSL certificate and then gives a 404 error.


alanakbik:
@fte10kso thanks for spotting and fixing this!

alanakbik:
:+1:

yosipk:
👍

PR #1179, opened 2019-10-02
Hash embeddings

carljohanrehn:
I have added hash embeddings using the hashing trick, see for example https://en.wikipedia.org/wiki/Feature_hashing#Feature_vectorization_using_hashing_trick

This code snippet explains how the suggested feature could be used:

from flair.embeddings import HashEmbeddings

word_embedding = HashEmbeddings()
sentence = corpus.train[0]
word_embedding.embed(sentence)

for token in sentence:
print(token)
print(token.embedding)

### DIFF
--- a/flair/embeddings.py
+++ b/flair/embeddings.py
@@ -565,7 +565,6 @@ class HashEmbeddings(TokenEmbeddings):
 
         self.to(flair.device)
 
-
     @property
     def num_embeddings(self) -> int:
         return self.__num_embeddings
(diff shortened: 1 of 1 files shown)


alanakbik:
@carljohanrehn thanks for adding this! Unfortunately this PR will undo some recent changes. Can you make sure that the PR only changes the code for the HashEmbeddings class and not the other parts?

carljohanrehn:
Yes, if you can guide me on how to do it. As I understand it, it will
suffice if I merge my branch into the master of my fork and then do a pull
to update my master. Then it should be possible create a new PR. Is this
correct?

alanakbik:
Yes there are several ways of doing this. You need to make sure that your master is the same as our master, by doing a 'git pull'. Then, you can merge the master into your branch by first switching to your branch with 'git checkout [your branch name]' and then doing 'git merge master'.

carljohanrehn:
Thanks, I'll try that...

PR #1185, opened 2019-10-05
Added hash embeddings with hashing trick

carljohanrehn:
I tried to make a pull request before, but failed twice, since I couldn't do a proper merge from my master. Thus, I killed my fork and created a new one. Hopefully it will work better this time.

I have added hash embeddings using the hashing trick, see for example https://en.wikipedia.org/wiki/Feature_hashing#Feature_vectorization_using_hashing_trick

This code snippet explains how the suggested feature could be used:

from flair.embeddings import HashEmbeddings

word_embedding = HashEmbeddings()
sentence = corpus.train[0]
word_embedding.embed(sentence)

for token in sentence:
print(token)
print(token.embedding)

### DIFF
--- a/flair/embeddings.py
+++ b/flair/embeddings.py
@@ -556,7 +556,6 @@ class HashEmbeddings(TokenEmbeddings):
         self.__embedding_length = embedding_length
 
         self.__hash_method = hash_method
-        self.__hash_function = hashlib.new(hash_method)
 
         # model architecture
         self.embedding_layer = torch.nn.Embedding(
@@ -578,8 +577,9 @@ class HashEmbeddings(TokenEmbeddings):
     def _add_embeddings_internal(self, sentences: List[Sentence]) -> List[Sentence]:
 
         def get_idx_for_item(text):
-            self.__hash_function.update(bytes(str(text), 'utf-8'))
-            return int(self.__hash_function.hexdigest(), 16) % self.__num_embeddings
+            hash_function = hashlib.new(self.__hash_method)
+            hash_function.update(bytes(str(text), 'utf-8'))
+            return int(self.hash_function.hexdigest(), 16) % self.__num_embeddings
 
         hash_sentences = []
         for i, sentence in enumerate(sentences):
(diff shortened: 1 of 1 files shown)


alanakbik:
Hello @carljohanrehn thanks for adding this! I fI may ask how are you currently using these embeddings, i.e. what are they useful for?

Another thing: I am getting a 

```console
TypeError: can't pickle _hashlib.HASH objects
```

error if I use this class in training - are you getting this as well?

carljohanrehn:
Sorry, since I'm new to Flair, I haven't really done any training with it yet. However, I have not seen this error applying word and document embedding on my data.

My intention is to extend the implementation as described in the paper "Hash Embeddings for Efficient Word Representations" by Svenstrup et al, see https://arxiv.org/abs/1709.03933.

alanakbik:
Ok thanks, I'll take a look at the serialization error!

alanakbik:
@carljohanrehn I am unsure how to address the serialization problem. Here's how to reproduce the error: 

```python
embeddings = HashEmbeddings()
torch.save(embeddings, 'hashfile.pt')
```

Is there perhaps a different hasing library you could use that lets you serialize the objects?

carljohanrehn:
I'm sure there is. I will take a look into it. Not now, but later...

flair/embeddings.py
```diff
@@ -542,6 +544,76 @@ def extra_repr(self):
...
+    @property
+    def embedding_length(self) -> int:
+        return self.__embedding_length
+
+    def _add_embeddings_internal(self, sentences: List[Sentence]) -> List[Sentence]:
+
+        def get_idx_for_item(text):
+            hash_function = hashlib.new(self.__hash_method)
+            hash_function.update(bytes(str(text), 'utf-8'))
+            return int(self.hash_function.hexdigest(), 16) % self.__num_embeddings
```
alanakbik: `self.hash_function` throws an error since `has_function` is no longer a global variable. This should be replaced with: 
`return int(hash_function.hexdigest(), 16) % self.__num_embeddings`

carljohanrehn:
Of course, how could I miss that.  As I remember, I tested the change
before committing it.  I'll fix it. Thank you for reviewing the code!

alanakbik:
The error is caused by the missing pymongo dependency, already fixed in master. Will merge this PR!

alanakbik:
:+1:

PR #1191, opened 2019-10-07
GH-1190 Fix OneHotEmbeddings

### LINKED ISSUE #1190: OneHotEmbeddings has error in implementation
**Describe the bug**
The implementation of OneHotEmbeddings for fields other than text has two issues:
* the value used in the counter is the instance of the class token, not the value of the token
* during the computation of embedding in `_add_embeddings_internal` the type of the field is not taken into account

**To Reproduce**
Use OneHotEmbedding with field other than `text` and check the contents of the dictionary, which is printed out. Should include all values with count >= 3, but only includes the `<unk>`.

**Expected behavior**
The dictionary should include all values with count >= 3. During computation, the value of the specific field, not the text field should be used.

 - OS [Linux]:
 - Version [flair-0.4.3]:

**Additional context**
I will send a PR with the fix shortly.

apohllo:
Fixes #1190

### DIFF
--- a/flair/embeddings.py
+++ b/flair/embeddings.py
@@ -449,7 +449,6 @@ class FastTextEmbeddings(TokenEmbeddings):
     def extra_repr(self):
         return f"'{self.embeddings}'"
 
-
 class OneHotEmbeddings(TokenEmbeddings):
     """One-hot encoded embeddings."""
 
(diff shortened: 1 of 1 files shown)


alanakbik:
@apohllo thanks for fixing this! 

Can I ask how you are using this feature, i.e. is there a task for which it is important to embed something other than text?

apohllo:
Yes, for instance we embed POS tags and for Polish it gives several p.p. improvement for sequence tagging task.

alanakbik:
:+1:

djstrong:
In the training a model tutorial (https://github.com/zalandoresearch/flair/blob/master/resources/docs/TUTORIAL_7_TRAINING_A_MODEL.md) we can append POS tags:
```
embedding_types: List[TokenEmbeddings] = [
    WordEmbeddings('glove'),
    OneHotEmbeddings(corpus=corpus, field='pos'),
]
```

yosipk:
👍

Issue #1198, opened 2019-10-08
Logging performance on the (subset) of train set


yosipk:
Flair currently supports logging performance measures only on the development set, while the performance on the train set is assessed only via loss. This loss is a) an average of losses over all train batches in an epoch, b) is a proxy for performance measure (e.g. accuracy for classification). 

In addition to this information, it would also be good to have the information about the performance of the model on the train set, to assess behavior of the model, i.e. to see if the model is over-fitting. However, since the train set is usually big, we could estimate this performance on the subset of the training data.

alanakbik:
Was added a while back.

PR #1202, opened 2019-10-09
GH-1198: Added option to evaluate on the (subset of) train split.

yosipk:
PR implements option to evaluate on the subset of train split via parameters of trainer function:
* `eval_on_train_fraction`: The fraction of train data to do the evaluation on.  If 0. the evaluation is not performed on fraction of training data. If 'dev' the size is determined from dev set size.
* `eval_on_train_shuffle`: If `True` the train data fraction is determined on the start of training and kept fixed during training, otherwise it's sampled at beginning of each epoch.

### DIFF
--- a/flair/trainers/trainer.py
+++ b/flair/trainers/trainer.py
@@ -31,6 +31,7 @@ from flair.training_utils import (
     Result,
     store_embeddings,
 )
+import random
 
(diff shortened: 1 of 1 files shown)


alanakbik:
:+1:

Issue #1203, opened 2019-10-09
Add support for simple convolutional network for images


yosipk:
So far Flair supports only pre-computed image features and image features from known architectures (ResNet-50, MobileNet.v2), optionally with weights pre-trained on ImageNet dataset. It would be nice to have an option to _define_ a simple convolutional architecture in Flair.

alanakbik:
Was added a while back.

PR #1204, opened 2019-10-09
GH-1203: Added support for simple convolutional network that operates on

yosipk:
images, with optional transformer on top instead of pooling.


alanakbik:
:+1:

PR #1257, opened 2019-10-31
Add staticmethod decorator

yahshibu:
Hello,
I suspect that it is better to attach the staticmethod decorator to some methods.
If you like it, please merge this into master.
Thank you.


yahshibu:
I've modified requirements.txt because I could not pass the test owing to the failure of pytorch importing.

yahshibu:
I believe that the test by travis-ci was stopped at the place not related to my modifications.
An error occurred on test_transformer_embeddings.py, but I didn't change any about embeddings.

Could you check the following three points?

1. the fact that my modifications are independent of this failure of the test 
2. whether the test by travis-ci properly works now
3. whether my modifications are acceptable

Thank you.

alanakbik:
@yahshibu thanks for adding this and please excuse the delay in checking the PR! The error was somewhere else and is now fixed so the PR is green!

flair/nn.py
```diff
@@ -68,13 +70,13 @@ def save(self, model_file: Union[str, Path]):
         torch.save(model_state, str(model_file), pickle_protocol=4)
 
     @classmethod
-    def load(cls, model: Union[str, Path]):
+    def load(cls, model_file: Union[str, Path]):
```
alanakbik: @yahshibu The `load()` method is also used to load a pre-trained tagger by passing a string, as in `SequenceTagger.load('ner')`. In this case, the string is not a model file, so I think the original variable name `model` might be better here.

yahshibu:
@alanakbik Thank you for your help! I changed the argument's name back to the original one. It passed the tests by travis-ci. Thanks!

alanakbik:
@yahshibu Thanks!

alanakbik:
:+1:

yosipk:
👍

Issue #1245, opened 2019-10-25
Startup fails non-deterministically due to transient dependency on both bson and pymongo


holgerschmeisky:
First of all, thank you very much for the great work! :)

**Description**
Flair fails to start up under certain conditions yielding this error
```
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/flair/datasets.py", line 9, in <module>
    import pymongo
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/pymongo/__init__.py", line 77, in <module>
    from pymongo.collection import ReturnDocument
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/pymongo/collection.py", line 20, in <module>
    from bson.code import Code
  File "/home/travis/virtualenv/python3.6.7/lib/python3.6/site-packages/bson/code.py", line 18, in <module>
    from bson.py3compat import abc, string_type, PY3, text_type
ImportError: cannot import name 'abc'
```

The error is described here as well in the bson project: https://github.com/py-bson/bson/issues/82

**To Reproduce**
The error happens non-deterministically if dependencies get installed in the wrong order, to fix it I just restarted my build twice and on third attempt it worked.

**Environment**
flair-0.4.4

**Root Cause**
The root cause is that Flair depends on both pymongo and bson (via hyperopt, see output of pipdeptree below). pymongo brings its own version of a bson library, so if the order of installing the dependencies is wrong, the bson library gets overwritten leading to the above error.

**Suggested Fix**
PyMongo is only used by MongoDataset (https://github.com/zalandoresearch/flair/blob/master/flair/datasets.py#L1186), I'd suggest to move it to its own PyPI module or a Gist.

pipdeptree output for flair:
```
flair==0.4.4
  - bpemb [required: >=0.2.9, installed: 0.3.0]
    - gensim [required: Any, installed: 3.8.1]
      - numpy [required: >=1.11.3, installed: 1.16.4]
      - scipy [required: >=0.18.1, installed: 1.3.1]
        - numpy [required: >=1.13.3, installed: 1.16.4]
      - six [required: >=1.5.0, installed: 1.12.0]
      - smart-open [required: >=1.8.1, installed: 1.8.4]
        - boto [required: >=2.32, installed: 2.49.0]
        - boto3 [required: Any, installed: 1.9.50]
          - botocore [required: >=1.12.50,<1.13.0, installed: 1.12.253]
[...]

PR #1296, opened 2019-11-27
GH-1245: slim down dependency tree

alanakbik:
We want to keep list of dependencies of Flair generally small to avoid errors like #1245. 

This PR removes some dependencies that are each only used for one particular feature, namely: 

- `ipython` and `ipython-genutils`, only used for visualization settings in iPython notebooks 
- `tiny_tokenizer`, used for Japanese tokenization (replaced with instructions for how to install for all users who want to use Japanese tokenizers)
- `pymongo`, used for MongoDB datasets (replaced with instructions for how to install for all users who want to use MongoDB datasets)

It also temporarily pins the version of `python-dateutil` which comes in through the `bpemb` and `transformers` dependencies and currently breaks if its version exceeds 0.2.8 (see [here](https://github.com/boto/botocore/commit/e87e7a745fd972815b235a9ee685232745aa94f9)).

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,8 +5,6 @@ env:
 python:
   - "3.6"
 install:
-  - sudo apt update -y
-  - sudo apt install mecab libmecab-dev mecab-ipadic
   - python setup.py develop -q
 before_script: cd tests
 script:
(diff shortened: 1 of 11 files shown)


alanakbik:
:+1:

stefan-it:
:+1

stefan-it:
:+1:

Issue #1298, opened 2019-11-28
Make torchvision dependency optional


basaldella:
To be removed, once it is done: Please add the appropriate label to this ticket, e.g. feature or enhancement.

**Is your feature/enhancement request related to a problem? Please describe.**

Torchvision is a mandatory dependency since 0.4.4. However, it's a very heavy library, which requires several hundreds of MB. For this reason, now Flair is unusable e.g. on the free tier of Heroku, since now the compressed blob weighs more than 800MB, while with 0.4.3 the exact same code requires less than 300MB. You can't even uninstall it since `embeddings.py` imports torchvision - no matter if you actually need it or not.

**Describe the solution you'd like**

Make torchvision optional and load it only when needed.

yosipk:
Thanks for pointing this out!

basaldella:
Thank you for your answer and for the fix, that was really quick! :)

alanakbik:
Flair 0.4.5 with optional torchvision was just released.

PR #1299, opened 2019-11-28
GH-1298: removed torchvision from requirements

yosipk:
This fixes the problem of having `torchvision` in `requirements.txt`. Now it is loaded only when needed.


yosipk:
👍

alanakbik:
:+1:

Issue #1266, opened 2019-11-08
BertEmbeddings unnecessarily moves embeddings to CPU


shoarora:
**Describe the bug**
in `_add_embeddings_internal`, we call `.detach()` and `.cpu()` unnecessarily inside the double for-loop [here](https://github.com/zalandoresearch/flair/blob/master/flair/embeddings.py#L2203).  

1. this whole block is run inside `torch.no_grad()`, so `.detach()` is basically a no-op.
2. `token.set_embedding()` should handle moving the tensor to the right device.  the only reason I can think of for moving off of GPU in the loop is that you need to make room to do something else immediately, but I can't see where that would be the case, since this entire block of code is dedicated to just aggregating the correct tensors.

Moving each layer's token embedding onto CPU one at a time is an unnecessary slowdown, when it could be done after all the tensors are concatenated together.  

**To Reproduce**
```python
# in embedding.py
for token_index, _ in enumerate(feature.tokens):
                    all_layers = []
                    for layer_index in self.layer_indexes:
                        if self.use_scalar_mix:
                            layer_output = all_encoder_layers[int(layer_index)][
                                sentence_index
                            ]
                        else:
                            layer_output = (
                                all_encoder_layers[int(layer_index)]
                                .detach()                               # <---- remove this line
                                .cpu()[sentence_index]        # <---- and this cpu() call
                            )
                        all_layers.append(layer_output[token_index])
```
Toggle those two lines of code, and do a timing comparison on `BertEmbeddings.embed()`

**Expected behavior**

I have limited devices/environments I can test on, but from what I've seen, this should give significant speed-up in runtime.  The exact improvement is highly dependent on both batch size and sentence length since it's inside this double for-loop.  

**Screenshots**

**Environment (please complete the following information):**
 - OS [e.g. iOS, Linux]: Ubuntu 16.04
 - Version: flair-0.4.4

run on an NVIDIA T4

**Additional context**

I'll try to put together a gist with the exact profiling code and outputs that I observed, but in the meantime this should be enough to begin the discussion.  Moreover, I could totally just be missing some important part of the system that explains these lines.

EDIT [gist is here](https://gist.github.com/shoarora/a8aa6ac39ee814b667907ced4bfda6c9)
I compared the flair runtime against just the huggingface runtime for the same model.  flair should take longer, since it runs the underlying huggingface model and does additional processing, but the `flair_runtime/huggingface_runtime` ratio decreases after removing the two lines mentioned

PR #1308, opened 2019-12-04
GH-1266 Remove redundant cpu() and detach() calls

shoarora:
#1266 

Currently we call `.cpu()` and `detach()` in for loop on a per-token basis.  AFAIK, this is redundant because when we assign the embedding to the token, we will allocate it to the right device/detach it as necessary.

As I outlined in the issue and the associated [gist](https://gist.github.com/shoarora/a8aa6ac39ee814b667907ced4bfda6c9), this will dramatically improve runtime, as we'll move to `cpu()` _after_ we've aggregated all the desired embeddings.  

I didn't receive confirmation on my issue that this functionality works the way that I understand it.  Hopefully, I'm not missing a key piece of the implementation that actually makes this required.

### DIFF
--- a/flair/embeddings.py
+++ b/flair/embeddings.py
@@ -2260,7 +2260,9 @@ class BertEmbeddings(TokenEmbeddings):
                                 sentence_index
                             ]
                         else:
-                            layer_output = all_encoder_layers[int(layer_index)][sentence_index]
+                            layer_output = all_encoder_layers[int(layer_index)][
+                                sentence_index
+                            ]
                         all_layers.append(layer_output[token_index])
 
                     if self.use_scalar_mix:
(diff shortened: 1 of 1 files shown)


alanakbik:
@shoarora thanks for spotting and fixing this and providing the runtime tests! Also please excuse the late reply!

Since everything is called inside `with torch.no_grad():` it makes sense that the detach is not necessary. @stefan-it do you agree? I notice also that the lines 
```python 
self.model.to(flair.device)
self.model.eval() 
```
are called each time `_add_embeddings_internal` is called - is this necessary? Isn't this already handled by `torch.nn.Module`?

alanakbik:
@shoarora again thanks for fixing this and excuse the long time to review - in my local tests this lets BERT run dramatically faster! Merging now!

alanakbik:
:+1:

Issue #1309, opened 2019-12-04
outsourcing embeddings to sqlite database at inference stage?


timnon:
Deploying e.g. the `multi-ner-fast`-model easily eats up some gig of memory just for one process. Most of this memory is probably used to hold the word embeddings of the underlying pytorch model. Has there been some thinking about outsourcing this to e.g. an sqlite database (or maybe this is already possible?). Similar to what is done in the python package magnitude. Since the inference takes anyway a few seconds, looking up some word vectors from an indexed database wouldnt add much to the inference time, but would decrease the memory consumption to almost nothing. Also, multiple processes could share the database (maybe then not a sqlite).

Hellisotherpeople:
This would be excellent as magnitude does inference faster than Flair due to the internal sqlite database representation.

timnon:
Just found a little problem in this approach: the e.g. NER-tagger in flair is heavily based on its own embedding, which are context-sensitive on a character-level, and so not just lookups of word vectors. These seem to be the ones that make the NER-tagger work so well (compared to other ones). So i basically stopped this approach. It would be interesting to see how a NER-tagger which only uses word lookups performs in comparison (basically the old standard), need to check if there is such a flair-model. In this case, building on the following snippet would allow to simply replace the `.embbedings`-attribute of a tagger with some external storage in a db:
```
from flair.models import SequenceTagger
import sqlite3
from tqdm import tqdm

class DbEmbedding():
    def __init__(self,filename='tmp.db'):
        db = sqlite3.connect(filename)
        self.db = db
    def load(self,embedding,table='embedding'):
        db = self.db
        k = embedding.vector_size
        vocab = list(embedding.vocab.keys())
        db.execute(f"DROP TABLE IF EXISTS {table};")
        db.execute(f"CREATE TABLE {table}(word text,{','.join('v'+str(i)+' float' for i in range(k))});")
        vectors_it = ( [word]+embedding.get_vector(word).tolist() for word in vocab )
        db.executemany(f"INSERT INTO {table}(word,{','.join('v'+str(i) for i in range(k))}) \
        values ({','.join(['?']*(1+k))})", tqdm(vectors_it))
        db.execute(f"CREATE INDEX {table}index ON {table}(word);")
        db.commit()
        return self
    def get_vector(self,word='house'):
        db = self.db
        cursor = db.cursor()
        cursor.execute(f"SELECT * FROM embedding WHERE word='{word}';")
        return list(cursor)

[...]

PR #1318, opened 2019-12-10
GH-1309: WorEmbeddings replacement that stores vectors in sqlite data…

timnon:
Note: reopenend this PR, since i accidentially based the old one on my master branch

#1309 

The ner-tagger of flair eats a few gigs of memory when e.g. run as the backend of a simple flask-application. This is not necessary, since the main memory consumer are word embeddings which are simple vector lookups. These can be externalized to some indexed database, e.g. sqlite, but any database does the job.

I did this by creating a class `WordEmbeedingsStore` which can be used to replace a `WordEmbeddings`-instance in a flair model via duck-typing. By using this, i was able to reduce our ner-servers memory consumption from 6gig to 600mb (10x decrease) by adding a few lines of code. It can be tested using the following lines (also in the docstring). First create a headless version of a model without word embeddings:

```
from flair.inference_utils import WordEmbeddingsStore
import pickle
tagger = SequenceTagger.load("multi-ner-fast")
WordEmbeddingsStore.create_stores(tagger)
pickle.dump(tagger, open("multi-ner-fast-headless.pickle", "wb"))
```
and then to run the stored headless model without word embeddings, use:
```
tagger = pickle.load(open("multi-ner-fast-headless.pickle", "rb"))
WordEmbeddingsStore.load_stores(tagger)
text = "Schade um den Ameisenbären. Lukas Bärfuss veröffentlicht Erzählungen aus zwanzig Jahren."
sentence = Sentence(text)
tagger.predict(sentence)
```
Please give some advice where to add this functionality, there seems to be no place for inference-related stuff, therefore i created `inference_utils.py`, but some other place might be more appropriate.

I like the current structure such that it is a very light integration via duck-typing, so no flair base classes need to get changed, but a tighter integration might also be possible.

### DIFF
--- a/flair/inference_utils.py
+++ b/flair/inference_utils.py
@@ -1,0 +1,131 @@
+import flair
+from flair.embeddings import WordEmbeddings
+import sqlite3
+import torch
+import re
+import os
+from tqdm import tqdm
+
+
+class WordEmbeddingsStore:
+    """
+    class to simulate a WordEmbeddings class from flair.
+
+    Run this to generate a headless (without word embeddings) model as well a stored word embeddings:
+
+    >>> from WordEmbeddingsStore import WordEmbeddingsStore
+    >>> from flair.models import SequenceTagger
+    >>> import pickle
+    >>> tagger = SequenceTagger.load("multi-ner-fast")
+    >>> WordEmbeddingsStore.create_stores(tagger)
+    >>> pickle.dump(tagger, open("multi-ner-fast-headless.pickle", "wb"))
+
+    Then this can be used as follows:
+
+    >>> from flair.data import Sentence
+    >>> tagger = pickle.load(open("multi-ner-fast-headless.pickle", "rb"))
+    >>> WordEmbeddingsStore.load_stores(tagger)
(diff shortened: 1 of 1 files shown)


timnon:
one more thing, the outsourcing is not restricted to sqlite, this could be handled by any database or also reddis, so a general "outsourcing connector" to different backends would be appropriate

alanakbik:
@timnon thanks a lot - this is a good solution to reduce the memory footprint of `WordEmbeddings` which many people will find useful! I've made a few small changes: the `db` variable was not always `self.db` causing errors and I've changed the zero tensor to avoid a torch warning.

I think a more integrated solution of this would be great, but I'm not sure how to best do this. I'll think a bit about this and let you know! Thanks again!

timnon:
Cool, thanks for the fixes, tested and everything seems to be good. As written above, any backend to store vectors would be good, but the elegance of sqlite is to have all in files, which suits the current storing mechanism.

alanakbik:
Great, will merge now! Thanks again!

alanakbik:
:+1:

PR #1321, opened 2019-12-17
Replace sklearn by scikit-learn

dilshatu:
Change the sklearn dependency to scikit-learn (as suggested in the description of sklearn)

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -6,7 +6,7 @@ tqdm>=4.26.0
 segtok>=1.5.7
 matplotlib>=2.2.3
 mpld3==0.3
-sklearn
+scikit-learn==0.21.3
 sqlitedict>=1.6.0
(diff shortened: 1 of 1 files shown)


alanakbik:
:+1:

alanakbik:
@dilshatu thanks for fixing this! Also please excuse the delayed response!

dilshatu:
Thank you!

dilshatu:
@alanakbik btw could you please advice me on a new release date, so these changes are applied?

alanakbik:
@dilshatu I'm not sure exactly when, but I want to do a release fairly soon since the current install is broken and we've removed a lot of problematic dependencies. In the meantime, you can always install the latest version via 

```console
pip install --upgrade git+https://github.com/zalandoresearch/flair.git 
```
which will contain all the fixes.

PR #1334, opened 2020-01-05
Fix off-by-one error in progress logging 

jjjamie:
Now logs after multiples of `modulo` batches rather than 1, 1+`modulo`, etc
Prevents samples/sec being too large by factor of `modulo` for batch_no 0 and division by zero crash if 0th batch is too quick.


alanakbik:
@jjjamie thanks for fixing this!

PR #1341, opened 2020-01-10
Fix typo and annotation

djstrong:
Actually, the code does not support list of sentences.


alanakbik:
@djstrong thanks for fixing this!

alanakbik:
:+1:

Issue #1344, opened 2020-01-11
Prioritizing classes and precission/recall for sequence tagger


klasocki:
Following @apohllo 's idea, I implemented two ways of prioritizing precission/recall during training

- Using [F-beta ](http://www.marcelonet.com/snippets/machine-learning/evaluation-metrix/f-beta-score) as the evaluation metric, has impact on logging and annealing schedule

- Adding class weights to cross-entropy loss, enabling class prioritization (e.g less weight to errors on 'O' tag if we want more recall). Works only if using standard loss - no CRF

I'll submit a pull request in case you'd want to include these changes in your library :)

alanakbik:
New version of flair released that contains this feature - thanks again @klasocki!

PR #1345, opened 2020-01-11
GH-1344: Prioritizing classes

klasocki:
Add possibility for including class weights in cross-entropy loss (when not using CRF) and using F-Beta as the evaluation metric

### DIFF
--- a/flair/models/sequence_tagger_model.py
+++ b/flair/models/sequence_tagger_model.py
@@ -82,7 +82,7 @@ class SequenceTagger(flair.nn.Model):
         rnn_type: str = "LSTM",
         pickle_module: str = "pickle",
         beta: float = 1.0,
-        weights: Dict = None,
+        weights: Dict[str, float] = None,
     ):
         """
         Initializes a SequenceTagger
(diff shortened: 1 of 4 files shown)


djstrong:
Weights can be added also to `TextClassifier`.

djstrong:
Class weighting is important in macro metrics.

alanakbik:
@klasocki thanks for adding this and please excuse the delay in checking this PR! I've ran some local tests with different parameters and it works well for prioritizing classes and controlling the precision/recall trade-off so lot's of people will surely find this useful!

alanakbik:
:+1:

klasocki:
> @klasocki thanks for adding this and please excuse the delay in checking this PR! I've ran some local tests with different parameters and it works well for prioritizing classes and controlling the precision/recall trade-off so lot's of people will surely find this useful!

No problem whatsoever! Thank you for caring about the quality of Flair. And I'm very glad you found it useful 😄

BasLaanPAM:
Thank you for adding this feature, it is very helpful. I was wondering if there is a specific reason it was only added in cross-entropy loss and not the viterbi loss (using CRF)? Having a quick look at the [code ](https://github.com/flairNLP/flair/blob/8d27a383810455cb45d650e7fe2003384780ae84/flair/models/sequence_tagger_utils/viterbi.py#L79) we should be able to multiple this loss function by weights as well.

PR #1347, opened 2020-01-13
Add some modifications

yahshibu:
I've added some trivial modifications.
- modify a mismatch between an actual argument name and its corresponding docstring
- delete an unused local variable
- delete unused import statements
- change to use a non-deprecated function
- simplify chained comparisons
- use "not in" for membership test
- modify some instance methods to static methods

I would be happy if my commits would be accepted.

### DIFF
--- a/flair/data.py
+++ b/flair/data.py
@@ -621,7 +621,7 @@ class Sentence(DataPoint):
 
         # log a warning if the dataset is empty
         if text == "":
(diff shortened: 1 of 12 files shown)


alanakbik:
@yahshibu this is great - thanks for adding these improvements!

alanakbik:
:+1:

PR #1353, opened 2020-01-14
Make load_big_file work with read-only file

jjjamie:
Use "rb" rather than "r+b" and specify access=mmap.ACCESS_READ

I would like to be able to read model files on a read-only share but current code crashes with permissions problems - this fixes for me

### DIFF
--- a/flair/file_utils.py
+++ b/flair/file_utils.py
@@ -22,16 +22,16 @@ import flair
 logger = logging.getLogger("flair")
 
 
-def load_big_file(f):
+def load_big_file(f: str) -> mmap.mmap:
     """
     Workaround for loading a big pickle file. Files over 2GB cause pickle errors on certin Mac and Windows distributions.
     :param f:
     :return:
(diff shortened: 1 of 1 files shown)


flair/file_utils.py
```diff
@@ -22,16 +22,16 @@
 logger = logging.getLogger("flair")
 
 
-def load_big_file(f):
+def load_big_file(f: str) -> mmap.mmap:
```
jjjamie: And some type annotations!

djstrong:
It must be made as an option. Models are also loaded for further training.

alanakbik:
@jjjamie thanks for adding this! 

@djstrong I've done some tests in resuming training on checkpoints or shipped models. It still works with the change because new models after training are written out as separate files.

djstrong:
@alanakbik Ok, you are right!

alanakbik:
:+1:

PR #1361, opened 2020-01-18
set whitespace_after in UniversalDependenciesDataset

djstrong:
Set `whitespace_after` in `UniversalDependenciesDataset`, so `to_plain_string` method is returning original sentence.

Issue #1021, opened 2019-08-22
Flair embeddings are computed over tokenized string instead of original


djstrong:
**Describe the bug**
Flair embeddings in `Sentence` are computed over `to_tokenized_string` instead of `to_plain_string`. It is inconsistent with trained language model. Sometimes whitespaces are important.

**Expected behavior**
Flair embeddings in `Sentence` should be computed over `to_plain_string`.

**Environment:**
 - Version: flair-0.4.2

**Additional context**
I have made the question here: https://github.com/zalandoresearch/flair/pull/72#issuecomment-523068227 but an issue is more appropriate place.

PR #1362, opened 2020-01-18
GH-1021: Compute Flair embeddings over original string and without spaces after token

djstrong:
Now, space is inserted after every token despite `whitespace_after` value. So sentences being embedded do not look the same as normal text on which Flair language models (embeddings) were trained.
Also, hidden state is taken as embedding after consuming a space.

This PR computes embeddings over text with original white spaces.
I am not sure if this approach is better in downstream tasks, but it is consistent with trained Flair language model.

Initial experiments shows that calculating token embedding without space gives better results. Using `whitespace_after` information is a little worse. First epoch of training with fixed seeds:
```
2020-01-18 21:54:41,014 epoch 1 - iter 43/431 - loss 12.58697907 - samples/sec: 104.66
2020-01-18 21:54:54,871 epoch 1 - iter 86/431 - loss 8.39764844 - samples/sec: 99.44
2020-01-18 21:55:08,917 epoch 1 - iter 129/431 - loss 6.48929740 - samples/sec: 98.12
2020-01-18 21:55:23,167 epoch 1 - iter 172/431 - loss 5.37756264 - samples/sec: 96.67
2020-01-18 21:55:36,985 epoch 1 - iter 215/431 - loss 4.63144744 - samples/sec: 99.72
2020-01-18 21:55:50,898 epoch 1 - iter 258/431 - loss 4.14733414 - samples/sec: 99.01
2020-01-18 21:56:04,915 epoch 1 - iter 301/431 - loss 3.76650934 - samples/sec: 98.27
2020-01-18 21:56:18,543 epoch 1 - iter 344/431 - loss 3.50686799 - samples/sec: 101.09
2020-01-18 21:56:32,710 epoch 1 - iter 387/431 - loss 3.27343043 - samples/sec: 97.24
2020-01-18 21:56:46,579 epoch 1 - iter 430/431 - loss 3.07420425 - samples/sec: 99.32
2020-01-18 21:56:46,824 ----------------------------------------------------------------------------------------------------
2020-01-18 21:56:46,824 EPOCH 1 done: loss 3.0724 - lr 0.1000
2020-01-18 21:57:00,554 DEV : loss 0.501313328742981 - score 0.9818
```
vs. with space after every token:
```
2020-01-18 21:50:48,533 epoch 1 - iter 43/431 - loss 12.42667790 - samples/sec: 103.46
2020-01-18 21:51:02,377 epoch 1 - iter 86/431 - loss 8.29389394 - samples/sec: 99.53
2020-01-18 21:51:16,640 epoch 1 - iter 129/431 - loss 6.39272680 - samples/sec: 96.59
2020-01-18 21:51:31,172 epoch 1 - iter 172/431 - loss 5.28999498 - samples/sec: 94.78
2020-01-18 21:51:45,272 epoch 1 - iter 215/431 - loss 4.54830917 - samples/sec: 97.70
2020-01-18 21:51:59,159 epoch 1 - iter 258/431 - loss 4.06829357 - samples/sec: 99.19
2020-01-18 21:52:13,226 epoch 1 - iter 301/431 - loss 3.69254654 - samples/sec: 97.92
2020-01-18 21:52:26,832 epoch 1 - iter 344/431 - loss 3.43669727 - samples/sec: 101.30
2020-01-18 21:52:41,262 epoch 1 - iter 387/431 - loss 3.21065094 - samples/sec: 95.45
2020-01-18 21:52:55,442 epoch 1 - iter 430/431 - loss 3.01424922 - samples/sec: 97.19
2020-01-18 21:52:55,693 ----------------------------------------------------------------------------------------------------
2020-01-18 21:52:55,693 EPOCH 1 done: loss 3.0125 - lr 0.1000
2020-01-18 21:53:09,711 DEV : loss 0.46862271428108215 - score 0.9831
```
vs. original code (embedding includes space after):
```
2020-01-18 19:11:18,763 epoch 1 - iter 43/431 - loss 13.77513606 - samples/sec: 100.09
2020-01-18 19:11:32,758 epoch 1 - iter 86/431 - loss 9.86612977 - samples/sec: 98.49
2020-01-18 19:11:46,444 epoch 1 - iter 129/431 - loss 7.84115263 - samples/sec: 100.66
2020-01-18 19:12:00,549 epoch 1 - iter 172/431 - loss 6.57365022 - samples/sec: 97.65
2020-01-18 19:12:14,326 epoch 1 - iter 215/431 - loss 5.71325676 - samples/sec: 99.98
2020-01-18 19:12:28,293 epoch 1 - iter 258/431 - loss 5.13455269 - samples/sec: 98.62
2020-01-18 19:12:42,390 epoch 1 - iter 301/431 - loss 4.66878698 - samples/sec: 97.71
2020-01-18 19:12:55,841 epoch 1 - iter 344/431 - loss 4.35207718 - samples/sec: 102.41
2020-01-18 19:13:09,813 epoch 1 - iter 387/431 - loss 4.08306071 - samples/sec: 98.59
2020-01-18 19:13:23,596 epoch 1 - iter 430/431 - loss 3.83598182 - samples/sec: 99.94
2020-01-18 19:13:23,847 ----------------------------------------------------------------------------------------------------
2020-01-18 19:13:23,847 EPOCH 1 done: loss 3.8318 - lr 0.1000
2020-01-18 19:13:37,732 DEV : loss 0.9926474690437317 - score 0.9577
```

You can test it on example `train.py` with POS tagging on UD after applying https://github.com/flairNLP/flair/pull/1361.

### DIFF
--- a/flair/embeddings.py
+++ b/flair/embeddings.py
@@ -2048,8 +2048,8 @@ class FlairEmbeddings(TokenEmbeddings):
             for i, sentence in enumerate(sentences):
                 sentence_text = sentence.to_plain_string()
 
-                offset_forward: int = len(start_marker)
-                offset_backward: int = len(sentence_text) + len(start_marker)
+                offset_forward: int = len(start_marker) - 1
+                offset_backward: int = len(sentence_text) + len(start_marker) - 1
 
                 for token in sentence.tokens:
 
--- a/flair/embeddings.py
+++ b/flair/embeddings.py
@@ -2031,7 +2031,7 @@ class FlairEmbeddings(TokenEmbeddings):
         with gradient_context:
 
             # if this is not possible, use LM to generate embedding. First, get text sentences
-            text_sentences = [sentence.to_tokenized_string() for sentence in sentences]
+            text_sentences = [sentence.to_plain_string() for sentence in sentences]
 
             start_marker = "\n"
             end_marker = " "
@@ -2046,7 +2046,7 @@ class FlairEmbeddings(TokenEmbeddings):
 
             # take first or last hidden states from language model as word representation
             for i, sentence in enumerate(sentences):
-                sentence_text = sentence.to_tokenized_string()
+                sentence_text = sentence.to_plain_string()
 
                 offset_forward: int = len(start_marker)
                 offset_backward: int = len(sentence_text) + len(start_marker)
@@ -2062,9 +2062,9 @@ class FlairEmbeddings(TokenEmbeddings):
 
                     embedding = all_hidden_states_in_lm[offset, i, :]
 
-                    # if self.tokenized_lm or token.whitespace_after:
-                    offset_forward += 1
-                    offset_backward -= 1
+                    if token.whitespace_after:
+                        offset_forward += 1
+                        offset_backward -= 1
 
                     offset_backward -= len(token.text)
 


djstrong:
Embedding offsets were tested with:
```
import torch

from flair.data import Token, Sentence

ff = FlairEmbeddings('pl-forward')
fb = FlairEmbeddings('pl-backward')

for whitespace_after in [False, True]:
    s1 = Sentence()
    s1.add_token(Token('Ala', whitespace_after=whitespace_after))
    s1.add_token(Token('ma', whitespace_after=whitespace_after))
    s1.add_token(Token('kota', whitespace_after=whitespace_after))
    ff.embed(s1)

    s2 = Sentence()
    s2.add_token(Token('Ala', whitespace_after=whitespace_after))
    s2.add_token(Token('ma', whitespace_after=whitespace_after))
    s2.add_token(Token('Xota', whitespace_after=whitespace_after))
    ff.embed(s2)

    a = s1.tokens[1].get_embedding()
    b = s2.tokens[1].get_embedding()

    assert all(torch.eq(a, b))

    s2 = Sentence()
    s2.add_token(Token('Ala', whitespace_after=whitespace_after))
    s2.add_token(Token('mX', whitespace_after=whitespace_after))
    s2.add_token(Token('kota', whitespace_after=whitespace_after))
    ff.embed(s2)

    a = s1.tokens[1].get_embedding()
    b = s2.tokens[1].get_embedding()

    assert not all(torch.eq(a, b))

    s1 = Sentence()
    s1.add_token(Token('Ala', whitespace_after=whitespace_after))
    s1.add_token(Token('ma', whitespace_after=whitespace_after))
    s1.add_token(Token('kota', whitespace_after=whitespace_after))
    fb.embed(s1)

    s2 = Sentence()
    s2.add_token(Token('AlX', whitespace_after=whitespace_after))
    s2.add_token(Token('ma', whitespace_after=whitespace_after))
    s2.add_token(Token('kota', whitespace_after=whitespace_after))
    fb.embed(s2)

    a = s1.tokens[1].get_embedding()
    b = s2.tokens[1].get_embedding()

    assert all(torch.eq(a, b))

    s2 = Sentence()
    s2.add_token(Token('Ala', whitespace_after=whitespace_after))
    s2.add_token(Token('Xa', whitespace_after=whitespace_after))
    s2.add_token(Token('kota', whitespace_after=whitespace_after))
    fb.embed(s2)

    a = s1.tokens[1].get_embedding()
    b = s2.tokens[1].get_embedding()

    assert not all(torch.eq(a, b))
```

alanakbik:
(closed and reopened to re-trigger Travis build)

alanakbik:
@djstrong this is really interesting - could you post the script you used to get the above experiment numbers on UD pos tags? Did you use the EnglishUD corpus and the 'news-forward' and 'news-backward' Flair embeddings?

djstrong:
I tested it on Polish. The script:
```
import random
from typing import List

import numpy as np
import torch

import flair.datasets
from flair.data import Corpus
    TokenEmbeddings,
    StackedEmbeddings,
    FlairEmbeddings,
)


torch.manual_seed(0)
random.seed(0)
np.random.seed(0)

# 1. get the corpus
corpus: Corpus = flair.datasets.UD_POLISH()
print(corpus)

tag_type = "upos"

print(tag_dictionary.idx2item)

# initialize embeddings
embedding_types: List[TokenEmbeddings] = [
    # WordEmbeddings("glove"),
    # comment in this line to use character embeddings
    # CharacterEmbeddings(),
    # comment in these lines to use contextual string embeddings
    #
    FlairEmbeddings('pl-forward'),
    FlairEmbeddings('pl-backward'),
]

embeddings: StackedEmbeddings = StackedEmbeddings(embeddings=embedding_types)

# initialize sequence tagger

tagger: SequenceTagger = SequenceTagger(
    hidden_size=256,
    embeddings=embeddings,
    tag_type=tag_type,
    use_crf=True,
)

# initialize trainer

trainer.train(
    "resources/taggers/example-ner",
    learning_rate=0.1,
    mini_batch_size=32,
    max_epochs=1,
    shuffle=False,
    embeddings_storage_mode="gpu"
)
```
I am not sure if I described the problem correctly:
- Flair computes token embedding in both direction after feeding space instead of after last character of a token
- Flair computes embedding over tokenized sentence instead of original sentence - it might not be wrong

alanakbik:
Ah yes this makes sense. To clarify, the above numbers you get for Polish are: 

- whitespace-tokenized sentence, token embeddings extracted after extra whitespace (original): 0.9577
- whitespace-tokenized sentence, token embeddings extracted at last token character: 0.9831
- plain text sentence, token embeddings extracted at last token character: 0.9818

Is this correct?

djstrong:
Yes, thank you.

alanakbik:
Ok thanks! 

I guess it's a bit counterintuitive that whitespace-tokenized is better than true text if the LM was trained on normal, untokenized text. However, from the Polish UD documentation it seems that tokenization can split of endings with different POS tags (wyprodukowalibyśmy = wyprodukowali + by + śmy). Maybe, since we want to classify each of these parts with a POS tag, it somehow helps the LM to have whitespaces between them even if this is not how the model was trained. 

The other observation that it is better to extract embeddings at the end of each token rather than at the whitespace after is really interesting. I can a quick test on English POS and from an initial experiment this also seems to work better here. This is interesting because the English LM was trained using the 1B word corpus, which is in fact whitespace tokenized. So again it is strange that the model works better this way even though the LM was trained differently. 

However, bigger changes are required to merge this PR. The problem is that all models that use Flair embeddings and are included in Flair were trained to expect token embeddings extracted at the whitespace. If we change this class without also retraining the included models, they will no longer work as well as before. So I would add these changes at a later time, together with retraining all models for the next major version release.

djstrong:
> I guess it's a bit counterintuitive that whitespace-tokenized is better than true text if the LM was trained on normal, untokenized text. However, from the Polish UD documentation it seems that tokenization can split of endings with different POS tags (wyprodukowalibyśmy = wyprodukowali + by + śmy). Maybe, since we want to classify each of these parts with a POS tag, it somehow helps the LM to have whitespaces between them even if this is not how the model was trained.

Yes, in Polish some words are split. Here, the scores difference might not be significant (0.9831 vs 0.9818).

> The other observation that it is better to extract embeddings at the end of each token rather than at the whitespace after is really interesting. I can a quick test on English POS and from an initial experiment this also seems to work better here. This is interesting because the English LM was trained using the 1B word corpus, which is in fact whitespace tokenized. So again it is strange that the model works better this way even though the LM was trained differently.

Have you compared here `whitespace-tokenized sentence, token embeddings extracted after extra whitespace (original)` with `whitespace-tokenized sentence, token embeddings extracted at last token character`? If yes, then the LM is trained the same as embeddings being calculated.

alanakbik:
I compared `whitespace-tokenized sentence, token embeddings extracted after extra whitespace (original)` with `plain text sentence, token embeddings extracted at last token character` but it was only one run each for 5 epochs and I should really also compare the third option.

djstrong:
I am not sure if this might be correct, but let's say we have a token `dog`. State of forward LM `at last token character` need to remember whole word to predict some suffixes of the `dog` or mark end of a word by space. State of forward LM `after extra whitespace` might forget the word `dog` and somehow generalize to a mammal, an animal or a noun and predict form a large group of appropriate verbs.

Maybe in this task (POS tagging) the first approach (`embeddings at last token character`) is better, but in more semantic task the second approach will be better.

alanakbik:
Yes that could be it. Maybe one could try training a model with access to both states and let the model decide which it wants to use (or even a combination of the two).

alanakbik:
@djstrong sorry for the long delay on this. I've pushed an implementation in which both parameters are optional (whether to whitespace tokenize and whether to use the last hidden state or the one after, or both). This includes a check that sets the parameters to the original setting so it is compatible with serialized models. 

Using 'news-forward-fast' and 'news-backward-fast' in the default setup over two runs each, I get the following: 

| Task       | Offset_mode        | Tokenized_LM | F1            |
|------------|--------------------|--------------|---------------|
| NER - CoNLL-03   |                    |              |               |
|            | With_whitespace    | True         | 96.44 +- 0.05 |
|            | Without_whitespace | True         | 96.66 +- 0.01 |
|            | Both               | True         | **96.75 +- 0.02** |
| POS - UD English |                    |              |               |
|            | With_whitespace    | True         | **92.62 +- 0.10** |
|            | Without_whitespace | True         | 92.22 +- 0.00 |
|            | Both               | True         | 92.27 +- 0.03 |

This confirms your observations: POS-tagging works better if we take the last hidden state, while NER works better if we take the one after the whitespace. What is interesting that if we give it both states concatenated, NER is also worse than if we just give it the hidden state after the whitespace. Here, all models are tokenized since that is how 1BW corpus is formatted.

Wildly guessing and going back to your `dog` example, it could be that for NER it is actually a good thing to forget all the morphological information after the whitespace as otherwise the model might overfit to non-relevant information.

djstrong:
Thank you, I will test it.

djstrong:
There is a problem with tokenized_lm=False, offset_mode='with_whitespace' and tokens with whitespace_after=False.

alanakbik:
Hm, in those cases it would still offset by 1 and pass a non-whitespace character. For instance a word before a comma would then get the state from the comma. I think that is still ok since still the model is trying to predict how the sentence continues from this point on. And this way the offset is consistent across all tokens. 

The alternative would be to do a check whether a whitespace follows and only do an offset if so, but then the hidden state passed onto the next layer would be inconsistent in that sometimes it gets the state after the whitespace and sometimes directly at the end of a word.

djstrong:
Yes, with a comma it makes sense. 

I have thought about other languages where a tokenizer splits some words into more tokens (in Polish `wyprodukowalibyśmy = wyprodukowali + by + śmy`). So, doing NER in Polish we should use `tokenized_lm=False`. Your results show that `offset_mode='with_whitespace'` should be better, but for this language it may be worse.

alanakbik:
Ah yes that could be, though would need to be tested because it might be that this type of splitting does not affect proper nouns too much.

djstrong:
Yes, you are right. I don't have a dataset to test it. However, for Polish POS and NER are better `without_whitespace`.

UD_POLISH UPOS 5 epochs
with_whitespace 0.9814
without_whitespace 0.9862

WIKINER_POLISH 1 epoch
with_whitespace 0.8538
without_whitespace 0.8691

djstrong:
Could you revise your results? For English, better is Without_whitespace for NER, and With_whitespace for POS, but your conclusions are the opposite.

alanakbik:
Ah yes updated the table. And thanks for sharing your results!

I'm also now running experiments with using the hidden state at the beginning of a word as well, i.e. the state that the LM would use to predict the first character of a word. Could be for tasks such as POS that there is some interesting information there.

djstrong:
I have done something similar:

LM_backward_start +LM_forward_end (original) | 88.52
-- | --
LM_backward_end +LM_forward_end (the same as in segmentation task) | 89.82
LM_backward_end +LM_forward_start (does not see the token) | 86.16
LM_backward_start +LM_forward_start (similar as in segmentation task, but for detection start of token) | 88.57
LM_backward_start +LM_forward_start +LM_backward_end +LM_forward_end | 90.36

alanakbik:
Wow those are great results! What task / embedding dimensionality is this with? Have you tested if this improvement holds across different tasks?

djstrong:
I have tested it only on a specific task and the nature of the task explains why "LM_backward_end +LM_forward_end" is better. It was Polish embeddings from Flair, so I guess 2048.

alanakbik:
@djstrong I've merged this functionality in another PR that make it configurable and is backwards-compatible with earlier models. Thanks again!

Michael95-m:
@djstrong @alanakbik, please let me ask u one thing about this?? I've trained flair language models(forward and backward) for Burmese. In that training, I added no space between each Burmese characters and only space is inserted between English character and Burmese character and between English words. So, if I want to use embedding from these language models, which settings should I use?? Should I make False for with_whitespace and tokenized_lm?? Thank u..

alanakbik:
Good question on both. I think the only way to know for sure is to try out the different combinations. Intuitively, I think both `False` is probably the best option!

Michael95-m:
Thanks for yr kind answer @alanakbik

Issue #1388, opened 2020-01-27
ValueError: invalid literal for int() with base 10: '0+cpu' | for info in torch.__version__.split('.'))


rahulvsharan:
Issue.:
2020-01-27 14:15:54,846 loading file C:\Users\1025289\.flair\models\en-pos-ontonotes-fast-v0.4.pt
Traceback (most recent call last):
  File ".\fmain.py", line 16, in <module>
    main()
  File ".\fmain.py", line 4, in main
    intent_obj = GET_INTENT(b_preprocess_query=True)
  File "C:\work\python\proj_fai\get_intent.py", line 10, in __init__
    super().__init__(b_preprocess_query=b_preprocess_query)
  File "C:\work\python\proj_fai\model_flair.py", line 24, in __init__
    self._load_flair_model(flair_model)
  File "C:\work\python\proj_fai\model_flair.py", line 30, in _load_flair_model
    self._seq_tag = SequenceTagger.load(flair_model)
  File "C:\Users\1025289\AppData\Local\Programs\Python\Python37\lib\site-packages\flair\nn.py", line 88, in load
    model = cls._init_model_with_state_dict(state)
  File "C:\Users\1025289\AppData\Local\Programs\Python\Python37\lib\site-packages\flair\models\sequence_tagger_model.py", line 273, in _init_model_with_state_dict
    loss_weights=weights,
  File "C:\Users\1025289\AppData\Local\Programs\Python\Python37\lib\site-packages\flair\models\sequence_tagger_model.py", line 217, in __init__
    self.to(flair.device)
  File "C:\Users\1025289\AppData\Local\Programs\Python\Python37\lib\site-packages\torch\nn\modules\module.py", line 425, in to
    return self._apply(convert)
  File "C:\Users\1025289\AppData\Local\Programs\Python\Python37\lib\site-packages\torch\nn\modules\module.py", line 201, in _apply
    module._apply(fn)
  File "C:\Users\1025289\AppData\Local\Programs\Python\Python37\lib\site-packages\torch\nn\modules\module.py", line 201, in _apply
    module._apply(fn)
  File "C:\Users\1025289\AppData\Local\Programs\Python\Python37\lib\site-packages\torch\nn\modules\module.py", line 201, in _apply
    module._apply(fn)
  File "C:\Users\1025289\AppData\Local\Programs\Python\Python37\lib\site-packages\flair\models\language_model.py", line 394, in _apply       
    for info in torch.__version__.split('.'))
[...]

PR #1393, opened 2020-01-28
GH-1388: fix issue where flair was crashing for cpu only version of pytorch

mittalsuraj18:
- replace +cpu part of the version with empty string during access


mittalsuraj18:
Fixes Issue https://github.com/flairNLP/flair/issues/1388

alanakbik:
@mittalsuraj18 thanks for fixing this! I'm not sure how to reproduce the error locally, but it looks like it fixes the problem so I'll merge!

PR #1413, opened 2020-02-07
GH-1414: Improve performance for loading not-in-memory corpus

david-westreicher:
When loading a "not-in-memory" corpus we don't need to construct the whole sentence, because we will read the sentence when necessary in `get_item`.


alanakbik:
Hi @david-westreicher this works great - speeds up loading corpora into memory quite a bit. Thanks for adding this!

alanakbik:
:+1:

PR #1416, opened 2020-02-07
Padding entry conversion from CPU to flair_device to avoid CUDA embedding errors

gitlost-murali:
Torch concatenate throws error if we try to concatenate "GPU" tensors with "CPU" tensors. This is happening in  Line 3151. `word_embeddings_tensor = torch.cat(word_embeddings, 0).to(flair.device)` 
throws ERROR, when some entries are in GPU (embeddings of elements) and some (Padding zeros) in CPU. Above change will bring all components in "word_embedings" list to common space (Flair Device).


alanakbik:
:+1:

gitlost-murali:
Thanks! 
I will take note of `DocumentRNNEmbeddings`.

PR #1419, opened 2020-02-07
Remove urllib3 from requirements.txt

jjjamie:
Prevents urllib3 being downgraded in my environment (nothing in flair depends on urllib3)

At least I don't think it's needed... have I missed something? The line was added in https://github.com/flairNLP/flair/commit/3336f8f1fa93204a43fb513fc76c80b3a96ae6c5

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -14,5 +14,4 @@ transformers>=2.3.0
 bpemb>=0.2.9
 regex
 tabulate
-urllib3<1.25,>=1.20
 langdetect


alanakbik:
(closed and reopened PR to re-trigger Travis build)

alanakbik:
@jjjamie thanks for spotting and fixing this! tbh I don't remember why we needed to make the version explicit, but it seems to run now without so whatever conflict it caused seems to be resolved. Fewer dependencies is always a good thing, so I'll merge!

alanakbik:
:+1:

PR #1428, opened 2020-02-14
fix encoding error for custom embeddings

abeermohamed1:
---> 51     WordEmbeddings(data_folder+'/model.bin',True),
3 frames
/usr/local/lib/python3.6/dist-packages/gensim/utils.py in any2unicode(text, encoding, errors)
    353     if isinstance(text, unicode):
    354         return text
--> 355     return unicode(text, encoding, errors=errors)
    356 
    357 

UnicodeDecodeError: 'utf-8' codec can't decode byte 0x80 in position 0: invalid start byte

### DIFF
--- a/flair/embeddings.py
+++ b/flair/embeddings.py
@@ -198,7 +198,7 @@ class StackedEmbeddings(TokenEmbeddings):
 class WordEmbeddings(TokenEmbeddings):
     """Standard static word embeddings, such as GloVe or FastText."""
 
-    def __init__(self, embeddings: str, field: str = None):
+    def __init__(self, embeddings: str, field: str = None,ignore: bool = False):
         """
         Initializes classic word embeddings. Constructor downloads required files if not there.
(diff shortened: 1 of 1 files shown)


flair/embeddings.py
```diff
@@ -317,9 +317,14 @@ def __init__(self, embeddings: str, field: str = None):
...
-                str(embeddings), binary=True
-            )
+            if(ignore_unicode_error):
+                self.precomputed_word_embeddings = gensim.models.KeyedVectors.load_word2vec_format(
+                    str(embeddings), binary=True, unicode_error=True
+                )
+            else:
+                self.precomputed_word_embeddings = gensim.models.KeyedVectors.load_word2vec_format(
+                    str(embeddings), binary=True, unicode_error=False
+                )
```
zafercavdar, line 327: Why not using `unicode_error=ignore_unicode_error` directly instead of if-else?

alanakbik:
Could you send me an example .bin that has this error to test the PR?

PR #1511, opened 2020-04-04
remove duplicated code

djstrong:


alanakbik:
@djstrong thanks for spotting and correcting this!

Issue #1521, opened 2020-04-09
POS Tagger model for Malayalam Language


sabiqueqb:
I have found that there is no POS Tagger model available for the Malayalam language currently. I have trained a model for UPOS Tagging and XPOS Tagging  for Malayalam language and would like to add it to Flair repository. 

The model has trained with more than 30000 sentences and have the average accuracy as follows:
* UPOS Model - 87%
* XPOS Model - 83%

alanakbik:
Hello @sabiqueqb thanks - we'd appreciate the contribution! Would you like to host the model yourself or should we host it on our servers? If the former, you could put in a PR for the new model. If the latter, if you send me the model, I'll put it onto our server and add the PR!

sabikla:
Thank you @alanakbik  for your quick response. I will host the model and will open a PR with the required changes.

sabiqueqb:
Thank you @alanakbik for the quick response. I will host the model and will open a PR with the required changes.

sabiqueqb:
Hi @alanakbik, I have hosted the models and have changed the code accordingly. I have raised a PR (https://github.com/flairNLP/flair/pull/1522) against this issue. Thank You

alanakbik:
Malayalam models are now part of Flair 0.5, just released. Thanks @sabiqueqb!

PR #1522, opened 2020-04-13
GH-1521: Add Malayalam POS and XPOS tagger model

sabiqueqb:
This pull request add Malayalam UPOS and XPOS models.

The models are trained with more than 30000 sentences and have the average accuracy as follows:
* UPOS Model - 87%
* XPOS Model - 83%


alanakbik:
@sabiqueqb thanks for adding this!

Issue #1524, opened 2020-04-16
LM Training: Structure of input files


leonweber:
I would like to train a flair LM and I am a bit confused about the expected format of the input files. Does the LM training assume sentence segmented text? Is it one document or one sentence per line? How are documents separated?

djstrong:
It is raw, unstructured text.

alanakbik:
Yes, there is no formatting required since the entire corpus is modeled as a very long sequence of characters. So even new lines are just another character in this sequence. This means that you can put sentences or documents in lines, but we generally just leave new lines as they appear in the original data. 

This also means that there is no explicit separation of documents if you put a number of them into a text file. We typically just cat them together. If you want to give the language model an explicit signal for new documents (rather than hoping that it learns this by itself) you can add any explicit separator between documents, like "[SEP]" or so. The language model will then learn that this character sequence denotes a separation of documents.

leonweber:
Thanks for clearing that up @alanakbik and @djstrong. Do you think that this should be explicitly mentioned in [Tutorial 9](https://github.com/flairNLP/flair/blob/master/resources/docs/TUTORIAL_9_TRAINING_LM_EMBEDDINGS.md)?

alanakbik:
Yes, since we get this question a lot. Care to do a PR? ;)

leonweber:
done: https://github.com/flairNLP/flair/pull/1525

PR #1525, opened 2020-04-17
GH-1524: Specify the structure of the input files for LM training in …

leonweber:
This PR states the expected structure of the input files (there is none) for LM training in the docs


alanakbik:
Thanks for adding this!

Issue #1529, opened 2020-04-19
Trapped in an infinite loop in the to_original_text method


psorianom:
**Describe the bug**
When using a faulty tokenizer, producing faulty span indices, the `Sentence` class method
`to_original_text`  is bound to enter an infinite loop. While we should not have faulty tokenizers, 
it may happen, as it happened to me. This error is related to this closed [PR](https://github.com/flairNLP/flair/issues/226).
https://github.com/flairNLP/flair/blob/dfa3fbfbc4d8ec237012a800415d1098abdb1aa4/flair/data.py#L406
**To Reproduce**
Steps to reproduce the behaviour (e.g. which model did you train? what parameters did you use? etc.).
```python
# Given a list of tokens
# We simulate a faulty tokenizer by using wrong spans. (_execution_ `start_position` is wrong).
# This faulty tokenizer may be used via a `build_faulty_tokenizer wrapper`,
# such as the one for spacy.
list_tokens = [Token("de", start_position=0), Token("pourvoir", start_position=3), 
Token("à", start_position=12), Token("l&apos;", start_position=14), 
Token("execution", start_position=16), Token("de", start_position=32)]
s = Sentence()
for token in list_tokens:
    s.add_token(token)

# We are now trapped in an infinite loop
print(s.to_original_text())
```
**Expected behaviour**
Possibly an error/warning telling me that my tokenizer is faulty. Another solution would be to check if the `start_pos` is actually bigger, and not only different, than the `pos` counter, as below.
This may produce extra spaces not existent in the original phrase though.
(`pytest tests/test_data.py::test_token_indices` still passes)
```python
for t in self.tokens:
            while t.start_pos > pos: 
                str += " "
                pos += 1
``` 
[Here](https://github.com/flairNLP/flair/compare/master...psorianom:original_text_infinite_loop) is a branch with this modif.

**Environment (please complete the following information):**
 - OS Linux
 - Version flair-0.4.5

alanakbik:
Hello @psorianom - thanks for reporting this and sharing the code to reproduce! The solution looks good - would you like to do a PR for this?

psorianom:
Hi, thank you for checking this. Yes, sure, it is done.

mhoyer:
Hi, I found the exact same issue also in the `Span` class: https://github.com/flairNLP/flair/blob/v0.9/flair/data.py#L505

A PR is on the way.

PR #1533, opened 2020-04-21
To avoid an infinite loop we check the value is greater, not only different

psorianom:
This PR fixes the issue #1529 . The problem is that we may get stuck in an infinite loop when the span `start_pos` indices of `Token` are incorrect, within the `to_original_text` method.

The solution is to simply stay in the `while` loop only if `start_pos` is **greater than** the `pos` counter.


alanakbik:
@psorianom thanks for fixing this!

Issue #1535, opened 2020-04-21
WordEmbeddingsStore: fix error when flair.cache_root is not the default .flair directory


oterrier:
Robustify the code of WordEmbeddingsStore, manage error cases:
- when location of cache_root is not the default one
- fallback to WordEmbedding when the database (sqlite or lmdb) is not found

PR #1536, opened 2020-04-22
GH-1535: Robustify WordEmbeddingsStore

oterrier:
#1535

### DIFF
--- a/flair/inference_utils.py
+++ b/flair/inference_utils.py
@@ -147,7 +147,7 @@ class WordEmbeddingsStoreBackend:
         self.name = embedding.name
         self.store_path : Path = WordEmbeddingsStore._get_store_path(embedding, backend)
         if verbose:
-            print("store filename:", str(self.store_path))
+            log.info("store filename:", str(self.store_path))
     @property
     def is_ok(self):
         return hasattr(self, 'k')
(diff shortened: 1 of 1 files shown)


flair/inference_utils.py
```diff
@@ -128,64 +133,70 @@ def delete_stores(model, backend='sqlite'):
...
 
 class WordEmbeddingsStoreBackend:
     def __init__(self, embedding, backend, verbose=True):
         # get db filename from embedding name
         self.name = embedding.name
-        self.store_filename = WordEmbeddingsStore._get_store_filename(embedding, backend)
+        self.store_path : Path = WordEmbeddingsStore._get_store_path(embedding, backend)
         if verbose:
-            print("store filename:", self.store_filename)
+            log.info("store filename:", str(self.store_path))
```
alanakbik, line 150: This is causing logging errors on my machine. Can you replace this with 
```python
log.info("store filename: " + str(self.store_path))
```

alanakbik:
@oterrier thanks for adding this! Can you fix the logging issue, then I'll merge!

oterrier:
Sure, I'm wondering how is it possible that I didn't see that during my
tests
I have an additional modification to enable the same mechanism for
TextClassifier

I will do that on monday

Have a nice we

Olivier
Kairntech

Le ven. 24 avr. 2020 à 18:10, Alan Akbik <notifications@github.com> a
écrit :

alanakbik:
@oterrier thanks for adding this!

Issue #647, opened 2019-04-08
Use of "Alternative Weighting Schemes for ELMo Embeddings"


stefan-it:
Hi,

this recent NAACL paper (@nreimers) proposes new weighting schemes for ELMo embeddings:

> ELMo embeddings (Peters et. al, 2018) had a huge impact on the NLP community and may recent publications use these embeddings to boost the performance for downstream NLP tasks. However, integration of ELMo embeddings in existent NLP architectures is not straightforward. In contrast to traditional word embeddings, like GloVe or word2vec embeddings, the bi-directional language model of ELMo produces three 1024 dimensional vectors per token in a sentence. Peters et al. proposed to learn a task-specific weighting of these three vectors for downstream tasks. However, this proposed weighting scheme is not feasible for certain tasks, and, as we will show, it does not necessarily yield optimal performance. We evaluate different methods that combine the three vectors from the language model in order to achieve the best possible performance in downstream NLP tasks. We notice that the third layer of the published language model often decreases the performance. By learning a weighted average of only the first two layers, we are able to improve the performance for many datasets. Due to the reduced complexity of the language model, we have a training speed-up of 19-44% for the downstream task.

Find the paper "Alternative Weighting Schemes for ELMo Embeddings" paper [here](https://arxiv.org/abs/1904.02954).

In our current `ELMoEmbeddings` implementation we use a `torch.cat` of all three layers. I think we could also implement an alternative weighting scheme :)

alanakbik:
Thanks for sharing this paper - interesting results. Table 1 for instance shows that concatenation of all three layers is worse on some tasks than just using one of the layers. This seems counterintuitive since if you concatenate everything, the downstream task NN can choose from all information what it needs instead of being limited to one layer. Any idea why this happens? Is this an overfitting issue?

nreimers:
@alanakbik Note that the differences are in general rather small, for example, for SNLI 88.48 vs. 88.50. Likely that these differences are due to randomness.

Adding more features and letting the classifier doing the work of feature selection does not work in all cases, as it can lead to a more heavily overfitting on the training data. Reducing dimension is in many cases a useful method to make the system more stable and leading to better test performances.

alanakbik:
@nreimers interesting, thanks! Since our embeddings are generally very large perhaps reducing dimensionality would also work for us - definitely something to look into.

nreimers:
[...]

Issue #886, opened 2019-07-16
The dimension choose of Elmo using flair


nininininini:
Hi, when I choose Elmo as word embedding, it seems that I can only choose a fixed dimension of 3072. Can I choose other dimensions for Elmo in flair? What part of this 3072 is composed of?

alanakbik:
Hi @nininininini ELMo exposes three layers a 1024 states and by default we take all three to make embeddings. It is currently not possible to select only a subset of these layers, but this is a feature that could be added in a future PR.

nininininini:
I get it, Thanks for your complete answer @alanakbik

gabrielcustodio:
I think that you can to use average layers.
Its possible to use flair with 1024 dims with garanted performance?

alanakbik:
@gabrielcustodio there is a PR #1547 by @falcaopetri that adresses this issue. You can now choose whether to use a concatenation of all three layers (default) or alternatively only the topmost layer or an average. The latter two options only have 1024 dimensions!

```python
# example sentence
sentence = Sentence("The grass is green")

# default option: concatenate all three layers (size: 3072)
embeddings = ELMoEmbeddings(embedding_mode='all')
embeddings.embed(sentence)
print(sentence[0].embedding.size())

# use only the topmost layer  (size: 1024)
embeddings = ELMoEmbeddings(embedding_mode='top')
sentence.clear_embeddings()
embeddings.embed(sentence)
print(sentence[0].embedding.size())

# use an average of all layers  (size: 1024)
embeddings = ELMoEmbeddings(embedding_mode='average')
sentence.clear_embeddings()
embeddings.embed(sentence)
print(sentence[0].embedding.size())
```

After the PR is merged, this feature is available in master branch and will be part of the next Flair release.

PR #1547, opened 2020-04-26
GH-647: add different ways to combine ELMo layers

falcaopetri:
Addresses #647 by adding different ways to combine ELMo layers. Besides the default/current behavior of concatenating the 3 ELMo layers, the user can also get only the topmost layer, or an average of the 3 layers. These combining strategies are the same provided by [AllenNLP's ELMo CLI](https://github.com/allenai/allennlp/blob/d709e59ff3f54e137ea473d411056f68c197c266/allennlp/commands/elmo.py#L31).


alanakbik:
@falcaopetri thanks for adding this! I think a lot of people were waiting for this feature :) see (#886)

PR #1560, opened 2020-04-29
Fix minor accuracy bug

djstrong:
if TP=FP=FN=0 and TN>0


alanakbik:
:+1:

alanakbik:
Thanks for fixing this!

PR #1561, opened 2020-04-29
Add stalebot to Flair

alanakbik:
This PR adds config for stalebot to automatically close issues with long periods of inactivity. In the first phase, issues need to be inactive for 120 days before they get closed. This may be shortened in the future when our release cycles become more frequent.


alanakbik:
:+1:

alanakbik:
@stefan-it can you also give a :+1:? then we can also check if zappr works.

stefan-it:
:+1:

stefan-it:
:heart: 

image

PR #1571, opened 2020-05-03
GH-1362: Add option for hidden state position in FlairEmbeddings

alanakbik:
This PR adds the option to choose which hidden state to use in FlairEmbeddings: either the state at the end of each word, or the state at the whitespace after. Default is the state at the whitespace after. 

You can change the default like this:
```python
embeddings = FlairEmbeddings('news-forward', with_whitespace=False)
```

This configuration seems to be better for syntactic tasks. For POS tagging, it seems that you should set `with_whitespace=False`. For instance, on UD_ENGLISH POS-tagging, we get **96.56 +- 0.03** with whitespace and **96.72 +- 0.04** without, averaged over three runs. 

See the discussion in #1362 for more details.

### DIFF
--- a/flair/embeddings.py
+++ b/flair/embeddings.py
@@ -1836,16 +1836,27 @@ class CharacterEmbeddings(TokenEmbeddings):
 class FlairEmbeddings(TokenEmbeddings):
     """Contextual string embeddings of words, as proposed in Akbik et al., 2018."""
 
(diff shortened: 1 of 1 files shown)


alanakbik:
Special thanks to @djstrong for noticing this and the discussion in #1362!

PR #1581, opened 2020-05-05
GH-1580: True whitespaces in ColumnCorpus

### LINKED ISSUE #1580: Extend ColumnCorpus with option to define "space after"
The Universal Dependencies column format has a way of marking whether or not a whitespace follows a token (our `UniversalDependenciesCorpus` is able to recognize this). However, a similar mechanism does not currently exist for other column-formatted datasets which we read with the `ColumnCorpus` object. So far this was not necessary as all datasets we've encountered, such as CoNLL-03, do not include this information. 

In the context of #1513, we've now come across a use case in which we need to know whether a whitespace follows a token in column format. This issue tracks the extension of `ColumnCorpus` to allow us to include this information.

alanakbik:
Closes #1580 

The `ColumnCorpus` currently assumes that there is always a whitespace between two tokens. This PR adds the option of specifying a special "space-after" column. If that column has a "-" it means that no whitespace follows this token. 

For instance, the following sentence in column format: 

```console
It O +
is O +
a O +
French B-LOC -
- O -
speaking O +
town O -
. O +
```

First column is the token, second the NER tag and third the whitespace after information. After the PR, you can read the corpus like this: 

```python
# read the corpus and specify which column holds the "space after" information
corpus = ColumnCorpus('path/to/corpus', column_format={0: 'text', 1: 'ner', 2: 'space-after'})

# go through sentences 
for sentence in corpus.get_all_sentences():

    # print tokenized version
    print(sentence.to_tokenized_string())

    # print plain version (true whitespacing)
    print(sentence.to_plain_string())
```

If the corpus only consists of the one sentence formatted like above, it will print: 

```console
It is a French - speaking town .
It is a French-speaking town.
```

So the true whitespacing information is now preserved.

### DIFF
--- a/flair/datasets/sequence_labeling.py
+++ b/flair/datasets/sequence_labeling.py
@@ -20,12 +20,10 @@ class ColumnCorpus(Corpus):
             test_file=None,
             dev_file=None,
             tag_to_bioes=None,
-            column_delimiter: str = "\s+",
             comment_symbol: str = None,
+            in_memory: bool = True,
             encoding: str = "utf-8",
             document_separator_token: str = None,
-            skip_first_line: bool = False,
-            in_memory: bool = True,
     ):
         """
         Instantiates a Corpus from CoNLL column-formatted task data such as CoNLL03 or CoNLL2000.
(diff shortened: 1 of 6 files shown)


stefan-it:
@alanakbik Is this already supported for Universal Dependencies 🤔

CoNLL-U format has this `SpaceAfter=No` information (>= UD 2.0):

https://universaldependencies.org/format.html#untokenized-text

alanakbik:
Yes that is already supported through the `UniversalDependenciesDataset` object: 

```python
corpus = UD_ENGLISH()

print(corpus.test[1].to_tokenized_string())
print(corpus.test[1].to_plain_string())
```

prints: 

```console
What if Google expanded on its search - engine ( and now e-mail ) wares into a full - fledged operating system ?
What if Google expanded on its search-engine (and now e-mail) wares into a full-fledged operating system?
```

flair/datasets/sequence_labeling.py
```diff
@@ -20,12 +20,10 @@ def __init__(
             test_file=None,
             dev_file=None,
             tag_to_bioes=None,
-            column_delimiter: str = "\s+",
```
leonweber, line 23: This seems to revert the addition of column_delimiters and the option to skip the first line. Is this intended?

alanakbik: Ah no absolutely not, thanks for catching this.

alanakbik:
Will close this one and do PR from different branch.

Issue #1422, opened 2020-02-08
Using model trained with camembert gives error


CatarinaPC:
**Describe the bug**
When using a model trained with camembert embeddings to predict sentences, I get the following error: `UnboundLocalError: local variable 'tokenizer' referenced before assignment`

The error occurs during SequenceTagger.load().

**To Reproduce**
I trained  model with all default parameters and camembert embeddings. Then I try to give it sentences to predict, as in the following code excerpt:

```
from flair.models import SequenceTagger

# load model
model = SequenceTagger.load("../../models/trained/flair/camembert/best-model.pt")

# by default whitespace based tokenizer is used
model.predict("testing sentence 1")
```

**Error**

```
2020-02-08 18:11:54,184 loading file ../../models/trained/flair/camembert/best-model.pt
Traceback (most recent call last):
  File "/home/catarinapc/cl-extraction-models_new/src/models/testing.py", line 4, in <module>
    model = SequenceTagger.load("../../models/trained/flair/camembert/best-model.pt")
  File "/home/catarinapc/.local/lib/python3.6/site-packages/flair/nn.py", line 86, in load
    state = torch.load(f, map_location=flair.device)
  File "/home/catarinapc/.local/lib/python3.6/site-packages/torch/serialization.py", line 426, in load
    return _load(f, map_location, pickle_module, **pickle_load_args)
  File "/home/catarinapc/.local/lib/python3.6/site-packages/torch/serialization.py", line 613, in _load
    result = unpickler.load()
  File "/home/catarinapc/.local/lib/python3.6/site-packages/flair/embeddings.py", line 1621, in __setstate__
    "-".join(self.name.split("-")[1:])
  File "/home/catarinapc/.local/lib/python3.6/site-packages/transformers/tokenization_utils.py", line 302, in from_pretrained
    return cls._from_pretrained(*inputs, **kwargs)
  File "/home/catarinapc/.local/lib/python3.6/site-packages/transformers/tokenization_utils.py", line 444, in _from_pretrained
    tokenizer.init_inputs = init_inputs
UnboundLocalError: local variable 'tokenizer' referenced before assignment
```

**Environment:**
OS: Ubuntu 18.04.3 LTS
Version: 0.4.4 (I think I'm using the version on the master branch)
Python version: 3.6.8

Thanks for the help!

### UPDATE

Just uninstalled and installed both flair (was using master branch), transformers library, and torch (one version bellow current) and now I get the following error:

```
Traceback (most recent call last):
[...]

PR #1585, opened 2020-05-06
Fix a bug in Camembert embeddings

Alea4jacta6est:


alanakbik:
Thanks! Does this error only occur if you use Camembert embeddings in a StackedEmbedding?

Alea4jacta6est:
No, I didn't use StackedEmbeddings,  my code: 
embeddings = CamembertEmbeddings()
tagger = SequenceTagger(
    hidden_size=256,
    embeddings=embeddings,
    tag_type=tag_type,
    use_crf=True,
    use_rnn=True,
    beta=0.5
)

alanakbik:
Ah I see! Could you modify the IF part such that it fires with any number in the beginning? So for '1-camembert', '2-camembert' etc?

Alea4jacta6est:
Yes, sure
This bug is also mentioned here https://github.com/flairNLP/flair/issues/1422

alanakbik:
Great, thanks for fixing this!

PR #1589, opened 2020-05-07
datasets: add support for BIOfid dataset

stefan-it:
Hi,

this PR adds support for the BIOfid dataset:

> The Specialized Information Service Biodiversity Research (BIOfid) has been launched to mobilize valuable biological data from printed literature hidden in German libraries for over the past 250 years. In this project, we annotate German texts converted by OCR from historical scientific literature on the biodiversity of plants, birds, moths and butterflies. Our work enables the automatic extraction of biological information previously buried in the mass of papers and volumes. For this purpose, we generated training data for the tasks of Named Entity Recognition (NER) and Taxa Recognition (TR) in biological documents. We use this data to train a number of leading machine learning tools and create a gold standard for TR in biodiversity literature. More specifically, we perform a practical analysis of our newly generated BIOfid dataset through various downstream-task evaluations and establish a new state of the art for TR with 80.23% F-score. In this sense, our paper lays the foundations for future work in the field of information extraction in biology texts.

More information can be found in the paper ["BIOfid Dataset: Publishing a German Gold Standard for Named Entity Recognition in Historical Biodiversity Literature"](https://www.aclweb.org/anthology/K19-1081/) from Ahmed et al.

Data is taken from [their repository](https://github.com/texttechnologylab/BIOfid/tree/master/BIOfid-Dataset-NER).

## Usage

Here's an example of how to use the dataset in Flair:

```python
import flair.datasets

biofid = flair.datasets.BIOFID()

print(biofid)
```

Some sanity checks:

```python
tag_type = "ner"
tag_dictionary = biofid.make_tag_dictionary(tag_type=tag_type)

labels = set([item[2:] for item in tag_dictionary.get_items() if "-" in item])

assert len(labels) == 6 # As mentioned in the paper
```

However, the number of sentences differ a bit:

```python
len_train = len(biofid.train)
len_dev = len(biofid.dev)
len_test = len(biofid.test)

print(len_train + len_dev + len_test)
```

Outputs: 15,836, whereas the paper reports a number of 15,833.

### DIFF
--- a/flair/datasets/__init__.py
+++ b/flair/datasets/__init__.py
@@ -24,6 +24,7 @@ from .sequence_labeling import WIKINER_PORTUGUESE
 from .sequence_labeling import WIKINER_POLISH
 from .sequence_labeling import WIKINER_RUSSIAN
 from .sequence_labeling import WNUT_17
+from .sequence_labeling import BIOFID
 
 # Expose all document classification datasets
 from .document_classification import ClassificationCorpus
--- a/flair/datasets/sequence_labeling.py
+++ b/flair/datasets/sequence_labeling.py
@@ -966,6 +966,38 @@ class WNUT_17(ColumnCorpus):
         )
 
 
+class BIOFID(ColumnCorpus):
+    def __init__(
+            self,
+            base_path: Union[str, Path] = None,
+            tag_to_bioes: str = "ner",
+            in_memory: bool = True,
+    ):
+        if type(base_path) == str:
+            base_path: Path = Path(base_path)
+
+        # column format
+        columns = {0: "text", 1: "lemma", 2: "pos", 3: "ner"}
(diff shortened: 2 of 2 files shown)


alanakbik:
Cool dataset, thanks for integrating this! Even some tags like TIME and TAXON that the other German NER datasets don't have. I hope we can get multi-task learning into Flair at some point so that we could train NER models using all German datasets for instance.

alanakbik:
@stefan-it the line `from .sequence_labeling import BIOFID` should be added to `datasets/__init__.py` to make the example code run.

stefan-it:
@alanakbik Oh, forgot to push that import 😅 thanks for that hint, should be added now :)

alanakbik:
:+1:

PR #1616, opened 2020-05-18
missing Embedding attributes; conll03 doc

cod3licious:
add required attributes to base Embedding class; improve documentation for conll03 corpus

### DIFF
--- a/flair/embeddings/base.py
+++ b/flair/embeddings/base.py
@@ -17,9 +17,11 @@ class Embeddings(torch.nn.Module):
 
     def __init__(self):
         """Set some attributes that would otherwise result in errors. Overwrite these in your embedding class."""
-        self.name: str = "unnamed_embedding"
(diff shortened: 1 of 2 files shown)


flair/embeddings/base.py
```diff
@@ -15,6 +15,12 @@
 class Embeddings(torch.nn.Module):
     """Abstract base class for all embeddings. Every new type of embedding must implement these methods."""
 
+    def __init__(self):
+        """Set some attributes that would otherwise result in errors. Overwrite these in your embedding class."""
```
alanakbik, line 19: You need to call the constructor of the super class here with `super().__init__()`. This should fix the unit tests.

cod3licious: and I think existing attributes would have been overwritten. both should be fixed now.

alanakbik:
@cod3licious thanks for fixing this!

Issue #1621, opened 2020-05-22
Keyphrase Extraction Dataset


whoisjones:
There are 3 datasets available for keyphrase extraction via sequence labeling from IIIT-Delhi:
[Keyphrase Datasets](https://github.com/midas-research/keyphrase-extraction-as-sequence-labeling-data)

I would like to add all three sets to the sequence labeling functionality of flair.

- [x] [Inspec](https://github.com/midas-research/keyphrase-extraction-as-sequence-labeling-data/tree/master/Inspec) 

- [x] [SemEval-2017](https://github.com/midas-research/keyphrase-extraction-as-sequence-labeling-data/tree/master/SemEval-2017) 

- [x] [Processed SemEval-2010](https://github.com/midas-research/keyphrase-extraction-as-sequence-labeling-data/tree/master/processed_semeval-2010) 

<img width="1121" alt="keyphrase-stats" src="https://user-images.githubusercontent.com/38491117/82667075-65023800-9c37-11ea-9584-5a26420f064f.png">

whoisjones:
Stats of corpora:

INSPEC files:
`{
    "TRAIN": {
        "dataset": "TRAIN",
        "total_number_of_documents": 4423,
        "number_of_documents_per_class": {},
        "number_of_tokens_per_tag": {},
        "number_of_tokens": {
            "total": 137086,
            "min": 1,
            "max": 172,
            "avg": 30.993895546009497
        }
    },
    "TEST": {
        "dataset": "TEST",
        "total_number_of_documents": 2080,
        "number_of_documents_per_class": {},
        "number_of_tokens_per_tag": {},
        "number_of_tokens": {
            "total": 65221,
            "min": 1,
            "max": 339,
            "avg": 31.35625
        }
    },
    "DEV": {
        "dataset": "DEV",
        "total_number_of_documents": 2032,
        "number_of_documents_per_class": {},
        "number_of_tokens_per_tag": {},
        "number_of_tokens": {
            "total": 64258,
            "min": 1,
            "max": 206,
            "avg": 31.623031496062993
        }
    }
}`

SEMEVAL 2017:
`{
    "TRAIN": {
        "dataset": "TRAIN",
        "total_number_of_documents": 2305,
        "number_of_documents_per_class": {},
        "number_of_tokens_per_tag": {},
        "number_of_tokens": {
            "total": 63013,
            "min": 1,
            "max": 142,
            "avg": 27.33752711496746
        }
    },
    "TEST": {
[...]

PR #1629, opened 2020-05-24
adding Keyword Extraction corpora

whoisjones:
closes gh-1621: added 3 datasets from [midas-research](https://github.com/midas-research/keyphrase-extraction-as-sequence-labeling-data) which can used for as keyword / keyphrase detection. data is provided as sequence labeling data, so it can be used just like NER, etc. documentation is updated as well.


alanakbik:
:+1:

Issue #1622, opened 2020-05-22
plot_weights figsize referenced before assignment


cod3licious:
`# figsize = (16, 16)` in [line 125 of Plotter.plot_weights](https://github.com/flairNLP/flair/blob/master/flair/visual/training_curves.py#L125) should be uncommented or the if statement below should be changed, otherwise this can throw `UnboundLocalError: local variable 'figsize' referenced before assignment`.

alanakbik:
Oha yes, care to do a PR? ;)

PR #1637, opened 2020-05-26
fix figsize issue; allow for str in addition to Path

cod3licious:
this should fix issue #1622. additionally in a bunch of places it is now possible to give file paths as strings instead of only Path instances. besides that some minor formatting things where flake8 in my text editor was being annoying.

### DIFF
--- a/flair/file_utils.py
+++ b/flair/file_utils.py
@@ -2,7 +2,7 @@
 Utilities for working with the local dataset cache. Copied from AllenNLP
 """
 from pathlib import Path
-from typing import Tuple, Optional, Sequence, cast
+from typing import Tuple, Union, Optional, Sequence, cast
 import os
 import base64
 import logging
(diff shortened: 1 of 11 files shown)


flair/file_utils.py
```diff
@@ -99,15 +101,17 @@ def cached_path(url_or_filename: str, cache_dir: Path) -> Path:
         )
 
 
-def unzip_file(file: Path, unzip_to: Path):
+def unzip_file(file: Union[str, Path], unzip_to: Union[str, Path]):
     from zipfile import ZipFile
 
-    with ZipFile(file, "r") as zipObj:
+    with ZipFile(Path(file), "r") as zipObj:
```
alanakbik, line 107: What if `file is` already a Path?

cod3licious: `Path(Path("lkasjdf/lkasjdf"))` executes without error so it shouldn't matter. Indeed probably all the checks for `str` types before conversion could probably be removed but I checked that too late 😅

alanakbik: oha didn't know that :D

alanakbik:
:+1:

alanakbik:
@cod3licious thanks for fixing this!

PR #1652, opened 2020-05-30
Ner corpus swedish

marcelmmm:
#3 created corpus for swedish NER using dataset https://github.com/klintan/swedish-ner-corpus/ 
a function is added that adds IOB2 tags to the dataset since they are missing.


alanakbik:
:+1:

whoisjones:
👍

PR #1659, opened 2020-06-03
GH-1628: serialize dropout

### LINKED ISSUE #1628: use_dropout not getting saved in checkpoint ???
I think the use_dropout parameter does not saved as part of state_dict due to bug in
 def _get_state_dict(self) 

as a result of which if we set a dropout layer in sequence tagger, and resume checkpoint training, the behavior changes.

alanakbik:
Adds `use_dropout` parameter to serialization and so closes #1628


alanakbik:
:+1:

yosipk:
👍

PR #1684, opened 2020-06-10
Short-circuit the embedding loop

bratao:
Hello, 
This is a simple MR, because during the embed function you keep checking for all tokens to see if someone needs a embedding. This MR just change the loop to stop as soon it find one

### DIFF
--- a/flair/embeddings/base.py
+++ b/flair/embeddings/base.py
@@ -50,10 +50,12 @@ class Embeddings(torch.nn.Module):
                 for token in sentence.tokens:
                     if self.name not in token._embeddings.keys():
                         everything_embedded = False
+                        break
         else:
             for sentence in sentences:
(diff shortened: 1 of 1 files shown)


alanakbik:
Thanks for adding this. Unfortunately, this logic would break if you have a list of sentences of which some are already embedded and others not. In this case, it would see one embedded sentence and assume they all are embedded and so not proceed to embed them.

bratao:
@alanakbik are you sure? Maybe I´m too tried ( In fact I´m now 😢). But from my understanding of the code, it will just stop the look if it find One sentence that is not embedded. In this case, it will re-do all sentences in the if below:
`if not everything_embedded`

alanakbik:
Ah yes you're right, sorry! :D

PR #1686, opened 2020-06-11
#1685 deserialization of tokenizer always gets loaded from huggingface

### LINKED ISSUE #1685: Deserialization of tokenizer gets loaded from huggingface
**Describe the bug**
At the deserialization of a model, the tokenizer gets loaded from huggingface (we're using TransformerWordEmbeddings). This throws an error when for instance using SCIBERT, because it is not supported in huggingface.

**To Reproduce**
Train a sequence tagger model using scibert. Then according to the docs: tagger = SequenceTagger("model_using_scibert"), which throws an errors since it is looking the scibert tokenizer in the huggingface lib.

**Expected behavior**
The tokenizer should not get loaded from huggingface if it is not available.

 - macOS Catalina
 - Version flair 0.5

whoisjones:
try catch block around tokenizer.from_pretrained in order to avoid bug

closes #1685

### DIFF
--- a/flair/embeddings/token.py
+++ b/flair/embeddings/token.py
@@ -1134,7 +1134,10 @@ class TransformerWordEmbeddings(TokenEmbeddings):
 
         # reload tokenizer to get around serialization issues
         model_name = self.name.split('transformer-word-')[-1]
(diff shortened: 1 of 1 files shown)


alanakbik:
Thanks for fixing this!

Issue #1647, opened 2020-05-28
Pretrained Model for Keyphrase Extraction


whoisjones:
Since we've integrated Keyphrase Detection Datasets it might a cool to have pretrained model for this. As a foundation we can either use Datasets from #1621 or #1646.
I would train some models and post the results here to check if it makes sense to integrate it.

whoisjones:
according to midas-research SciBert gives (almost) the best results among all three keyphrase datasets in flair (metric: f1-score):

||INSPEC|SEMEVAL2010|SEMEVAL2017|
|---|---|---|---|
|midas|0.593|0.357|0.521|
|our model||0.351|0.444|

alanakbik:
What are the embeddings used in our model?

whoisjones:
SciBert, here is also the data from midas:
![image](https://user-images.githubusercontent.com/38491117/83887287-87707700-a748-11ea-947f-c67e525acd0c.png)

alanakbik:
Model added to Flair in #1689

djstrong:
@whoisjones How the model was trained? The scibert was finetuned in the process? What head is used (linear, rnn, crf)?

PR #1689, opened 2020-06-11
gh-1647: pretrained keyphrase tagger model

whoisjones:
a model has been pretrained on the semeval2017 data. it can be loaded
via the SequenceTagger class using the param ("keyphrase"). Results from
model are attached to the issue.


alanakbik:
Thanks for adding this!

Issue #1634, opened 2020-05-25
AttributeError: 'ELMoEmbeddings' object has no attribute 'embedding_mode_fn'


yzhaoinuw:
Hi Flair developers, thank you for releasing version 0.5 of flair. I am very excited and wasted no time trying it. I upgraded it today to release 0.5 from release 0.45. Then I went ahead to run an NER module I made afew days ago to see if I need to make changes to my module because of the new flair release. My NER module uses the flair sequence tagger and the ELMo embeddings. But now I encounter the following error message:

```
  File "/home/Desktop/name_parser.py", line 44, in parse_name
    self.model.predict(name,all_tag_prob=True)

  File "/home/anaconda3/envs/nlp/lib/python3.7/site-packages/flair/models/sequence_tagger_model.py", line 372, in predict
    feature: torch.Tensor = self.forward(batch)

  File "/home/anaconda3/envs/nlp/lib/python3.7/site-packages/flair/models/sequence_tagger_model.py", line 513, in forward
    self.embeddings.embed(sentences)

  File "/home/anaconda3/envs/nlp/lib/python3.7/site-packages/flair/embeddings/base.py", line 59, in embed
    self._add_embeddings_internal(sentences)

  File "/home/anaconda3/envs/nlp/lib/python3.7/site-packages/flair/embeddings/token.py", line 1594, in _add_embeddings_internal
    word_embedding = self.embedding_mode_fn(elmo_embedding_layers)

  File "/home/anaconda3/envs/nlp/lib/python3.7/site-packages/torch/nn/modules/module.py", line 585, in __getattr__
    type(self).__name__, name))

AttributeError: 'ELMoEmbeddings' object has no attribute 'embedding_mode_fn'
```

One code in my module loads a NER model I trained, which used the ELMo emneddings. Then, I use the loaded model to predict NER tags. My module was working well with flair 0.45. All that changed was was that I upgraded flair to 0.5. I tried to look in to ELMoEmbeddings class in [token.py](https://github.com/flairNLP/flair/blob/083eb114505ede0ea3070317a5a08d13fc71891b/flair/embeddings/token.py) and found from line 1543 to 1548 where it is defined as follows

```
 if embedding_mode == "all":
        self.embedding_mode_fn = lambda x: torch.cat(x, 0)
elif embedding_mode == "top":
        self.embedding_mode_fn = lambda x: x[-1]
elif embedding_mode == "average":
[...]

Issue #1618, opened 2020-05-20
Using ElmoEmbeddings crashes at the first save attempt. 


nstylia:
Using stacked flair embeddings with elmo (or even standalone elmo I assume), results in an error while saving the model due to the lambda function in the elmo embeddings. As a result, the pickler in pytorch cannot save the model and crashes after the first epoch, while trying to save the model. The same code snippet works without the elmo embeddings (omitted parameters in SequenceTagger and train). 

`embedding_types: List[TokenEmbeddings] = [
    PooledFlairEmbeddings('pubmed-forward'),
    PooledFlairEmbeddings('pubmed-backward'),
    ELMoEmbeddings('pubmed'),
]`
`embeddings: StackedEmbeddings = StackedEmbeddings(embeddings=embedding_types)`
`tagger: SequenceTagger = SequenceTagger(...)`
`trainer: ModelTrainer = ModelTrainer(tagger, corpus)`
`trainer.train(...)` 

The error message: 

> Traceback (most recent call last):
  File "flair_main.py", line 43, in <module>
    max_epochs=100)
  File "/home/ns/envs/mldev/lib/python3.6/site-packages/flair/trainers/trainer.py", line 552, in train
    self.model.save(base_path / "best-model.pt")
  File "/home/ns/envs/mldev/lib/python3.6/site-packages/flair/nn.py", line 70, in save
    torch.save(model_state, str(model_file), pickle_protocol=4)
  File "/home/ns/envs/mldev/lib/python3.6/site-packages/torch/serialization.py", line 328, in save
    _legacy_save(obj, opened_file, pickle_module, pickle_protocol)
  File "/home/ns/envs/mldev/lib/python3.6/site-packages/torch/serialization.py", line 401, in _legacy_save
    pickler.dump(obj)
AttributeError: Can't pickle local object 'ELMoEmbeddings.\_\_init\_\_.\<locals\>.\<lambda\> '

Environment information: 
Ubuntu 16.04
torch-1.4.0+cu101
flair-0.4.5

Is there any work around ? Am I doing something wrong? I don't think this is expected behavior. Also, I assume that pickle would be able to handle lambda functions.

yzhaoinuw:
Same error here with using the ELMo embeddings only. I tried to train a NER model with pretty much the old code (except I wanted to choose 'average' as the embedding mode for ELMo, since this is the new feature in Release 0.5) I used for Release 0.45, which worked well. My code is as follows:

```
from flair.data import Corpus
[...]

PR #1692, opened 2020-06-13
resolve issues caused by embedding mode and lambda functions in ELMoEmbeddings in Release 0.5

yzhaoinuw:
Made this PR to address issues #1618 and #1634 . Replaced the lambda functions in ELMoEmbeddings class (in token.py) with three instance methods. Also added lines inside _add_embeddings_internal to check that if embedding_mode_fn exists, if not, assign it to be the embedding mode function for all layers to prevent "AttributeError: 'ELMoEmbeddings' object has no attribute 'embedding_mode_fn'".


alanakbik:
@yzhaoinuw thanks a lot for fixing this! This will surely help many people!

yzhaoinuw:
@alanakbik You are welcome. Glad to contribute to this awesome library!

PR #1694, opened 2020-06-16
Add option to pass kwargs into transformer models when initializing model

albalaka:


alanakbik:
@albalaka thanks for adding this!

PR #1700, opened 2020-06-18
fix condition to assign whitespace_after attribute in the build_spacy_tokenizer wraper

mauryaland:
The old condition assigned False to whitespace_after when there is actually a whitespace after the token.


alanakbik:

Issue #1636, opened 2020-05-26
Dataset of Communicative Functions of Sentences


alanakbik:
A recent [LREC paper](https://www.researchgate.net/publication/339658767_An_Evaluation_Dataset_for_Identifying_Communicative_Functions_of_Sentences_in_English_Scholarly_Papers) presents a dataset of communicative functions of sentences in scholarly papers. The dataset is available [here](https://github.com/Alab-NII/FECFevalDataset). Perhaps it would be interesting to add support for this dataset to Flair.

PR #1706, opened 2020-06-21
GH-1636 integrated communicative functions model in TextClassifier

marcelmmm:


alanakbik:

Issue #1726, opened 2020-07-01
Transformers: AttributeError: 'BertTokenizer' object has no attribute 'prepare_for_model'


stefan-it:
Hi,

latest version of 🤗 Transformers no longer supports the `prepare_for_model` method, that we're using in our `TransformerWordEmbeddings` class:

https://github.com/flairNLP/flair/blob/8c09e62d9a5a3c227b9ca0fb9f214de9620d4ca0/flair/embeddings/token.py#L957-L960

The following error message is thrown:

```bash
Traceback (most recent call last):
  File "train_conll_en.py", line 58, in <module>
    shuffle=True
  File "/mnt/flair-paper/flair/trainers/trainer.py", line 349, in train
    loss = self.model.forward_loss(batch_step)
  File "/mnt/flair-paper/flair/models/sequence_tagger_model.py", line 599, in forward_loss
    features = self.forward(data_points)
  File "/mnt/flair-paper/flair/models/sequence_tagger_model.py", line 604, in forward
    self.embeddings.embed(sentences)
  File "/mnt/flair-paper/flair/embeddings/token.py", line 71, in embed
    embedding.embed(sentences)
  File "/mnt/flair-paper/flair/embeddings/base.py", line 61, in embed
    self._add_embeddings_internal(sentences)
  File "/mnt/flair-paper/flair/embeddings/token.py", line 892, in _add_embeddings_internal
    self._add_embeddings_to_sentences(batch)
  File "/mnt/flair-paper/flair/embeddings/token.py", line 957, in _add_embeddings_to_sentences
    encoded_inputs = self.tokenizer.prepare_for_model(subtoken_ids_sentence,
AttributeError: 'BertTokenizer' object has no attribute 'prepare_for_model'
```

It should be replaced with `encode_plus` method, see https://github.com/huggingface/transformers/issues/5447.

I'll prepare a PR for that!

alanakbik:
Closed by #1727 - thanks @stefan-it !

PR #1727, opened 2020-07-01
🤗 Transformers: Update to latest 3.0.0 version

stefan-it:
Hi,

this PR updates the awesome 🤗 Transformers to latest 3.0.0 version, see release noted [here](https://github.com/huggingface/transformers/releases/tag/v3.0.0).

In version >= 3.0.0 the `prepare_for_model` method was removed, `encode_plus` should be used instead, see #1726 .

This PR fixes it :) I couldn't see any performance degradation when training one epoch for CoNLL-2003 in comparison to the `prepare_for_model` method.

### DIFF
--- a/flair/embeddings/document.py
+++ b/flair/embeddings/document.py
@@ -6,8 +6,6 @@ import torch
 from torch.nn.utils.rnn import pack_padded_sequence, pad_packed_sequence
 from transformers import BertTokenizer, AlbertTokenizer, AutoTokenizer, AutoConfig, AutoModel
 
-from sentence_transformers import SentenceTransformer
-
 import flair
 from flair.data import Sentence
 from flair.embeddings.base import Embeddings, ScalarMix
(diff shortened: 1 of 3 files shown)


stefan-it:
CI build is currently failing because of:

```
error: transformers 3.0.0 is installed but transformers==2.11.0 is required by {'sentence-transformers'}
```

So the `sentence_transformers` dependency currently requires a fixed and older version of Transformers, but it seems that they will upgrade to latest transformers soon, see this issue: https://github.com/UKPLab/sentence-transformers/issues/289

alanakbik:
Hm, maybe we could remove the sentence-transformers dependency for now and make it optional (like `allennlp`)?

stefan-it:
I will remove the dependency in `requirements file and add a check in the `SentenceTransformerDocumentEmbeddings` implementation now!

alanakbik:
@stefan-it thanks for fixing this! Also can a quick conll-03 experiment and am getting the same numbers as before.

nreimers:
Hi @stefan-it 
I updated sentence-transformers to version 0.3.0, which will now work with transformers >=3.0.2.

The missing prepare_for_model  method was added back to huggingface in version 3.0.1.

Issue #1751, opened 2020-07-09
Long sentences while training SequenceTagger with TransformerWordEmbeddings


djstrong:
Most of transformer models handle texts up to 512 subtokens. If we train SequenceTagger with a longer sentence then we get:
```
Traceback (most recent call last):
  File "train_tagger.py", line 240, in <module>
    anneal_with_restarts=args.anneal_with_restarts
  File "venv/lib/python3.7/site-packages/flair/trainers/trainer.py", line 349, in train
    loss = self.model.forward_loss(batch_step)
  File "venv/lib/python3.7/site-packages/flair/models/sequence_tagger_model.py", line 599, in forward_loss
    features = self.forward(data_points)
  File "venv/lib/python3.7/site-packages/flair/models/sequence_tagger_model.py", line 604, in forward
    self.embeddings.embed(sentences)
  File "venv/lib/python3.7/site-packages/flair/embeddings/token.py", line 71, in embed
    embedding.embed(sentences)
  File "venv/lib/python3.7/site-packages/flair/embeddings/base.py", line 61, in embed
    self._add_embeddings_internal(sentences)
  File "venv/lib/python3.7/site-packages/flair/embeddings/token.py", line 897, in _add_embeddings_internal
    self._add_embeddings_to_sentences(batch)
  File "venv/lib/python3.7/site-packages/flair/embeddings/token.py", line 1053, in _add_embeddings_to_sentences
    final_embedding: torch.FloatTensor = current_embeddings[0]
IndexError: index 0 is out of bounds for dimension 0 with size 0
```
I am using flair master.

Is there any automatic way of truncating such sentences?

djstrong:
Now, I see there is a code which divides sentences, but probably is not working properly.

alanakbik:
If you set `allow_long_sentences=True` in the embedding class it should be able to do this.

```python
TransformerWordEmbeddings('distilbert-base-uncased', allow_long_sentences=True)
```

djstrong:
Thank you, I missed that.

UrszulaCzerwinska:
Hi, I am getting an error with camembert-base

```
embedding=TransformerWordEmbeddings(model="camembert-base",
    layers= -1,
    pooling_operation= "mean",
    fine_tune=False,
    allow_long_sentences=True)
```

`
TypeError: __init__() got an unexpected keyword argument 'allow_long_sentences'
`

PR #1752, opened 2020-07-09
allow_long_sentences documentation #1751

djstrong:


alanakbik:
@djstrong thanks for adding this!

Issue #1445, opened 2020-02-24
Targeted Sentiment Analysis? (or Aspect-Based Sentiment Analysis)


snosrap:
Flair is great for assessing the sentiment of a sentence or document.

Is there a recommended approach for performing **targeted sentiment analysis** (that is, assessing the sentiment associated with a _particular entity_ in a sentence)?

* This would be useful for online reviews, where different sentiments are expressed about different attributes of the product/service ("the atmosphere was noisy, but the food was great"). See the [SemEval-2014 Task 4 dataset](http://alt.qcri.org/semeval2014/task4/).

* This would also be useful for clinical text, where doctors frequently express different sentiments about medical conditions a patient may or may not have ("Pt denies dyspnea and chest pain, but does endorse palpitations"). The [n2b2 2010 Relations dataset](https://portal.dbmi.hms.harvard.edu/projects/n2c2-nlp/) focuses on this use case. [BioScope](https://rgai.inf.u-szeged.hu/node/105) is another dataset that focuses on the sentiment associated with  biomedical entities.

Would appreciate your thoughts.

alanakbik:
@whoisjones can you check out the BioScope dataset?

PR #1758, opened 2020-07-15
GH-1445: Targeted Sentiment Analysis

whoisjones:
closes gh-1445: we've added a pretrained tagger model for negation and speculation based on the bioscope data


alanakbik:
:+1:

alanakbik:

nipunsadvilkar:
@whoisjones : Interesting PR! Was curious to know why you opted for a sequence tagger approach rather than just text classification approach to do aspect-based sentiment analysis?

It will help me understand how you have framed the problem and understand the pros & cons of it. Thanks!

whoisjones:
@nipunsadvilkar a sequence tagger gives a way more information compared to a text classification. Classification by definition will aggregate information. consider you are having a nested negation and / or speculation in one document. instead of assigning it to a class, we can say that part of the sentence / document is negated oder speculative.

nipunsadvilkar:
@whoisjones Thanks for prompt reply! Appreciate it 👍 
so let's say we have a sentence :
sent
```
The patient states of no fever or chills but has asthma and a sore throat that has been going on for 3 months.
```

Entities/Aspects would be Disease mentions - `fever`, `chills`, `asthma`, `sore throat`

Negated - `fever`, `chills`
Assertive - `asthma`, `sore throat`

**Approach A)** In the text classifier approach, plausible setting would be to label the above sentence with 4 different mentions and train aspect-based text classifier.


Text | Entity | Label
-- | -- | --
sent | fever | Negative
sent | chills | Negative
sent | asthma | Assertive
sent | sore throat | Assertive

**Approach B)** Sequence Tagger 
How would you frame it into a sequence tagger problem? What would be scope for those 4 entities? Does scope also need to be annotated?

from what I see in your repo - [whoisjones/BioScopeSequenceLabelingData](https://github.com/whoisjones/BioScopeSequenceLabelingData). You would tag it like following?

<details>
  <summary>Click to see IOB tagging scheme!</summary>
The     O</br>
patient O</br>
states  O</br>
of      O</br>
no      B-NEGATION</br>
fever   I-NEGATION</br>
or      I-NEGATION</br>
chills  I-NEGATION</br>
but     O</br>
has     O</br>
asthma  O</br>
and     O</br>
a       O</br>
sore    O</br>
throat  O</br>
that    O</br>
has     O</br>
been    O</br>
going   O</br>
on      O</br>
for     O</br>
3       O</br>
months  O</br>
.       O</br>
</details>

I Agree this scope of negation or speculation becomes tricky to handle. In above example negation for 2 entities is consecutive. How do you handle disjoint spans?

Also, would like to hear your thoughts on **Approach A)**

whoisjones:
@nipunsadvilkar sorry for the late reply. here's what I think:
**Approach A** as you described also needs at least something pretty close to a sequence tagger since you already identified the entities. To my knowledge most of NER downstream task are sequence labeling problems. However if you have identified the entities (in our case diseases) and do it like you descirbed, then you still need information about the dependencies inside the phrase, check out this example here: [Aspect Based Analysis](http://alt.qcri.org/semeval2014/task4/).

`“I hated their fajitas, but their salads were great” --> {fajitas: negative, salads: positive}`

If we would only give labels to this phrase like this, we would learn with a text classifier approach that fajitas in general are bad and salads in general are delicious. However we want to know which item is rated good or bad and this depends on where in the phrase fajitas or salads is positioned. So we would like a sequence labeling problem around it.

**Approach B**
This corpora we used was annotated by linguistic professionals containing speculations and negations in the medical space. We can't identify with our pretrained tagger what kind of disease it is - for that we would need another tagger exactly as you mentioned. But it can be a lot easier - think about this:

Let our tagger annotate your text whether it is speculation or negation. Then search for all the diseases you would like to know about. This obviously only includes all diseases you are specifying beforehand.
To answer your question: If you want a want general sequence tagger with which you can identify all kinds of diseases, you would need a training set in a similar format as the BioScope dataset to train one.

PR #1764, opened 2020-07-17
Adapt BytePairEmbeddings class to new version of bpemb

mauryaland:
Now, we can load custom BPEmb into BytePairEmbeddings class since bpemb package has been upgraded to 0.3.2.


alanakbik:
@mauryaland thanks for adding this and please excuse the delay in responding!

PR #1815, opened 2020-08-17
Conditional flatten_parameters for quantization

kabirkhan:
This is only change required to get dynamic PyTorch quantization working. This is because the DynamicQuantizedLSTM that Pytorch replaces the LSTM layer doesn't implement the flatten_parameters method in it's RnnBase class. 

So now, standard dynamic quantization will work. Quantizing both the LSTM and Linear layers results in about 4X decrease in model size and 3-4X increase in inference speed on CPU. (I don't have any concrete benchmark on inference speed as I've just been testing on my laptop but the 3-4X is consistent with more official benchmarks from Pytorch and others)

e.g.
```python
import os

import torch


tagger = SequenceTagger.load('ner')
tagger.cpu() # Dynamic Quantization currently only works on CPU

quantized_tagger = torch.quantization.quantize_dynamic(
    tagger, {torch.nn.LSTM, torch.nn.Linear}, dtype=torch.qint8
)

def print_size_of_model(model):
    torch.save(model.state_dict(), "temp.p")
    print('Size (MB):', os.path.getsize("temp.p")/1e6)
    os.remove('temp.p')

print("TAGGER: ner")
print_size_of_model(tagger)
print_size_of_model(quantized_tagger)

# Outputs:
# TAGGER: ner
# Size (MB): 253.036663
# Size (MB): 63.571843
```

### DIFF
--- a/flair/models/language_model.py
+++ b/flair/models/language_model.py
@@ -75,7 +75,8 @@ class LanguageModel(nn.Module):
         encoded = self.encoder(input)
         emb = self.drop(encoded)
 
-        self.rnn.flatten_parameters()
+        if hasttr(self.rnn, "flatten_parameters"):
+            self.rnn.flatten_parameters()
 
         output, hidden = self.rnn(emb, hidden)
 
--- a/flair/models/language_model.py
+++ b/flair/models/language_model.py
@@ -75,7 +75,7 @@ class LanguageModel(nn.Module):
         encoded = self.encoder(input)
         emb = self.drop(encoded)
 
-        if hasttr(self.rnn, "flatten_parameters"):
+        if hasattr(self.rnn, "flatten_parameters"):
             self.rnn.flatten_parameters()
 
         output, hidden = self.rnn(emb, hidden)


alanakbik:
Interesting, thank for putting in this PR! In my setup, the above code only runs if I specify `flair.device = 'cpu' `, i.e. if I make sure flair is not using the GPU. Does quantization only work on CPU currently?

kabirkhan:
Yep it only works on CPU at the moment although I think PyTorch has GPU dynamic quantization on the roadmap.

kabirkhan:
The first Note here mentions this for reference. https://pytorch.org/docs/stable/quantization.html

alanakbik:
Ah thanks for the pointer!

helpmefindaname:
Thanks for pointing out this PR, we tried it with our (private dataset).
For NER it dropped from 59.96 F1 to 59.58 and for classification it dropped from 64.08 F1 to 63.85.
Hence we experience almost no prediction-performance loss at all.

However it is still hard to use it, as saving is basically not supported.
I think it would require to rework the way embeddings are handled/saved

alanakbik:
Hello @kabirkhan and @helpmefindaname - first of all sorry for not moving forward on this PR for so long. We tried it a bit locally and also found very limited performance drops and good increases in speed. @marcelmmm is currently benchmarking some models so we get better numbers. 

I think we can add support for this as 'cpu-only' models, though we have to think a bit about usability. @helpmefindaname what specific usability issues did you encounter?

helpmefindaname:
@alanakbik the problem is that quantized models cannot simply be pickled. For the main model that isn't a problem, as it uses the state dict anyways https://github.com/flairNLP/flair/blob/master/flair/models/sequence_tagger_model.py#L256 however the embeddings themselves are also torch models and those are provided as parameters, so they fail.
I suppose to have proper save functionality, it requires to have a similar state_dict functionality for the embeddings too.

kabirkhan:
@helpmefindaname I'm able to dynamically quantize the Linear and LSTM layers in the embeddings as well, just not using the implemented `SequenceTagger.load/.save` methods. 

Saving and loading to/from disk works just fine for me following PyTorch guidelines saving and loading state_dict. Here's a code snippet below of wrapping a Sequence tagger in a class that implements from_disk/to_disk interface:

> Note: This is incomplete as it is an internal system. 

```python
class FlairEntityRecognizer(BaseEntityRecognizer):
    ...
    @classmethod
    def from_disk(cls, model_path: Union[Path, str]):
        """Load pre-trained model

        Parameters
        ----------
        model_path : str
            Path to a directory containing best-model.pt, the trained model

        Returns
        -------
        FlairEntityRecognizer
            The newly constructed object
        """
        model_path = Path(model_path)
        metadata = Metadata.parse_file(model_path / "metadata.json")
        recognizer = cls(name=metadata.name, metadata=metadata)
        weights_path = model_path / "best-model.pt"
        logger.info("Loading model weights from %s", weights_path)

        if metadata.quantized:
            state = torch.load(weights_path)
            tag_dictionary = Dictionary.load_from_file(model_path / "tag_dictionary.pkl")
            model: SequenceTagger = cls._sequence_tagger(
                tag_dictionary=tag_dictionary, fast=metadata.fast
            )
            quantized_model = torch.quantization.quantize_dynamic(
                model, {torch.nn.LSTM, torch.nn.Linear}, dtype=torch.qint8
            )
            quantized_model.load_state_dict(state)
            recognizer._model = quantized_model
        else:
            recognizer._model = SequenceTagger.load(weights_path)

        return recognizer

    @classmethod
    def _sequence_tagger(
        cls,
        tagger: Optional[SequenceTagger] = None,
        tag_dictionary: Dictionary = None,
        fast: Optional[bool] = False,
        hidden_size: Optional[int] = 256,
    ) -> SequenceTagger:
        """Build Flair SequenceTagger

        Parameters
        ----------
        tagger : Optional[SequenceTagger], optional
            Pass in an already initialized SequenceTagger for more control, by default None
        tag_dictionary : Dictionary, optional
            Tag Dictionary for NER task
        fast : bool, optional
            If True, use fast Flair embeddings for fast train/inference and smaller model size
        hidden_size : int, optional
            Hidden size of last linear classifier layer

        Returns
        -------
        SequenceTagger
            Initialized Flair SequenceTagger model
        """

        if not tagger:
            tag_type = "ner"
            embeddings: StackedEmbeddings = StackedEmbeddings(
                embeddings=[
                    WordEmbeddings("glove"),
                    FlairEmbeddings("news-forward-fast" if fast else "news-forward"),
                    FlairEmbeddings("news-backward-fast" if fast else "news-backward"),
                ]
            )

            tagger = SequenceTagger(
                hidden_size=hidden_size,
                embeddings=embeddings,
                tag_type=tag_type,
                use_crf=True,
            )

        return tagger

    def to_disk(self, model_path: Union[Path, str]):
        """Override to_disk to ensure quantized model saves correctly

        Parameters
        ----------
        model_path : Union[Path, str]
            Directory to save model to
        """
        if self._model is None:
            raise ValueError("Recognizer is not yet associated with a trained model.")
        model_path = Path(model_path)
        super().to_disk(model_path) # Superclass saves a metadata object with `quantized` and `fast` properties
        if self.metadata.quantized:
            torch.save(self._model.state_dict(), model_path / "best-model.pt")
            self._model.tag_dictionary.save(model_path / "tag_dictionary.pkl")
        else:
            self._model.save(model_path / "best-model.pt")
```

marcelmmm:
I tried running

quantized_tagger = torch.quantization.quantize_dynamic(
    tagger, {torch.nn.LSTM, torch.nn.Linear}, dtype=torch.qint8
)

on windows and got the error: RuntimeError: Didn't find engine for operation quantized::linear_prepack NoQEngine
The error is due to the fact that I do not have pytorch/FBGEMM on my system and I have a hard time  to install it on windows.

On linux the code works when 'fbgemm' is enabled in torch.backends.quantized.supported_engines and flair.device = 'cpu'.  But then I often do not get the expected results with respect to the running time. The results vary a lot and often the quantized models have the same running time as the normal models or are even slower. E.g. this is the prediction time of 50 sentences.

Model: ner                                   size (MB): 253.036903  Running time: 15.671910285949707
Model: quantized-ner                   size (MB): 19.111023    Running time: 15.782427787780762
Model: ner-fast                            size (MB): 77.593215    Running time: 5.081735849380493
Model: quantized-ner-fast           size (MB): 5.424519       Running time: 6.742673635482788

Does anybody have an idea what the problem could be?

helpmefindaname:
Hello @marcelmmm I'm sorry for replying so late.

I did repeaditive tests with 10k short sentences using the following code:

```
import os
import cProfile
import random
from pathlib import Path
import pstats
from pstats import SortKey
from typing import Union

import flair.models.language_model
import torch

class QuantizableLanguageModel(flair.models.LanguageModel):

    def forward(self, input, hidden, ordered_sequence_lengths=None):
        encoded = self.encoder(input)
        emb = self.drop(encoded)

        if hasattr(self.rnn, "flatten_parameters"):
            self.rnn.flatten_parameters()

        output, hidden = self.rnn(emb, hidden)
        if self.proj is not None:
            output = self.proj(output)
        output = self.drop(output)
        decoded = self.decoder(
            output.view(output.size(0) * output.size(1), output.size(2))
        )
        return (
            decoded.view(output.size(0), output.size(1), decoded.size(1)),
            output,
            hidden,
        )

    @classmethod
    def load_language_model(cls, model_file: Union[Path, str]):

        state = torch.load(str(model_file), map_location=flair.device)

        document_delimiter = state[
            "document_delimiter"] if "document_delimiter" in state else '\n'

        model = cls(
            dictionary=state["dictionary"],
            is_forward_lm=state["is_forward_lm"],
            hidden_size=state["hidden_size"],
            nlayers=state["nlayers"],
            embedding_size=state["embedding_size"],
            nout=state["nout"],
            document_delimiter=document_delimiter,
            dropout=state["dropout"],
        )
        model.load_state_dict(state["state_dict"])
        model.eval()
        model.to(flair.device)

        return model


flair.models.LanguageModel = QuantizableLanguageModel
flair.models.language_model.LanguageModel = QuantizableLanguageModel


nouns = ("puppy", "car", "rabbit", "girl", "monkey")
verbs = ("runs", "hits", "jumps", "drives", "barfs") 
adv = ("crazily.", "dutifully.", "foolishly.", "merrily.", "occasionally.")
adj = ("adorable", "clueless", "dirty", "odd", "stupid")
l=[nouns, verbs, adj, adv]

sentences = [
    ' '.join([random.choice(i) for i in l])
    for j in range(10000)
]

model = SequenceTagger.load("ner")

def benchmark_model(m, sents, name):
    print(name)
    pr = cProfile.Profile()
    pr.enable()
    data = [Sentence(s) for s in sents]
    m.predict(data)
    pr.disable()
    pstats.Stats(pr).sort_stats(SortKey.CUMULATIVE).print_stats(0.1)
    

benchmark_model(model, sentences, "ner - no-quanitsation")
print("/" * 100)


model = SequenceTagger.load("ner")
model = torch.quantization.quantize_dynamic(
    model, {torch.nn.LSTM, torch.nn.Linear}, dtype=torch.qint8
)

benchmark_model(model, sentences, "ner - with-quanitsation")
```
(notice that I used the current release + class injection instead of the branch, as model download didn't work anymore)


for me the results were quite consistent with a factor of ~2,2 (ranging 2,18-2,29 in 4 tries). Besides the prediction speed varying over the tries.

Can you show us how you got the inconsistent results?

marcelmmm:
Hi @helpmefindaname . No problem. This is the code I used

```
#This Fct just returns a list of 100 sentences
def get_sentence_list():
    splitter = SegtokSentenceSplitter()
    sentences=[]
    file = open('/vol/fob-vol7/nebenf19/milich/quantization/10k_wiki_sentences.txt', 'r',encoding='utf-8')
    for _ in range(100):
        sentences.append(splitter.split(file.readline())[0])
    file.close()
    return sentences




print(torch.backends.quantized.supported_engines)
flair.device = 'cpu'

model_names = ['pos','quantized-pos', 'pos-fast', 'quantized-pos-fast']
model_running_times = []

for model in ['pos', 'pos-fast']:   
    
        

        print("MODEL: ",model)
        tagger = SequenceTagger.load(model)

        
        sentences = get_sentence_list()
        
        #prediction
        start = time.time()
        tagger.predict(sentences)
        end = time.time()
        elapsed_time = end - start
        model_running_times.append(elapsed_time)
        print(elapsed_time)
        
    
        print("MODEL: Quantized ",model)

        quantized_tagger = torch.quantization.quantize_dynamic(
            tagger, {torch.nn.LSTM, torch.nn.Linear}, dtype=torch.qint8
        )

        
        sentences= get_sentence_list()

        #prediction
        start = time.time()
        quantized_tagger.predict(sentences)
        end = time.time()
        elapsed_time = end - start
        model_running_times.append(elapsed_time)
        print(elapsed_time)
```

kzinmr:
I encountered the same issue when I measured time performance for 50 short sentences. When we pass multiple sentences at once in multicore machine, the magnitude of speed improvement by batch processing in dataloader might overwhelm that of the quantization. If we process sencences one by one or long sentences, quantization might actually improve the processing speed.

PR #1839, opened 2020-09-01
fix upos-fast link

jxmorris12:
In the current master branch, this fails:

```
SequenceTagger.load("upos-fast")
```

I think I figured out the typo and updated the link correctly.


alanakbik:
@jxmorris12 thanks for fixing this!

PR #1857, opened 2020-09-10
Changed the communicative functions model since the old one had probl…

marcelmmm:
New communicative functions model

### DIFF
--- a/flair/models/text_classification_model.py
+++ b/flair/models/text_classification_model.py
@@ -478,7 +478,7 @@ class TextClassifier(flair.nn.Model):
         
         #Communicative Functions Model
         model_map["communicative-functions"] = "/".join(
-            [hu_path, "comfunc", "communicative-functions-v0.5b.pt"]
+            [hu_path, "comfunc", "communicative-functions.pt"]
         )
 
         cache_dir = Path("models")


flair/models/text_classification_model.py
```diff
@@ -478,7 +478,7 @@ def _fetch_model(model_name) -> str:
         
         #Communicative Functions Model
         model_map["communicative-functions"] = "/".join(
-            [hu_path, "comfunc", "communicative-functions-v0.5b.pt"]
+            [hu_path, "comfunc", "communicative-functions.pt"]
```
alanakbik, line 481: This should be the new model, right?

alanakbik (changes requested):

flair/models/text_classification_model.py
marcelmmm: The new model is called "communicative-function.pt" without the end part "-v0.5b". I admit its confusing.

oneSebastian: This change is indeed still required for the new model to be downloaded.

alanakbik (approved):

alanakbik:
@marcelmmm finally merged, sorry for waiting so long!

PR #1864, opened 2020-09-14
Add support for wikigold NER dataset

fabianhenning:


alanakbik:
@fabianhenning thanks for adding this!

PR #1870, opened 2020-09-18
Added UP data for finnish, chinese, italian and portuguese

Dabendorf:
I have added the UP data for Chinese, Finnish, Italian and Portuguese as well
They all download properly and get tagged correctly, I checked in a dictionary that all marked words are verbs

The commit might look like I touched German as well, but that's just due to a weird "interpretation" of the GitHub-diff-tool. 
I rearranged the order of the languages such that they are alphabetically ordered, therefore it claims that German got moved

Spanish and English still missing, coming soon


alanakbik:
Hello @Dabendorf thanks a lot for adding this! 

One small thing is that the corpus statistics of UP_PORTUGUESE look weird in that most sentences are in the dev dataset (and not test). But I see the same statistics in the [Portuguese UP](https://github.com/System-T/UniversalPropositions/tree/master/UP_Portuguese-Bosque) dataset, so it looks like the error is there.

alanakbik:
Update: I think this problem is known to UP and discussed [here](https://github.com/System-T/UniversalPropositions/issues/4).

Dabendorf:
> Hello @Dabendorf thanks a lot for adding this!
> 
> One small thing is that the corpus statistics of UP_PORTUGUESE look weird in that most sentences are in the dev dataset (and not test). But I see the same statistics in the [Portuguese UP](https://github.com/System-T/UniversalPropositions/tree/master/UP_Portuguese-Bosque) dataset, so it looks like the error is there.

Should I maybe exclude the Portuguese one temporarily and push the other three first?

alanakbik:
Yes I think this is a good idea - we may need to disregard the Portuguese UP until the issues are resolved!

Dabendorf:
> Yes I think this is a good idea - we may need to disregard the Portuguese UP until the issues are resolved!

I have now removed Portuguese from it, the rest stayed the same

alanakbik:

PR #1880, opened 2020-09-22
add UD_Chinese_Kyoto datasets from UniversalDependencies/UD_Classical…

wuqi0704:
…_Chinese-Kyoto


alanakbik:
@wuqi0704 thanks for adding this!

PR #1884, opened 2020-09-23
Remove pytest from the requirements.txt

neophnx:
pytest is not an install dependency.
It should be in another requirements file maybe requirements_dev.txt


alanakbik:
@neophnx makes sense - thanks for fixing this!

PR #1885, opened 2020-09-25
add method get_names to satisfy new flair version

timnon:


alanakbik:
@timnon thanks for adding this!

PR #1887, opened 2020-09-27
UP-datasets for English and Spanish added

Dabendorf:
I have added the UP datasets for English and Spanish (Spanish times)
Interestingly, the frame data was in column 10 and not in 9 in the English dataset, so it was a bit different from they other ones

But they should all work


Dabendorf:
Hm, that commit seems to include the other three commits from the other pull request, so I am not quite sure what happens if you accept both

alanakbik:
Oops just accepted the other one. You likely created this branch from the other branch instead of master.

Dabendorf:
> Oops just accepted the other one. You likely created this branch from the other branch instead of master.

Yes, that happened. I have never had 2 open PullRequests in one project before, so I wasn't aware of that. 

I will close that one here and open another one

PR #1888, opened 2020-09-27
UP-datasets for English and Spanish added

Dabendorf:
I have added the UP datasets for English and Spanish (Spanish times)
Interestingly, the frame data was in column 10 and not in 9 in the English dataset, so it was a bit different from they other ones

But they should all work


alanakbik:
@Dabendorf thanks for adding this!

PR #1912, opened 2020-10-17
Added Simple and Complex MIT Movie Corpus with automatic download

pharnisch:
Created two different corpus classes to maintain the original separation


alanakbik:
@pharnisch thanks for adding this!

PR #1914, opened 2020-10-19
added GoEmotions text classification corpus

marcelmmm:
Added GoEmotions corpus for textclassification


alanakbik:

(@kishaloyhalder the GoEmotions dataset added to Flair)

PR #1933, opened 2020-11-02
Add Greek UD Treebank

malamasn:
Integrate Greek language in flair using the Greek Universal Dependency Treebank


alanakbik:
@malamasn thanks for adding this!

alanakbik:
BTW there is also [this Greek NER dataset](https://github.com/nmpartzio/elNER) that looks interesting.

malamasn:
@alanakbik thank you! I'll have a look at it

PR #1942, opened 2020-11-05
Adding WNUT_2020_NER dataset support

aynetdia:
This implementation appropriately prepares and loads the aforementioned dataset. 

I could also add commit the testing function for the pipenv testing, if needed. 

Edit: typo

### DIFF
--- a/flair/datasets/sequence_labeling.py
+++ b/flair/datasets/sequence_labeling.py
@@ -1,6 +1,7 @@
 import logging
 import re
 import os
+import shutil
 from pathlib import Path
 from typing import Union, Dict, List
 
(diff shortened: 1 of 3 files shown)


flair/datasets/sequence_labeling.py
```diff
@@ -448,6 +449,73 @@ def __init__(
...
+
+            sample_file = data_folder / (sample + ".txt")
+            if not sample_file.is_file():
+
+                zip_path = cached_path(
+                    f"{github_url}", Path("datasets") / dataset_name
+                    )
+
+                # unzip the downloaded repo and merge the train, dev and test datasets 
+                unpack_file(zip_path, data_folder, "zip", False) # unzipped folder name: WNUT_2020_NER-master
```
alanakbik, line 497: I believe the method `unpack_file` is missing. Perhaps you forgot to check it in?

alanakbik: Ah no, only the import statement is missing. Can you add `from flair.file_utils import unpack_file`?

alanakbik (changes requested):

alanakbik:
@aynetdia thanks for adding this! There is an import statement missing - can you add it? 

Also, I notice the statistics of the corpus are slightly off. If I print the `WNUT_2020_NER` object, I get: 

Corpus: **8075** train + 2740 dev + 2691 test sentences

But the task website at https://github.com/jeniyat/WNUT_2020_NER lists the following statistics: 

-    train_data: 370 protocols with **8444** sentences
-    dev_data: 122 protocols with 2839 sentences
-    test_data: 123 protocols with 2813 sentences

Could you double-check?

aynetdia:
I was able to fix the corpus length problem, however only for the train_data and test_data. Now they contain 8444 and 2813 sentences respectively. However the dev_data contains 2862 sentences instead of 2839 as specified in the GitHub repo. 

Not sure what the reason for the discrepancy - I went over the whole dev sample manually and couldn't notice any obvious irregularities.

alanakbik (approved):

alanakbik:
@aynetdia thanks for updating the PR! 

Just one more item: There is a folder for the [2020 version of the test data](https://github.com/jeniyat/WNUT_2020_NER/tree/master/data/test_data_2020) (called `test_data_2020`), which contains a lot more sentences. Their [paper](http://noisy-text.github.io/2020/pdf/2020.d200-1.33.pdf) (see Table 1) distinguishes between Text-18 and Test-20 datasets. Test-20 is what was used to evaluate and has 3562 sentences. Can you change it so that `test_data_2020` is used instead of `test_data`?

alanakbik:
@aynetdia thanks for adding this!

PR #1944, opened 2020-11-06
Adding WEIBO_NER dataset support 

87302380:
Added dataset about NER for Chinese Social Media (Weibo).

The original data comes from https://github.com/hltcoe/golden-horse 
The formatted data is stored at https://github.com/87302380/WEIBO_NER


alanakbik:
@87302380 thanks for adding this!

PR #1945, opened 2020-11-07
Added support for the Livvi dataset from UD

hebecked:
This PR concludes my homework assignment.
It is not recommended to accept the PR at this time, since the data source is currently in the dev branch in my own rep (PR to UD pending) and I'm not qualified to determine if my changes have any linguistic implications.

My changes to the UD dataset:
- Swap the training and test dataset since the old test dataset was much larger than the training dataset. 
- Remove part of the (new) training dataset to create a dev dataset. (Ratios might need adjustment.)

In flair/flair/datasets/document_classification.py Line 48 and flair/flair/datasets/treebanks.py  Line 34 it is stated:
":param dev_file: the name of the dev file, if None, dev data is sampled from train."
This only holds true if the dataset is loaded manually. 
When automatically loaded from UD Line 67 in flair/flair/datasets/treebanks.py prevents this. since "assert path_to_conll_file.exists()" is called for all "path_to_conll_file", "None" or "str".

### DIFF
--- a/flair/datasets/__init__.py
+++ b/flair/datasets/__init__.py
@@ -111,6 +111,7 @@ from .treebanks import UD_CHINESE
 from .treebanks import UD_KOREAN
 from .treebanks import UD_BASQUE
 from .treebanks import UD_GREEK
+from .treebanks import UD_LIVVI
 
 # Expose all text-text datasets
 from .text_text import ParallelTextCorpus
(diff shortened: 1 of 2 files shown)


alanakbik:
@hebecked thanks for adding this! As discussed on [here](https://github.com/UniversalDependencies/UD_Livvi-KKPP/issues/2) can you use the original splits? I.e. the large test split and the small training split? 

Since there is no dev split, you can init the object only using a test and training split. The object `UD_CHINESE_KYOTO` is an example of a UD treebank initialized only with test and train.

alanakbik (changes requested):
Usage of original splits.

PR #1946, opened 2020-11-07
add TURKU_NER corpus

riherrmann:


alanakbik:
@riherrmann thanks for adding this!

Issue #1931, opened 2020-11-02
Store embeddings configuration as kwargs for hyperparameter opt. 


whoisjones:
**Is your feature/enhancement request related to a problem? Please describe.**
Related to #1838. In order to reduce memory, we instanciate word embedding in training process but want to retain AutoCompletion and Argument Errors for our embeddings when configuring a hyperparameter search space.

**Describe the solution you'd like**
Upon creation we store all non-default parameters in a dict attached to the parent class of all word embeddings.

PR #1949, opened 2020-11-08
Add instance parameters to word embedding object upon creation

whoisjones:
closes GH-1931


alanakbik:

PR #1953, opened 2020-11-09
Adding support for the UD_Livvi dataset

hebecked:
Added support for UD_Levvi from UD, based on the original splits. (see previous PR)
Merged with the current upstream changes and tested.
The upstream changes now allow for operation without a dev set file, creating a dev set on the fly based on the training set file.


alanakbik:
@hebecked thanks for adding this!

PR #1954, opened 2020-11-09
Added support for UD_BURYAT

MaxDall:
Added support for UD_BURYAT


alanakbik:
@MaxDall thanks for adding this!

PR #1955, opened 2020-11-09
Added support for UD_NORTH_SAMI

dobbersc:
Added support for UD_NORTH_SAMI. This dataset does not define a dev split.


alanakbik:
@dobbersc thanks for adding this!

PR #1957, opened 2020-11-10
added support for UD_MALTESE

phkuep:


alanakbik:
@phkuep thanks for adding this!

PR #1958, opened 2020-11-10
added support for UD_MARATHI

polarlyset:


flair/datasets/treebanks.py
```diff
@@ -1226,6 +1226,34 @@ def __init__(self, base_path: Union[str, Path] = None, in_memory: bool = True):
...
+        # this dataset name
+        dataset_name = self.__class__.__name__.lower()
+
+        # default dataset folder is the cache root
+        if not base_path:
+            base_path = Path(flair.cache_root) / "datasets"
+        data_folder = base_path / dataset_name
+
+        # download data if necessary
+        web_path = "https://raw.githubusercontent.com/UniversalDependencies/UD_English-EWT/master"
```
alanakbik, line 1245: These paths still point to the English UD treebank. They should point to the Marathi treebank.

alanakbik (changes requested):
The object downloads the English UD treebank, but should download the Marathi one.

flair/datasets/treebanks.py
polarlyset: Path and file names amended.

alanakbik (approved):

alanakbik:
@polarlyset thanks for adding this!

PR #1961, opened 2020-11-11
added support for UD_GOTHIC

wjSimon:


alanakbik:
@wjSimon thanks for adding this!

PR #1964, opened 2020-11-12
Support for UD_OLD_FRENCH added

Weyaaron:
I'm not  sure if additional formating is needed, this loads the corpus without additional reformatings of the source files.

### DIFF
--- a/flair/datasets/treebanks.py
+++ b/flair/datasets/treebanks.py
@@ -1277,3 +1277,31 @@ class UD_AFRIKAANS(UniversalDependenciesCorpus):
         )
 
         super(UD_AFRIKAANS, self).__init__(data_folder, in_memory=in_memory)
+
+
+class UD_OLD_FRENCH(UniversalDependenciesCorpus):
+    def __init__(self, base_path: Union[str, Path] = None, in_memory: bool = True):
+
+        if type(base_path) == str:
+            base_path: Path = Path(base_path)
+
+        # this dataset name
+        dataset_name = self.__class__.__name__.lower()
+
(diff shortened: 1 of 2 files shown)


flair/datasets/treebanks.py
```diff
@@ -1303,3 +1306,30 @@ def __init__(self, base_path: Union[str, Path] = None, in_memory: bool = True):
...
+        # this dataset name
+        dataset_name = self.__class__.__name__.lower()
+
+        # default dataset folder is the cache root
+        if not base_path:
+            base_path = Path(flair.cache_root) / "datasets"
+        data_folder = base_path / dataset_name
+
+        # download data if necessary
+        web_path = "https://github.com/UniversalDependencies/UD_Old_French-SRCMF/tree/master"
```
alanakbik, line 1326: This does not download the correct file. Please check the downloaded file if it looks correct.

alanakbik (changes requested):
This code does not run. If I do: 

```python
# load corpus (throws error)
corpus = UD_OLD_FRENCH()

# print corpus
print(corpus)

# print some sentences in corpus
print(corpus.train[1])
print(corpus.dev[1])
print(corpus.test[1])

# print UPOS tags
dictionary = corpus.make_tag_dictionary('upos')
print(dictionary)
```

It throws an error. Please check if the code runs locally before doing a PR!

flair/datasets/treebanks.py
Weyaaron: The urls are fixed with the new commits added.

alanakbik:
@Weyaaron thanks for adding this!

PR #1967, opened 2020-11-12
Added support for UD_WOLOF

LukasOpp:


alanakbik:
@LukasOpp thanks for adding this!

Issue #1827, opened 2020-08-21
Error when loading trained model


lucaventurini:
When I load a model trained (on a different machine) with TransformerDocumentEmbeddings, the script fails with the following error:

`Not found: "/root/.cache/torch/transformers/0c370616ddfc06067c0634160f749c2cf9d8da2c50e03a2617ce5841c8df3b1d.309f0c29486cffc28e1e40a2ab0ac8f500c203fe080b95f820aa9cb58e5b84ed": No such file or directory Error #2 `

Apparently, it's looking for some file in cache from transformers, even if I expected the model to contain everything needed for the prediction.

The workaround is to call again, before the loading,
`_ = TransformerDocumentEmbeddings('xlm-roberta-base')`
but you need to know which pre-trained model was used in the training, in this case 'xlm-roberta-base'.

alanakbik:
I think the problem is that it serializes a path specific to the machine it was trained on. I am seeing a similar error. What transformers version are you running? I think it might work with the current transformers.

alanakbik:
Correction: I think the error persists even with the most current version.

PR #1978, opened 2020-11-23
GH-1827: Pretrained TARS model and small fixes

### LINKED ISSUE #1977: Add TARS base model
For the upcoming release we want to include a good base TARS model to train few-shot classifiers or do zero-shot prediction. @kishaloyhalder has trained an English TARS model using 9 classification tasks. 

This issue tracks the progress of adding this model to Flair.

alanakbik:
This PR adds a pretrained TARS model (closes #1977). Load and use for zero-shot like this: 

```python
# 1. Load our pre-trained TARS model for English
tars = TARSClassifier.load('tars-base')

# 2. Prepare a test sentence
sentence = flair.data.Sentence("I am so glad you liked it!")

# 3. Define some classes that you want to predict using descriptive names
classes = ["happy", "sad"]

#4. Predict for these classes
tars.predict_zero_shot(sentence, classes)

# Print sentence with predicted labels
print(sentence)
```

It also adapts the single label prediction logic so that confidence must be above threshold 0.5 (@kishaloyhalder does this make sense?), a convenience function to predict all TARS tasks at once, a convenience method to pass just one label as string instead of a list and updates the TARS tutorial.

### DIFF
--- a/flair/models/text_classification_model.py
+++ b/flair/models/text_classification_model.py
@@ -594,7 +594,7 @@ class TARSClassifier(TextClassifier):
     def add_and_switch_to_new_task(self,
                                    task_name,
                                    label_dictionary: Union[List, Set, Dictionary, str],
-                                   multi_label: bool = None,
+                                   multi_label: bool = True,
                                    multi_label_threshold: float = 0.5,
                                    label_type: str = None,
                                    beta: float = 1.0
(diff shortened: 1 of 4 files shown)


flair/models/text_classification_model.py
```diff
@@ -848,20 +845,29 @@ def _get_multi_label(self, label_scores) -> List[Label]:
 
     def _get_single_label(self, label_scores) -> List[Label]:
         conf, idx = torch.max(label_scores, 0)
-        label = self.label_dictionary.get_item_for_index(idx.item())
-
-        return [Label(label, conf.item())]
+        # TARS does not do a softmax, so confidence of the best predicted class might be very low.
+        # Therefore enforce a min confidence of 0.5 for a match.
+        if conf > 0.5:
```
kishaloyhalder, line 850: This could be problematic in some cases when TARS figures "True" for some sample with max score say 0.3. Let's leave it to the end-user to determine the threshold in a post-hoc manner.

kishaloyhalder (changes requested):

PR #1989, opened 2020-11-26
Add support for datasets in UFSAC format for WSD task

stolzenp:


flair/datasets/sequence_labeling.py
```diff
@@ -1508,6 +1508,308 @@ def __init__(
...
+            self,
+            base_path: Union[str, Path] = None,
+            in_memory: bool = True,
+            document_as_sequence: bool = False,
+    ):
+        """
+        Initialize a custom corpus with any two WSD datasets in the UFSAC format. This is only possible if you've
+        manually downloaded the WSD datasets in UFSAC format to your machine.
+        Obtain the most recent datasets from https://drive.google.com/file/d/1Oigo3kzRosz2VjyA44vpJZ58tDFyLRMO and copy
+        any two of the datasets in a folder called 'wsd_ufsac'. Then rename the files in the folder to train.xml and
```
alanakbik, line 1691: This is inconvenient since this requires me to always rename different files to `train.xml` if I want to train on a different dataset. It would be easier if train and test file names are passed as arguments in the contrunctor.

flair/datasets/sequence_labeling.py
```diff
@@ -1508,6 +1508,308 @@ def __init__(
...
+            convert_ufsac_to_conll(data_file=Path(data_folder / "test.xml"), encoding="latin-1")
+
+            super(WSD_UFSAC, self).__init__(
+                data_folder,
+                columns,
+                tag_to_bioes=None,
+                encoding="latin-1",
+                in_memory=in_memory,
+                document_separator_token=None if not document_as_sequence else "-DOCSTART-",
+                train_file='train.xml',
```
alanakbik, line 1747: instead of hard-coding this, allow this to be specified through the constructor.

alanakbik (changes requested):
The method is currently difficult to use since it requires frequent renaming of files.

alanakbik (approved):

alanakbik:

PR #1995, opened 2020-11-30
Studienprojekt pha

pharnisch:
@alanakbik: As discussed before, an easy implementation of a working Sequence Tagger. (as a starting point of my study project)


alanakbik:
@pharnisch thanks for adding this! I will merge it and then move it to a special "sandbox" module to make it more clear that this is only intended as a starting point for model development.

PR #2017, opened 2020-12-09
Optimize WSD_UFSAC corpus and add option to handle multiple senses for one word

stolzenp:


alanakbik:

PR #2045, opened 2020-12-27
HUN_NER_EXTENSION

alibektas:
Hungarian NER Business Corpus has been added. There are two assumptions which , above all , need to be reviewed. These assumptions are commented in sequence_labeling.py

### DIFF
--- a/flair/datasets/sequence_labeling.py
+++ b/flair/datasets/sequence_labeling.py
@@ -604,65 +604,14 @@ class BUSINESS_HUN(ColumnCorpus):
             path_to_zipped_corpus , 
             data_folder,
             mode = "zip",
-            keep= False #We can delete the .zip file once we extracted it
+            keep= True 
             )
-
-        """
-            Please review these 2 assumptions , as the writer was not an expert on the field.
-
-            Assumption 1:
-            We will try to keep the ratio of elements referring to entities the same in each of 3 files.
-            In order to achieve this distribution , we will split the corpus from sentences.
-            For that we will look for periods.
-
-            Assumption 2:
-            We will split the corpus into train dev , test , train sets each of which will respectively 
-            receive 60% , 20% , 20%. You can modify this by changing the next line. 
-        """
-        portions = {
-            'train' : (0,60) , 
-            'dev' : (60,80) , 
-            'test' : (80,100)
-            }
-
-        dev = open(data_folder / "dev.txt" , "w" , encoding = "latin1")
-        test = open(data_folder / "test.txt" , "w" , encoding = "latin1" )
-        train = open(data_folder / "train.txt" , "w" , encoding = "latin1")
-
-
-        from random import randint
-        with open(data_folder / "hun_ner_corpus.txt" , encoding = "latin1") as corpus:
-            sentence = []
-            
-            while True:  
-                line = corpus.readline()
-                if not line:
-                    break
-                elif line == '\n':
-
-                    sentence.append(line) 
-                    portion = randint(0,100) 
-                    
-                    if portions['test'][0] <= portion and portion < portions['test'][1]:
-                        test.writelines(sentence)
-                    elif portions['dev'][0] <= portion and portion < portions['dev'][1]:
-                        dev.writelines(sentence)
-                    else:
-                        train.writelines(sentence)
-                    sentence = []
-                    
-                else :
-                    sentence.append(line)        
-
-        dev.close()
-        test.close()
-        train.close()
-
 
         super(BUSINESS_HUN, self).__init__(
             data_folder,
             columns,
-            tag_to_bioes=None,
+            train_file='hun_ner_corpus.txt',
+            tag_to_bioes= tag_to_bioes,
             encoding="latin-1",
             in_memory=in_memory,
             document_separator_token=None if not document_as_sequence else "-DOCSTART-",
--- a/flair/datasets/sequence_labeling.py
+++ b/flair/datasets/sequence_labeling.py
@@ -8,7 +8,7 @@ from typing import Union, Dict, List
 import flair
 from flair.data import Corpus, MultiCorpus, FlairDataset, Sentence, Token
 from flair.datasets.base import find_train_dev_test_files
-from flair.file_utils import cached_path, unpack_file
+from flair.file_utils import cached_path, unpack_file , unzip_file
 
 log = logging.getLogger("flair")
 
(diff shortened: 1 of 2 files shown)


flair/datasets/sequence_labeling.py
```diff
@@ -564,6 +564,112 @@ def __init__(
         )
 
 
+class BUSINESS_HUN(ColumnCorpus):
+    def __init__(
+            self,
+            base_path: Union[str, Path] = None,
+            tag_to_bioes: str = "ner",
```
alanakbik, line 571: This argument is currently unused. You can add it to the `super()` call below as `tag_to_bioes=tag_to_bioes` so that the `ColumnCorpus` can use the `tag_to_bioes` setting.

flair/datasets/sequence_labeling.py
```diff
@@ -564,6 +564,112 @@ def __init__(
...
+        dataset_name = self.__class__.__name__.lower()
+
+        # default dataset folder is the cache root
+        if not base_path:
+            base_path = Path(flair.cache_root) / "datasets"
+        data_folder = base_path / dataset_name
+
+        # download data if necessary
+        hun_ner_path = "https://rgai.sed.hu/sites/rgai.sed.hu/files/business_NER.zip"
+        path_to_zipped_corpus = cached_path(hun_ner_path, Path("datasets") / dataset_name)
```
alanakbik, line 602: This will re-download the corpus each time the method is called, causing unnecessary traffic. Easiest option would be to set `keep=True` below so this does not happen.

flair/datasets/sequence_labeling.py
```diff
@@ -564,6 +564,112 @@ def __init__(
...
+                    sentence = []
+                    
+                else :
+                    sentence.append(line)        
+
+        dev.close()
+        test.close()
+        train.close()
+
+
```
alanakbik, line 661: Only one blank line here.

flair/datasets/sequence_labeling.py
```diff
@@ -564,6 +564,112 @@ def __init__(
...
+        path_to_zipped_corpus = cached_path(hun_ner_path, Path("datasets") / dataset_name)
+        unpack_file(
+            path_to_zipped_corpus , 
+            data_folder,
+            mode = "zip",
+            keep= False #We can delete the .zip file once we extracted it
+            )
+
+        """
+            Please review these 2 assumptions , as the writer was not an expert on the field.
```
alanakbik, line 611: Makes a lot of sense! However, this is already in-built into Flair - if you instantiate a `Corpus` with only a train file, it will perform automatic sampling of splits (without creating new txt files) using the original sentence boundaries. So this code until the super() call can be safely removed - the super() call will automatically take care of sampling.

flair/datasets/sequence_labeling.py
```diff
@@ -564,6 +564,112 @@ def __init__(
...
+
+            Assumption 1:
+            We will try to keep the ratio of elements referring to entities the same in each of 3 files.
+            In order to achieve this distribution , we will split the corpus from sentences.
+            For that we will look for periods.
```
alanakbik, line 616: Sentence boundaries are by default marked as empty lines in the text file, so the punctuation heuristic is not necessary.

flair/datasets/sequence_labeling.py
```diff
@@ -564,6 +564,112 @@ def __init__(
...
+
+        dev.close()
+        test.close()
+        train.close()
+
+
+        super(BUSINESS_HUN, self).__init__(
+            data_folder,
+            columns,
+            tag_to_bioes=None,
```
alanakbik, line 665: Add the following here: 
```python
            train_file='hun_ner_corpus.txt',
            tag_to_bioes=tag_to_bioes,
```
This will indicate to the super object that the file `hun_ner_corpus.txt` is the training file. The object will then realize that no test and dev splits are passed and automatically sample these splits from the train file.

alanakbik (changes requested):
@alibektas thanks for adding this! 

I've added some in-line comments on re-downloading and using the auto-sampling logic of Flair instead.

flair/datasets/sequence_labeling.py
```diff
@@ -564,6 +564,61 @@ def __init__(
...
+
+        # default dataset folder is the cache root
+        if not base_path:
+            base_path = Path(flair.cache_root) / "datasets"
+        data_folder = base_path / dataset_name
+
+        # download data if necessary
+        hun_ner_path = "https://rgai.sed.hu/sites/rgai.sed.hu/files/business_NER.zip"
+        path_to_zipped_corpus = cached_path(hun_ner_path, Path("datasets") / dataset_name)
+        unpack_file(
```
alanakbik, line 603: The file is now re-unpacked every time the corpus is instantiated. Can you check if the unpacked file is already there and only if not unpack the zip?

flair/datasets/sequence_labeling.py
```diff
@@ -564,6 +564,61 @@ def __init__(
...
+            path_to_zipped_corpus , 
+            data_folder,
+            mode = "zip",
+            keep= True 
+            )
+
+        super(BUSINESS_HUN, self).__init__(
+            data_folder,
+            columns,
+            train_file='hun_ner_corpus.txt',
```
alanakbik, line 613: Can you also add the line 
`label_name_map={'0': 'O'}, `
here? Most datasets use O ("out") instead of 0 to denote that something is not an entity. This line will map all tags from 0 to O.

alanakbik (changes requested):
@alibektas thanks, getting there :) However, I noticed some more problems. If you execute this code: 

```python
# load corpus
corpus = BUSINESS_HUN()

# get all NER tags in corpus
tag_dictionary = corpus.make_tag_dictionary('ner')
print(tag_dictionary)
```

You get a list of all NER tags in the corpus, as follows: 
~~~
Dictionary with 17 tags: <unk>, O, 0, I-ORG, I-PER, I-LOC, I-MISC, B-ORG, B-LOC, B-MISC, -, 1, milliárd, millió, és, <START>, <STOP>
~~~

So, while good tags like "B-LOC" and "I-PER" are in there, there are also some errors like "milliárd" and "0". I think the problem lies in some lines list two words separated by a whitespace as a single token line. like the token "56 millió" here:

~~~
A	0
Budapesti	I-ORG
Árutõzsde	I-ORG
devizapiacán	0
12,68	0
milliárd	0
forint	0
,	0
a	0
gabonaszekcióban	0
,	0
56 millió	0
forint	0
értékben	0
kereskedtek	0
szerdán	0
.	0
~~~

To address this, you must make sure that lines are split only on tabs and not whitespaces. Check the `TURKU_NER` object for how this is done.

alibektas:
I managed to reduce tags down to following:
<unk>, O, I-PER, I-ORG, I-MISC, I-LOC, B-ORG, B-LOC, <START>, <STOP>

Reading the source Dictionary made me conclude that <unk> will be in any case present. With that in mind, I guess this error has also been resolved.

alanakbik (approved):

alanakbik:
@alibektas thanks for adding this!

PR #2059, opened 2021-01-05
GH-2058: do not shuffle first epoch during training

### LINKED ISSUE #2058: Do not shuffle on first epoch
Normally, we shuffle sentences at each epoch during training in the ModelTrainer class. However, in some cases it makes sense to see sentences in their natural order during the first epoch, and shuffle only from the second epoch onward. 

This PR monitors the progress of adding this feature to Flair.

alanakbik:
Closes #2058


alanakbik:
@cod3licious

cod3licious (approved):
Didn't think it was really _that_ simple :D

PR #2073, opened 2021-01-14
Fix linear input dimension if the reproject 

djstrong:
is set to different dimension than length of embeddings.


alanakbik:
@djstrong thanks for fixing this - and please excuse the delay in approving this.

PR #2086, opened 2021-01-29
Added the DocumentTFIDFEmbeddings

### LINKED ISSUE #2085: TF-IDF DocumentEmbeddings
It would be nice to have some old-school embeddings, so why not start with the legendary TF-IDF document embeddings. These are often good baselines, and additionally they keep NLP veterans nostalgic, if not happy.

Good artists copy, great artists steal, building on the shoulder of giants... We could just wrap [Scikit-learn's TfidfVectorizer](https://scikit-learn.org/stable/modules/generated/sklearn.feature_extraction.text.TfidfVectorizer.html).

yosipk:
This PR adds DocumentTFIDFEmbeddings in a way described in issue #2085. 

Closes #2085.

### DIFF
--- a/flair/embeddings/__init__.py
+++ b/flair/embeddings/__init__.py
@@ -23,6 +23,7 @@ from .token import NILCEmbeddings
 from .document import DocumentEmbeddings
 from .document import TransformerDocumentEmbeddings
 from .document import DocumentPoolEmbeddings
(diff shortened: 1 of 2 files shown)


alanakbik:
@yosipk thanks for adding this!

yosipk:
@alanakbik thanks for fixing it!

Issue #2076, opened 2021-01-15
Improve model loading, label tokenization, and sampling for TARS


kishaloyhalder:
**Is your feature/enhancement request related to a problem? Please describe.**
* The `TARSClassifier` constructor creates an instance which loads the `bert-base-uncased` model first. Then it loads the pre-trained model provided. As a result, it unnecessarily downloads the pre-trained `bert-base-uncased` model if it not in the cache.
* The default tokenizer splits the `[SEP]` incorrectly.
* In case there are multiple labels in the ground truth for a text, there is a small probability that the sampler might pick a positive label as negative.

**Describe the solution you'd like**
* Improve the `_init_model_with_state_dict` method to load the pre-trained model directly.
* Disbale `default_tokenizer` when constructing the `<label, text>` tuple.
* Improve the sampling to check all positive labels while sampling in a multi-labelling context.

**Additional context**
We do not expect any major effect on normal multi-class classification tasks. Performance for multi-labelling tasks might benefit from this change.

PR #2090, opened 2021-02-01
GH-2076: Improve model loading, label tokenization, sampling for TARS

kishaloyhalder:


alanakbik:
@kishaloyhalder thanks for adding this!

PR #2091, opened 2021-02-02
add to_plain_string method to Span class

mauryaland:


alanakbik:
@mauryaland thanks for adding this!

PR #2102, opened 2021-02-10
Added GermEval 18 Offensive Language dataset

lukasgarbas:
Added GermEval 2018 data for binary text classification of German tweets. The dataset was taken from https://github.com/uds-lsv/GermEval-2018-Data.

### DIFF
--- a/flair/datasets/document_classification.py
+++ b/flair/datasets/document_classification.py
@@ -1543,20 +1543,23 @@ class TREC_6(ClassificationCorpus):
 
 class GERMEVAL_2018_OFFENSIVE_LANGUAGE(ClassificationCorpus):
     """
-    GermEval 2018 Task 1 corpus for identification of offensive language.
-    Classifying German tweets into 2 coarse-grained categories OFFENSIVE or OTHER.
+    GermEval 2018 corpus for identification of offensive language.
+    Classifying German tweets into 2 coarse-grained categories OFFENSIVE and OTHER
+    or 4 fine-grained categories ABUSE, INSULT, PROFATINTY and OTHER.
     """
 
     def __init__(self,
                  base_path: Union[str, Path] = None,
+                 tokenizer: Union[bool, Callable[[str], List[Token]], Tokenizer] = SegtokTokenizer(),
                  memory_mode: str = 'full',
-                 tokenizer: Union[bool, Callable[[str], List[Token]], Tokenizer] = SpaceTokenizer(),
+                 fine_grained_classes: bool = False,
                  **corpusargs):
         """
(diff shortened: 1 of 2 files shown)


flair/datasets/document_classification.py
```diff
@@ -1541,6 +1541,73 @@ def __init__(self,
...
+class GERMEVAL_2018_OFFENSIVE_LANGUAGE(ClassificationCorpus):
+    """
+    GermEval 2018 Task 1 corpus for identification of offensive language.
+    Classifying German tweets into 2 coarse-grained categories OFFENSIVE or OTHER.
+    """
+
+    def __init__(self,
+                 base_path: Union[str, Path] = None,
+                 memory_mode: str = 'full',
+                 tokenizer: Union[bool, Callable[[str], List[Token]], Tokenizer] = SpaceTokenizer(),
```
alanakbik, line 1553: Perhaps change the default tokenizer to `SegtokTokenizer`? It seems the tweets are not tokenized, so the whitespace tokenizer will create some words that include punctuation.

flair/datasets/document_classification.py
```diff
@@ -1541,6 +1541,73 @@ def __init__(self,
         )
 
 
+class GERMEVAL_2018_OFFENSIVE_LANGUAGE(ClassificationCorpus):
+    """
+    GermEval 2018 Task 1 corpus for identification of offensive language.
+    Classifying German tweets into 2 coarse-grained categories OFFENSIVE or OTHER.
+    """
+
+    def __init__(self,
```
alanakbik, line 1550: Could you add one more boolean constructor argument called `fine_grained_classes` (or so) that's set to `False` by default? If set to `True` you could load the fine-grained offense classes instead. 

To do this in the current logic, you'd need to generate separate FastText-formatted files for the fine-grained case.

alanakbik (changes requested):
Thanks for adding this! Could you check out the review comments?

flair/datasets/document_classification.py
lukasgarbas: Done! Having SegtokTokenizer as the default makes more sense for tweets.

flair/datasets/document_classification.py
lukasgarbas: Sure. I added support for the fined-grained case. This way one could use the data from both GermEval 2018 tasks (binary and 4-way classification). As you've mentioned before, this requires creating separate files for both tasks.

alanakbik (approved):

alanakbik:
@lukasgarbas thanks for adding this!

PR #2108, opened 2021-02-12
huggingface_hub: drop the first param name

julien-c:
for compatibility with huggingface_hub>=0.0.2

we handle datasets now using the same git-based system so that first param is now named repo_id not model_id 😇


stefan-it:
:+1:

stefan-it:
Thanks @julien-c , I can confirm that is working with latest *0.0.2* version (and also with *0.0.1*)!

PR #2113, opened 2021-02-25
Fix the evaluate() method in the SimilarityLearner class

tpsatish95:
The evaluate method does not have a similar function signature as other models, which is causing errors when training it.


alanakbik:
@tpsatish95 thanks for fixing this!

PR #2125, opened 2021-03-02
add quiet=False in gdown.download

ma7555:
a small edit that fixes this bug in WIKIANN corpus download with gdown
```
TypeError                                 Traceback (most recent call last)
<ipython-input-7-ea5a420416fe> in <module>()
----> 1 wikiann_corpus: Corpus = WIKIANN(languages=['ar'])
      2 print(wikiann_corpus)

/usr/local/lib/python3.7/dist-packages/flair/datasets/sequence_labeling.py in __init__(self, languages, base_path, tag_to_bioes, in_memory)
   2043 
   2044                 # download from google drive
-> 2045                 gdown.download(url, str(language_folder / language) + '.tar.gz')
   2046 
   2047                 # unzip

TypeError: download() missing 1 required positional argument: 'quiet'
```

### DIFF
--- a/flair/datasets/sequence_labeling.py
+++ b/flair/datasets/sequence_labeling.py
@@ -2507,7 +2507,7 @@ class WIKIANN(MultiCorpus):
                 url = google_drive_path + google_id
 
                 # download from google drive
-                gdown.download(url, str(language_folder / language) + '.tar.gz')
+                gdown.download(url, str(language_folder / language) + '.tar.gz', quiet=False)
 
                 # unzip
                 print("Extract data...")


alanakbik:
Hi @ma7555 thanks for the PR! This error does not occur for me. Which gdown version are you using?

ma7555:
Hello @alanakbik, I am using gdown v3.6.4

ma7555:
tried with v3.12.2, it works on this version. please specify a  gdown version in requirments.txt

BastianZim:
Hi @alanakbik, is there a reason gdown has to be `==3.12.2` or could it also be `>=3.12.2`?

alanakbik:
@BastianZim does it work for you with a higher version? We could change the dependency.

BastianZim:
@alanakbik I think `3.12.2` is currently the highest version. The reason I'm asking is that I'm one of the maintainers of the flair [conda-forge package](https://anaconda.org/conda-forge/python-flair) and we generally prefer relaxing the upper bound of packages so that the solver is more flexible when a new version of a dependency is released. 
But if I understand it correctly, you're just narrowing it so much because you aren't testing for anything else, is that correct?

PR #2129, opened 2021-03-02
in example fixed the argument from layer_mean to use_scalar_mix

ajay-sreeram:
In `TransformerWordEmbeddings` example with scalar_mix, updated the argument from layer_mean to use_scalar_mix


stefan-it:
Hi @ajay-sreeram ,

this parameter was recently changed to `layer_mean`, but it is only available in `master` version of Flair and not in a tagged release:

https://github.com/flairNLP/flair/blob/9b46ccd36f62ffd8f2649f34afe5877db403b59e/flair/embeddings/token.py#L790

Whenever we release the upcoming *0.8* version of Flair - @alanakbik  is working on it - then `layer_mean` will be available.

alanakbik:
We'll release soon, so I'm closing this PR! Thanks for the feedback!

PR #2141, opened 2021-03-09
add DocumentCNNEmbeddings

djstrong:
I have implemented a simple CNN document embeddings based on Kim's CNN model.

Tests on TREC_6 give worse results than using RNN.

However, I am not sure if all code copied from `DocumentRNNEmbeddings` make sense in CNN.


alanakbik:
@djstrong thanks for adding this!

Issue #2145, opened 2021-03-10
enhancement/save_model_each_k_epochs


troeshust96:
I use an external custom script to check the model quality and compute the metrics. In order to monitor overfitting and to perform an early stopping, I need to run this script not only on the final best-model, but also on the models resulting from the intermediate epochs. There is already a `save_model_at_each_epoch` parameter [(this line)](https://github.com/flairNLP/flair/blob/master/flair/trainers/trainer.py#L602) in the Flair, but due to the large size of the model, there is not enough disc space to save each epoch.
It would be nice to enhance this parameter and add the ability to save the model every k-th epoch.

PR #2146, opened 2021-03-10
GH-2145: Added new param to save model each k epochs

troeshust96:
Solution for https://github.com/flairNLP/flair/issues/2145


alanakbik:
@troeshust96 thanks for adding this!

PR #2148, opened 2021-03-12
Adding the Reddit Entity Linking Dataset

aynetdia:
Adding support for the [Reddit Entity Linking Dataset](https://arxiv.org/abs/2101.01228). 

Only gold annotations were considered. The dataset contains both the post titles and the comment threads. It is transformed into a CoNLL-like column format for consistency with other Flair Datasets.


alanakbik:
@aynetdia thanks a lot for adding this - this is really cool!

Issue #2140, opened 2021-03-09
The `multi-X-fast` embeddings appear to be missing


tadejmagajna:
While reviewing [documentation](https://github.com/flairNLP/flair/blob/master/resources/docs/embeddings/FLAIR_EMBEDDINGS.md) I noticed that `multi-X-fast` should be one of the available `FlairEmbeddings `, but they appear to be missing.

For example, when running:
```python
from flair.embeddings import FlairEmbeddings
FlairEmbeddings("multi-forward-fast")
```
I get:
```
ValueError: The given model "multi-forward-fast" is not available or is not a valid path.
```
This happened both on 0.7 and 0.8 versions of Flair.

Given that [here](https://github.com/flairNLP/flair/blob/69d7c7d523b9111956d6fecd781d0b4654a4b13a/flair/embeddings/token.py#L412) a version called `multi-v0-forward-fast` does exist, but no other similar versions exist in Flair's [model index](https://flair.informatik.hu-berlin.de/resources/embeddings/flair/), I would presume the model dictionary key simply needs to be renamed in the code from `multi-v0-forward-fast` to `multi-forward-fast`. Unless there is a v1 version that was never uploaded.

Happy to open a PR for this.

alanakbik:
Hello @tadejmagajna thanks for spotting this! You're right, we don't have a "v1" for the fast model. A PR with your fix would be very much appreciated!

PR #2150, opened 2021-03-13
Fix references to multi-X-fast Flair embedding models

tadejmagajna:
This PR fixes: https://github.com/flairNLP/flair/issues/2140.

The [documentation](https://github.com/flairNLP/flair/blob/master/resources/docs/embeddings/FLAIR_EMBEDDINGS.md) suggests that the CPU friendly mix of corpora Flair embedding models should be referenced as `multi-X-fast`, but they were defined as `multi-v0-X-fast` in the `PRETRAINED_MODEL_ARCHIVE_MAP` dict [here](https://github.com/flairNLP/flair/blob/69d7c7d523b9111956d6fecd781d0b4654a4b13a/flair/embeddings/token.py#L412).

This issue resulted in errors when loading the model using the ID mentioned in the documentation. This PR fixes this.


alanakbik:
@tadejmagajna thanks for spotting and fixing this!

PR #2154, opened 2021-03-15
Japanese ner

FranziskaKuhls:
Added a japanese NER Corpus for usage in FlairNLP.


alanakbik:
@FranziskaKuhls thanks for adding this!

PR #2161, opened 2021-03-17
Gh 2117 more flexibility on main metric

MLDLMFZ:
Allow different metric than micro f1 as main score (calculated in the eval method of models), which is used for monitoring model performance on dev data during training, saving of best model and learning rate scheduling


MLDLMFZ:
More flexibility on the main metric used is currently only implemented for the text_classification model. If you think that's a good idea in general, we could also implement it for other model types.

alanakbik:
@MLDLMFZ thanks for adding this - very useful!

MLDLMFZ:
Thanks for merging!

PR #2162, opened 2021-03-17
Gh 2118 return predictions during evaluation

MLDLMFZ:
In the evaluate method of models, give the option to return the predictions obtained on the eval data


alanakbik:

PR #2163, opened 2021-03-17
Gh 2119 saving best model for cyclic lr

MLDLMFZ:
Save the best model seen so far (based on dev data) also if the cyclic learning rate scheduler is used


alanakbik:

MLDLMFZ:
Thanks for merging!

PR #2165, opened 2021-03-17
Gh 2121 improvements best model saving

MLDLMFZ:
Encode epoch from which the best model is in the filename of the saved model, give option to save not only the best model but also the corresponding checkpoint


alanakbik:

MLDLMFZ:
Thanks for merging!

Issue #2159, opened 2021-03-16
Slovenian word embeddings ID is incorrect in the documentation


tadejmagajna:
The [documentation](https://github.com/flairNLP/flair/blob/master/resources/docs/embeddings/CLASSIC_WORD_EMBEDDINGS.md) suggests we should use `si` as the ID for the Slovenian FastText embeddings. But if we use the `si` ID we get the following results:

When running:
```python
from flair.embeddings import WordEmbeddings
WordEmbeddings('si')
```
We get:
`OSError: HEAD request failed for url https://flair.informatik.hu-berlin.de/resources/embeddings/token/si-wiki-fasttext-300d-1M.vectors.npy with status code 404.`

However, if we instead use the `sl` ID, the embeddings will load correctly. I also manually verified the `sl` embeddings are indeed the Slovene.

The documentation says that the IDs are language codes, but the ID used for Slovene (`si`) is actually a country code for Slovenia. Whereas the language code for Slovenian language is actualy `sl` (the ID that actually works right now).

The fix involves simply updating `si` to `sl` in the [documentation](https://github.com/flairNLP/flair/blob/master/resources/docs/embeddings/CLASSIC_WORD_EMBEDDINGS.md).

Happy to open a PR for this 1 character change

alanakbik:
@tadejmagajna thanks for spotting this! A PR is always appreciated :)

PR #2166, opened 2021-03-17
GH-2159: Fixing ID for Slovene word embeddings in the doc

tadejmagajna:
This PR fixes https://github.com/flairNLP/flair/issues/2159 .
The ID for Slovene embeddings in the doc is incorrect and raises an error if passed to the `WordEmbeddings` constructor. This PR fixes it.

I also checked other pars of Flair documentation for references to Slovene and they seem to be correct.


alanakbik:
@tadejmagajna thanks for correcting this!

PR #2168, opened 2021-03-18
Datasets: Support for conll yago

routerdieb:


alanakbik:
@routerdieb thanks for adding this!

Issue #2174, opened 2021-03-22
IMDB class does not return original splits with `rebalance_corpus=False`


kishaloyhalder:
**Describe the bug**
IMDB class does not return original splits with `rebalance_corpus=False`. It always splits the data in 90-10 between train and test.

**To Reproduce**
Use the following to reproduce:

`
from flair.datasets import IMDB
corpus = IMDB(rebalance_corpus=False)

print(corpus.train)
print(corpus.dev)
print(corpus.test)
`

**Expected behavior**
It should have 25K point in `train` and `dev` combined and the same in test.

PR #2175, opened 2021-03-22
GH-2174: fixed IMDB data splitting logic

kishaloyhalder:
Fixed a bug described [here](https://github.com/flairNLP/flair/issues/2174).


alanakbik:
@kishaloyhalder thanks for spotting and fixing this!

PR #2188, opened 2021-03-26
Add AQMAR NER modified corpus

megantosh:
Draft of a Modern Standard Arabic dataset from 28 Wikipedia articles used in https://www.aclweb.org/anthology/W19-4607/

### DIFF
--- a/flair/datasets/__init__.py
+++ b/flair/datasets/__init__.py
@@ -8,6 +8,7 @@ from .base import MongoDataset
 from .sequence_labeling import ColumnCorpus
 from .sequence_labeling import ColumnDataset
 from .sequence_labeling import ANER_CORP
+from .sequence_labeling import AQMAR
 from .sequence_labeling import BIOFID
 from .sequence_labeling import BIOSCOPE
 from .sequence_labeling import BUSINESS_HUN
--- a/flair/datasets/sequence_labeling.py
+++ b/flair/datasets/sequence_labeling.py
@@ -325,8 +325,7 @@ class ANER_CORP(ColumnCorpus):
         The first time you call this constructor it will automatically download the dataset.
         :param base_path: Default is None, meaning that corpus gets auto-downloaded and loaded. You can override this
         to point to a different folder but typically this should not be necessary.
-        :param tag_to_bioes: NER by default, need not be changed, but you could also select 'pos' to predict
-        POS tags instead
+        :param tag_to_bioes: NER by default, need not be changed.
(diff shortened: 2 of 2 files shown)


alanakbik:
@megantosh why "draft"? Looks good ;)

megantosh:
thought I would mark as draft since this is not the original dataset. Subsequent edits possible (i.e. split corpus into docs, split MISC into MIS0, MIS1, MIS2, MIS3).

Just tested branch after uploading now. PR ready for merge 👍

megantosh:
according to the original dataset's [README.txt](https://github.com/flairNLP/flair/files/6229891/README.txt):
_Richard_Stallman.txt
	MIS-1: Name of software or hardware (e.g. Emacs, Kernel)
	**MIS-2: English entities**_
appears later as mistagged once all MIS-1, -2 etc are merged into only MISC.
Hence it looks like an error, however the name in arabic
_ريتشارد B-PER
ستالمان I-PER_
is tagged correctly as PER.

Problem is, that in another file, MIS-2 represents something else.
E.g. in _atom.txt_ >>	MIS-2: Name of theories (e.g. Dalton Theory)

The dataset has indeed a few issues beyond that, too. I thought, however, to keep these "mistakes" for the sake of comparability with the published scores in https://www.aclweb.org/anthology/W19-4607/ (same paper as above).

Alternatively, we can establish different versions of the dataset if that helps.. What do you think?

megantosh:
for clarification:

ريتشارد B-PER
ستالمان I-PER
 is a transliteration of the english name
Richard B-MISC
Stallman I-MISC

alanakbik:
Good question, but I think merging them into one MISC label for now is a good way to go. 

What about the issue of the B-ENGLISH and B-SPANISH tag? Is this an error in the conversion script? 

I.e.: 
~~~
كان O
الفريق O
يستخدم O
بعض O
الملاعب O
البسيطة O
في O
مدريد B-LOC
للعب O
المباريات O
حتى O
انتقل O
لأول O
ملعب O
كان O
اسمه O
ملعب O
O'Donnell B-SPANISH
في O
عام O
~~~

megantosh:
that is my assumption yes. Same issue exists in [original corpus].(http://www.cs.cmu.edu/~ark/ArabicNER/AQMAR_Arabic_NER_corpus-1.0.zip)

alanakbik:
@megantosh thanks for adding this - will merge this version now!

PR #2191, opened 2021-03-29
Fix XLNet and Transformer-XL Execution

tiagokv:
If you execute the following snippet exception `OverflowError: int too big to convert` is thrown:

```python

model = TransformerWordEmbeddings("xlnet-large-cased", "all", "mean", layer_mean=False)
model.embed(Sentence("This is a test that will not work"))
```

This happens because both XLNet and Transformer-XL do not have a max sequence length and `tokenizer.model_max_length` comes with a very big integer.


alanakbik:
@tiagokv thanks for fixing this!

PR #2198, opened 2021-04-04
Added Yahoo_Answers text classification corpus

hoehnisb:
I added the Yahoo_Answers corpus for text classification.


alanakbik:
@hoehnisb thanks for adding this!

Issue #2202, opened 2021-04-08
Enhance TextDataset for LM training


r0mainK:
Please add the appropriate label to this ticket: enhancement.

**Is your feature/enhancement request related to a problem? Please describe.**

Currently the `TextDataset` class in the `language_model_trainer.py` file is suboptimal, and quite slow. This is due to the usage of for loops instead of comprehension. Furthermore, the way random case flipping is done is error prone for weird character, something which was swept under the rug by the code. Specifically, if random casing changes the length of a sentence, then it is not caught due to to the `if token >= tokens: break` statement, and results in tokens being lost. Since we do it after expanding the vocabulary, it can also lead to token being UNKed for no reason, or not added.

**Describe the solution you'd like**

I would like to refactor the code to make it faster and remove the errors related to casing mentionned above. Specifically, it would entail applying the case changes directly when reading the text, and then doing everything with comprehension. Unless there is reason to do otherwise, I would also like to remove the unused  (to my knowledge) `tokenize` function, and merge the `__getitem__` and `charsplit` methods, as I don't quite get why they are split. 

**Additional context**

Although I did not directy use your trainer as it was a bit too much for my needs, I initially was using a close replica of the `TextDataset`. Modifying it as I described reduced loading time by an order of magnitude, simplified the code, and removed all errors. For reference, here is a snippet of what my `__getitem__` currently looks like:

```python
        with gzip.open(self.files[split_id]) as fin:
            lines = list(map(self.apply_random_case_change, jsonlines.Reader(fin)))
        if self.shuffle_lines:
            random.shuffle(lines)
        lines = (
            [self.char_to_ids.get(char.encode("utf-8"), self.unk_id) for char in line]
            + [self.delimiter_id]
            for line in lines
        )
        ids = torch.tensor([char_id for line in lines for char_id in line], dtype=torch.uint8)
        if not self.is_forward_lm:
            ids = ids.flip(0)
```

alanakbik:
Hello @r0mainK looks very interesting. If you'd like to prepare a PR for this we'd appreciate it. Would be great to speed up training the LM!

r0mainK:
Hey @alanakbik I just prepared the PR for this, feel free to review whenever :)

PR #2203, opened 2021-04-08
Lm dataset improve

r0mainK:
Hey there, this is the PR related to [this issue](https://github.com/flairNLP/flair/issues/2202) I raised earlier. I split the PR in 4 commits:
- remove the unused `TextDataset.tokenize` method, which was used nowhere in the code
- merge the `TextDataset.charsplit` into `TextDataset.__getitem__` as it was ugly to have these separated, and again the method `TextDataset.charsplit` was not used elsewhere
- apply the random case changes before expanding the vocab and counting the tokens, so as to remove two potential errors:
    - a case change which modified a sentence length (happens on rare tokens), which would lead to text shifting and a corrupt input - this was not caught due to this statement: `if token >= tokens: break`
    - a case change which led to a character being added to the vocab but not used, and the actual character buing UNKed
- split the lines only once (instead of twice if vocab expansion enabled) and use list comprehension instead of nested `for` loops to create the `ids`

I ran tests and all seems to work. I also tested the code on mock inputs on my CPU, and to give you an idea of the speedup with an input of 500,000 lines of length 20, the tensor is created in 1s now, vs over a minute before.

Now obviously during training this won't be a huge gain as we load data asynchronously, however the first batch of each epoch is always loaded in real time, so this will represent a net gain at each epoch. I was training on the recently released Wiki40b, and loading time of each split I had went from ~20min to ~1min, and I trained for over 30 epochs :p 

Anyway tell me what you think !

### DIFF
--- a/flair/trainers/language_model_trainer.py
+++ b/flair/trainers/language_model_trainer.py
@@ -149,32 +149,6 @@ class TextDataset(Dataset):
             line = line.upper()
         return line
 
-    def tokenize(self, path: Union[str, Path]):
-        """Tokenizes a text file."""
-        if type(path) is str:
-            path = Path(path)
-        assert path.exists()
-        # Add words to the dictionary
-        with open(path, "r") as f:
-            tokens = 0
-            for line in f:
-                words = line.split() + ["<eos>"]
-                tokens += len(words)
-                for word in words:
-                    self.dictionary.add_word(word)
-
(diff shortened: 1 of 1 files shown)


alanakbik:
@r0mainK wow the code is much more succinct and faster, thanks for adding this! And sorry for only getting around to review it now!

PR #2211, opened 2021-04-15
Added UD-data for the faroese language

Dabendorf:
I have added the faroese language from the UD-data*, the dataset works and is usable for training on my computer. :) 

* see here: https://github.com/UniversalDependencies/UD_Faroese-FarPaHC


alanakbik:
@Dabendorf thanks for adding this!

PR #2212, opened 2021-04-15
Support for Kinyarwanda language dataset (NER) added

grushety:
Kinyarwanda dataset for NER added from
https://github.com/masakhane-io/masakhane-ner/tree/main/data/kin


alanakbik:
@grushety thanks for adding this!

PR #2214, opened 2021-04-15
Support for Luo language dataset (NER) added

lukasvoege:
Added African Luo language dataset from https://github.com/masakhane-io/masakhane-ner/tree/main/data/luo for NER.


alanakbik:
@lukasvoege thanks for adding this!

PR #2216, opened 2021-04-16
Support for UD Estonian (PoS) added

melismat:
Estonian language dataset from https://github.com/UniversalDependencies/UD_Estonian-EDT/tree/master is added to Flair for PoS.


alanakbik:
@melismat thanks for adding this!

PR #2219, opened 2021-04-16
add coptic POS model

KralaBenjamin:
Adds a POS model for coptic


alanakbik:
@KralaBenjamin thanks for adding this!

PR #2221, opened 2021-04-17
Added Old Church Slavonic language for Part-Of-Speech tagging

iliyanaaaa:
Added OLD CHURCH SLAVONIC for POS


alanakbik:
@iliyanaaaa thanks for adding this!

PR #2222, opened 2021-04-17
IGBO_NER addition to FlairNLP

PhilippThamm:
Added the corpus IGBO_NER for named entity recognition in the language Igbo to the flair framework.

### DIFF
--- a/flair/datasets/sequence_labeling.py
+++ b/flair/datasets/sequence_labeling.py
@@ -1636,7 +1636,7 @@ class IGBO_NER(ColumnCorpus):
             data_folder,
             columns,
             tag_to_bioes=tag_to_bioes,
-            #encoding="latin-1",
+            encoding="utf-8",
             in_memory=in_memory,
             **corpusargs,
         )
--- a/flair/datasets/sequence_labeling.py
+++ b/flair/datasets/sequence_labeling.py
@@ -1598,7 +1598,7 @@ class IGBO_NER(ColumnCorpus):
     def __init__(
             self,
             base_path: Union[str, Path] = None,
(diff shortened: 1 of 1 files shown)


alanakbik:
@PhilippThamm I am getting some weird results when I load this dataset: 

```python
corpus = IGBO_NER()

# print statistics
print(corpus)

# print some example sentences
print(corpus.train[0])
print(corpus.train[0])
print(corpus.train[1])
print(corpus.train[2])
print(corpus.train[3])

# print the NER dictionary (should be only NER tags like B-PER, I-PER, E-PER, B-ORG, I-ORG)
corpus.make_label_dictionary('ner')
```

This prints a lot of weird characters. Perhaps the encoding is wrongly set?

alanakbik (changes requested):
Fix encoding and set tag_to_bioes variable

Dabendorf:
> @PhilippThamm I am getting some weird results when I load this dataset:
> 
> ```python
> corpus = IGBO_NER()
> 
> # print statistics
> print(corpus)
> 
> # print some example sentences
> print(corpus.train[0])
> print(corpus.train[0])
> print(corpus.train[1])
> print(corpus.train[2])
> print(corpus.train[3])
> 
> # print the NER dictionary (should be only NER tags like B-PER, I-PER, E-PER, B-ORG, I-ORG)
> corpus.make_label_dictionary('ner')
> ```
> 
> This prints a lot of weird characters. Perhaps the encoding is wrongly set?

As I understand it, the encoding should be set to utf-8 instead of Latin-1

alanakbik:
@PhilippThamm thanks for adding this!

PR #2223, opened 2021-04-17
Add support for UD_BELARUSIAN-HSE dataset (POS)

IrynaRepinetska-dev:
Context: add support for a new dataset UD_BELARUSIAN-HSE (POST)

Summary:
 Belarusian is an East Slavic language spoken by the Belarusians. 
 Part-of-speech tagging (POS tagging, POST) is the process of marking up a word in a text (corpus) as  corresponding to a particular part of speech based on both its definition and its context.


alanakbik:
@IrynaRepinetska-dev thanks for adding this!

Issue #2224, opened 2021-04-17
<unk> label in ner training


helpmefindaname:
**Describe the bug**
The `corpus.make_tag_dictionary(tag_type)` method adds the default `<unk>` label to the dictionary, 

**To Reproduce**
```
from flair.datasets import BIOFID
corpus = BIOFID() # it doesn't matter which coprus to load, as long is it uses ner.
dictionary = corpus.make_tag_dictionary("ner")
print(dictionary.get_items())
# ['<unk>', 'O', 'S-TME', 'S-LOC', 'S-TAX', 'B-OTHER', 'E-OTHER', 'B-TAX', 'E-TAX', 'I-OTHER', 'B-LOC', 'I-LOC', 'E-LOC', 'S-OTHER', 'B-ORG', 'I-ORG', 'E-ORG', 'B-PER', 'I-PER', 'E-PER', 'B-TME', 'E-TME', 'I-TME', 'S-ORG', 'S-PER', 'I-TAX', '<START>', '<STOP>']
```
notice, `<START>` and `<STOP>` won't be tagged due to the way the viterbi encoding works, however `<unk>` has no purpose and there is no mechanism to exclude it. 

**Expected behavior**
I'd expect the output to be
`['O', 'S-TME', 'S-LOC', 'S-TAX', 'B-OTHER', 'E-OTHER', 'B-TAX', 'E-TAX', 'I-OTHER', 'B-LOC', 'I-LOC', 'E-LOC', 'S-OTHER', 'B-ORG', 'I-ORG', 'E-ORG', 'B-PER', 'I-PER', 'E-PER', 'B-TME', 'E-TME', 'I-TME', 'S-ORG', 'S-PER', 'I-TAX', '<START>', '<STOP>']`

**Environment (please complete the following information):**
 - Windows 10 
 - flair 0.8.0.post1

**Additional context**
We noticed the problem, when we trained a model with few examples and only a single epoch. That lead to the first token being tagged as `<unk>` which is then interpreted as it's own entity when using the `sentence.to_dict(tag_type=tag_type)` method and our automated evaluation script cannot handle unspecified entities.

PR #2225, opened 2021-04-17
remove unk token from Ner labeling

helpmefindaname:
fixes https://github.com/flairNLP/flair/issues/2224


alanakbik:
@helpmefindaname makes sense, thanks for fixing this!

PR #2227, opened 2021-04-18
Added Naija Pidgin NER dataset (Introduction to NLP @ HU Berlin)

mart1nro:
Adds the Naija Pidgin dataset for NER from https://github.com/masakhane-io/masakhane-ner/tree/main/data/pcm


alanakbik:
@mart1nro thanks for adding this!

PR #2230, opened 2021-04-18
Added NER Support for Yoruba

paula813:
NER Support for Yoruba (Introduction task for NLP lecture)

### DIFF
--- a/flair/datasets/__init__.py
+++ b/flair/datasets/__init__.py
@@ -32,6 +32,7 @@ from .sequence_labeling import MIT_RESTAURANT_NER
 from .sequence_labeling import NER_BASQUE
 from .sequence_labeling import NER_FINNISH
 from .sequence_labeling import NER_SWEDISH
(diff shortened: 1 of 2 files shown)


alanakbik (changes requested):
Fix download bug

paula813:
Thanks for the information, I fixed the URL.

alanakbik:
The URL is fixed, but the code still does not run: 

```python
from flair.datasets import YORUBA_NER

corpus = YORUBA_NER()
```

You can check #2227 for a related PR that works. (And please test your code before putting in a PR.)

paula813:
I tested it and I think training a model works now. Only after the training is finished it says:
Process finished with exit code 139 (interrupted by signal 11: SIGSEGV)
But i do not have any idea why that happens?

alanakbik (approved):

alanakbik:
@paula813 thanks,  not sure about the error but the dataset seems to work now!

PR #2231, opened 2021-04-19
Added NER support for Amharic

MarcoMW:
Added support for the Amharic NER dataset to sequence_labeling.py
Source: https://github.com/masakhane-io/masakhane-ner/tree/main/data/amh

Encoding is set to utf-8 since the dataset uses characters from the Ethiopic unicode block.

(Introduction to NLP @ HU Berlin)


alanakbik:
@MarcoMW thanks for adding this!

PR #2232, opened 2021-04-19
flair with the webpages dataset

TatianaMoteuN:


flair/datasets/sequence_labeling.py
```diff
@@ -665,6 +667,54 @@ def __init__(
...
+               parent directory where the webpages folder resides.
+               :param base_path: Path to the webpages corpus (i.e. 'webpage-ner' folder) on your machine
+               :param tag_to_bioes: NER by default, need not be changed, but you could also select 'pos' or 'np' to predict
+               POS tags or chunks respectively
+               :param in_memory: If True, keeps dataset in memory giving speedups in training.
+               :param document_as_sequence: If True, all sentences of a document are read into a single Sentence object
+               """
+        if type(base_path) == str:
+            base_path: Path = Path(base_path)
+
```
alanakbik, line 693: Can you add a routine here that downloads and unpacks the tgz file? The resulting files should then be concatenated into a single file for reading.

alanakbik (changes requested):

alanakbik:
@TatianaMoteuN thanks for adding this - everything seems to work now!

Can you remove the file `test.2.py` from the PR? Then we can merge.

TatianaMoteuN:
> @TatianaMoteuN thanks for adding this - everything seems to work now!
> 
> Can you remove the file `test.2.py` from the PR? Then we can merge.

 @alanakbik the PR does not contain that file. not sure but where do you see it?

alanakbik (approved):

alanakbik:
@TatianaMoteuN thanks for adding this!

PR #2233, opened 2021-04-19
Added parameter to allow negative examples in ClassificationCorpus

### LINKED ISSUE #2213: Multilabel text classfication: examples without label are removed
I'm trying to train a classification model, using training data in the [fasttext format](https://github.com/flairNLP/flair/blob/master/resources/docs/TUTORIAL_6_CORPUS.md#fasttext-format).
There are some example sentences that do not have a label (this is intentional).
However they are filtered out when the data is loaded.
Is this behavior intentional? How can I load them anyway?

schelv:
Closes #2213


alanakbik:
@schelv thanks for adding this! And thanks for including a unit test!

PR #2236, opened 2021-04-20
added kazakh as new language for pos

enricoboos:
Kazakh is a Turkic language spoken in Central Asia.


alanakbik:
@enricoboos thanks for adding this!

Issue #2207, opened 2021-04-13
Issue with cache root


Narsil:
**Describe the bug**

It seems the cache root is wrong if it is defined in the environment instead of the default

**To Reproduce**
Steps to reproduce the behavior (e.g. which model did you train? what parameters did you use? etc.).

Create `test.py`:

```python
from flair.data import Sentence
from flair.models import SequenceTagger

# make a sentence
sentence = Sentence('I love Berlin .')

# load the NER tagger
tagger = SequenceTagger.load('ner')

# run NER over sentence
tagger.predict(sentence)
```
and run 
`FLAIR_CACHE_ROOT=$HOME/.cache/ python test.py`

**Expected behavior**

Example should work out of th  box

**Screenshots**

**Environment (please complete the following information):**
 - OS [e.g. iOS, Linux]: ubuntu
 - Version [e.g. flair-0.3.2]: 0.8.1 (or master)

**Additional context**

Simple fix is to cast to Path too when coming from env.

HallerPatrick:
I also looked into this and can reproduce the bug. In this context I searched every occurrence of `flair.cache_root`, which holds the env value. And every time it is used to construct a `Path` object.

What I therefore would propose is to initially construct the `Path` object in the `flair/__init__.py` and remove every other construction of it in the source code.

Otherwise only a `Path` construction in `flair/models/sequence_tagger.py` has to be added.

I could do a PR for both.

alanakbik:
@HallerPatrick thanks for looking into this - a PR would be great! I guess it would be easier to construct it once in the init.

HallerPatrick:
This can be closed :)

PR #2241, opened 2021-04-21
Moving Path construction of flair.cache_root

HallerPatrick:
This PR solves the bug #2207 and also removes all Path construction and being replaced by one in the modules init file.

This PR introduces a huge diff, but without logic changes.  So should be easily reviewable.


alanakbik:
@HallerPatrick thanks for fixing this - simplifies the code quite a bit!

PR #2242, opened 2021-04-21
Add Luganda_NER Support

Leonard-Hangen:
Luganda or the Ganda language is a Bantu language spoken in the African Great Lakes region. It is one of the major languages in Uganda and is spoken by more than eight million Baganda and other people. ~Source Wikipedia (https://en.wikipedia.org/wiki/Luganda)


alanakbik:
@Leonard-Hangen thanks for adding this!

PR #2243, opened 2021-04-21
added object that downloads the files for corpus with  Swahili language

stw2:
Swahili language is a lingua franca of the African Great Lakes region and other parts of East and Southern Africa. The dataset can be used for Named-entity recognition (NER) task.


alanakbik:
@stw2 thanks for adding this!

PR #2244, opened 2021-04-21
Adds support for POS model LATVIAN

Cubeshit:


alanakbik:
@Cubeshit thanks for adding this!

PR #2245, opened 2021-04-21
add latin dataset

Gloriaky:


alanakbik:
@Gloriaky thanks for adding this!

PR #2246, opened 2021-04-21
Support for POS model Lithuanian

punxund:
Pos model Lithuanian dataset

### DIFF
--- a/flair/datasets/__init__.py
+++ b/flair/datasets/__init__.py
@@ -148,7 +148,7 @@ from .treebanks import UD_OLD_CHURCH_SLAVONIC
 from .treebanks import UD_COPTIC
 from .treebanks import UD_IRISH
 from .treebanks import UD_LATVIAN
-from .treebanks import UD_Lithuanian
+from .treebanks import UD_LITHUANIAN
 
 # Expose all text-text datasets
 from .text_text import ParallelTextCorpus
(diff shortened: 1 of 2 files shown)


flair/datasets/treebanks.py
```diff
@@ -1722,3 +1722,30 @@ def __init__(self, base_path: Union[str, Path] = None, in_memory: bool = True, s
         )
 
         super(UD_LATVIAN, self).__init__(data_folder, in_memory=in_memory, split_multiwords=split_multiwords)
+        
+class UD_Lithuanian(UniversalDependenciesCorpus):
```
alanakbik, line 1726: Can you capitalize the object name like the other datasets, i.e. "UD_LITHUANIAN" instead of "UD_Lithuanian"

flair/datasets/__init__.py
```diff
@@ -148,6 +148,7 @@
 from .treebanks import UD_COPTIC
 from .treebanks import UD_IRISH
 from .treebanks import UD_LATVIAN
+from .treebanks import UD_Lithuanian
```
alanakbik, line 151: Can you capitalize the object name like the other datasets, i.e. "UD_LITHUANIAN" instead of "UD_Lithuanian"

alanakbik (changes requested):
Thanks, it works. Can you capitalize the dataset name?

alanakbik (approved):

flair/datasets/treebanks.py
punxund: Sure, i did it. Could you check one more time?

alanakbik:
@punxund thanks for adding this!

flair/datasets/__init__.py
punxund: Sure

PR #2247, opened 2021-04-21
Capitalizing

punxund:
I rewrite Lithuanian to LITHUANIAN.


alanakbik:
Ah missed that one, thanks!

PR #2248, opened 2021-04-21
Added Galician POS support to treebanks

JoseDosSantos:
Added support for the Galician language. Galician is an Indo-European language spoken mainly in northwestern Spain.


alanakbik:
@JoseDosSantos thanks for adding this!

PR #2250, opened 2021-04-22
Adds most frequent sense (MFS) baseline for WSD evaluation

stolzenp:
This implementation of the MFS baseline picks the first sense annotated in WordNet for the prediction when the model would otherwise make no prediction.


alanakbik:

PR #2257, opened 2021-04-29
Lemmatization

87302380:
Lemmatization based on  Sequence-to-Sequence approach

### DIFF
--- a/flair/models/lemmatization_model.py
+++ b/flair/models/lemmatization_model.py
@@ -163,36 +163,22 @@ class Lemmatization(flair.nn.Model):
             use_teacher_forcing = True if random.random() < self.teacher_forcing_ratio else False
 
             decoder_input = None
-            if use_teacher_forcing:
-                for t in range(max_lemma_lenght):
-                    decoder_input = self._generate_decoder_input(decoder_input, all_seqs, batch_size)
-                    decoder_output, decoder_hidden = self._decode(decoder_input, decoder_hidden, encoder_out)
-                    mask_loss = self._calculate_loss(decoder_output, lemma[t], mask[t])
-                    loss += mask_loss
-                    # Use the correct value as input for the next moment
+
+            for t in range(max_lemma_lenght):
+                decoder_input = self._generate_decoder_input(decoder_input, all_seqs, batch_size)
+                decoder_output, decoder_hidden = self._decode(decoder_input, decoder_hidden, encoder_out)
+                mask_loss = self._calculate_loss(decoder_output, lemma[t], mask[t])
+                loss += mask_loss
+                # Use the correct value as input for the next moment
+                if use_teacher_forcing:
                     decoder_input = lemma[t].view(1, -1)
-
-                    # Using contextualized word embeddings requires recording the prediction results at each moment to generate word vectors.
-                    if self.contextualized_embedding:
-                        for i in range(batch_size):
-                            all_seqs[i] += self.character_dictionary.get_item_for_index(decoder_input[0][i])
-
-            else:
-                for t in range(max_lemma_lenght):
-                    decoder_input = self._generate_decoder_input(decoder_input, all_seqs, batch_size)
-                    decoder_output, decoder_hidden = self._decode(decoder_input, decoder_hidden, encoder_out)
-                    mask_loss = self._calculate_loss(decoder_output, lemma[t], mask[t])
-                    loss += mask_loss
-
-                    # Use the value with the highest probability as the predicted value, i.e. the input for the next moment.
+                else:
                     _, topi = decoder_output.topk(1)
 
-                    decoder_input = torch.LongTensor([[topi[i][0] for i in range(batch_size)]]).to(flair.device)
-
-                    if self.contextualized_embedding:
-                        for i in range(batch_size):
-                            all_seqs[i] += self.character_dictionary.get_item_for_index(decoder_input[0][i])
-
+                # Using contextualized word embeddings requires recording the prediction results at each moment to generate word vectors.
+                if self.contextualized_embedding:
+                    for i in range(batch_size):
+                        all_seqs[i] += self.character_dictionary.get_item_for_index(decoder_input[0][i])
 
         return loss, len(sentences)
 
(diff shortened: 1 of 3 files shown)


flair/models/lemmatization_model.py
```diff
@@ -0,0 +1,516 @@
...
+        # If tag_list is not empty, it is considered that the user needs to use tag.
+        if len(tag_list):
+            self.use_tag: bool = True
+            self.tag_list = tag_list
+            self.tag_dictionary = tag_dictionary
+            self.tag_embeddings = nn.Embedding(len(self.tag_dictionary), self.pre_embeddings.embedding_length)
+            self.tag_embeddings.weight.data.uniform_(-1, 1)
+        else:
+            self.use_tag: bool = False
+
```
alanakbik, line 91: `self.tag_list = tag_list` should be set in the else clause as well, otherwise we encounter is a problem when saving the model (`_get_state_dict` looks for `self.tag_list`).

flair/models/lemmatization_model.py
```diff
@@ -0,0 +1,516 @@
...
+    def _get_state_dict(self):
+        model_state = {
+            "state_dict": self.state_dict(),
+            "hidden_size": self.hidden_size,
+            "pre_embeddings": self.pre_embeddings,
+            "character_dictionary": self.character_dictionary,
+            "n_layers": self.n_layers,
+            "tag_list": self.tag_list,
+            "tag_dictionary": self.tag_dictionary,
```
alanakbik, line 495: tag_list and tag_dictionary are not set if tag_list = [] in the init. This throws an error here that you could fix with:
```python
 "tag_list": self.tag_list if "tag_list" in self.__dict__.keys() else None,
"tag_dictionary": self.tag_dictionary if "tag_dictionary" in self.__dict__.keys() else None,
```

flair/models/lemmatization_model.py
```diff
@@ -0,0 +1,617 @@
...
+        always use teacher forcing. If less than or equal to 0, never use.
+        :param longest_word_length: record the maximum length of words and lemma to use in prediction.
+        Based on experience, the default value is set at 50.
+        :param contextualized_embedding: Whether to use context-sensitive word embedding. Using context-sensitive word
+        embedding can improve the accuracy of the model, but it takes more time for training and prediction.
+        """
+
+        super(Lemmatization, self).__init__()
+        self.hidden_size = hidden_size
+        self.embeddings = embeddings.to(flair.device)
```
alanakbik, line 58: I think the `.to(flair.device)` is unnecessary here. Embeddings are torch modules so they will be moved together with the Lemmatization model if you send it to a device.

87302380: Exactly! I will fix it.

flair/models/lemmatization_model.py
```diff
@@ -0,0 +1,630 @@
...
+    @staticmethod
+    def _fetch_model(model_name) -> str:
+
+        google_drive_url = {
+            "multi": "https://drive.google.com/uc?id=1b1gHiOsqu2bovHlO0MZAeB8NWNNR88CJ",
+            "multi-with-pos": "https://drive.google.com/uc?id=1svkCI1HRGK-rD8AAWgNphGunNA62UEEv"}
+
+        if model_name in google_drive_url:
+            output = str(flair.cache_root) + "/models/lemma/"
+            model_path = gdown.download(google_drive_url[model_name], output, quiet=False)
```
alanakbik, line 621: If the model already exists, it should not download again. Currently, it is downloading the model every time this method is called.

flair/models/lemmatization_model.py
```diff
@@ -0,0 +1,630 @@
...
+
+        # model_map = {
+        #     "lemma": "https://box.hu-berlin.de/seafhttp/files/dcc0676a-d55f-4f48-b555-5e92bc82668b/en-lemma-without-tag.pt",
+        #     "lemma-with-pos": "https://box.hu-berlin.de/seafhttp/files/ed33a77b-25cb-414e-b4fa-e237d9da3d89/en-lemma-with-pos.pt"}
+        #
+        # if model_name in model_map:
+        #     model_path = cached_path(model_map[model_name], cache_dir=Path("models") / "lemma")
+
+        return model_path
```
alanakbik, line 630: You currently get an error if you want to load a model you trained yourself. That's because `model_path` is empty unless the `model_name` exists in `google_drive_url`

flair/models/lemmatization_model.py
```diff
@@ -0,0 +1,630 @@
...
+            else:
+                all_seqs = None
+
+            # If you use teacher forcing, the next input uses the correct value, otherwise use the predicted value as input for the next moment.
+            # The probability of using Teacher Forcing can be modified by adjusting the value of teacher_forcing_ratio.
+            use_teacher_forcing = True if random.random() < self.teacher_forcing_ratio else False
+
+            decoder_input = None
+            if use_teacher_forcing:
+                for t in range(max_lemma_lenght):
```
alanakbik, line 167: This if-clause could be placed only around the code that generates `decoder_input` since the rest is (I think) the same

87302380: Yes, I've refactored the code here.

flair/models/lemmatization_model.py
87302380: This issue should have been fixed.  Now loading will also first check whether the file already exists in the cache.

flair/models/lemmatization_model.py
```diff
@@ -0,0 +1,623 @@
...
+            for t in range(max_lemma_lenght):
+                decoder_input = self._generate_decoder_input(decoder_input, all_seqs, batch_size)
+                decoder_output, decoder_hidden = self._decode(decoder_input, decoder_hidden, encoder_out)
+                mask_loss = self._calculate_loss(decoder_output, lemma[t], mask[t])
+                loss += mask_loss
+                # Use the correct value as input for the next moment
+                if use_teacher_forcing:
+                    decoder_input = lemma[t].view(1, -1)
+                else:
+                    _, topi = decoder_output.topk(1)
```
alanakbik, line 176: This (topi) is not used. I think you need to set the decoder_input with the prediction.

87302380: Yes i forgot that code.

flair/models/lemmatization_model.py
```diff
@@ -0,0 +1,624 @@
...
+    def _attn(self, hidden, encoder_outputs, method='dot'):
+        # TODO: Add another attention score
+        if method == 'dot':
+            attn_energies = self._dot_score(hidden, encoder_outputs)
+            attn_energies = attn_energies.t()
+
+        return F.softmax(attn_energies, dim=1).unsqueeze(1)
+
+    def _dot_score(self, hidden, encoder_output):
+        return torch.sum(hidden * encoder_output, dim=2)
```
alanakbik, line 209: I am a bit confused about how the attention scores are calculated. Why are you doing a sum here? Shouldn't there be a dot product that directly gives the attention scores?

87302380: Yes, because what is required is the value of the dot product as a score. But here is * (Hadamard product). So finally calculate their sum in the second dimension. 

for example, We can set input_lengths= 10.
so the size of hidden is (1, batch  , 2048). The size of  encoder_output is (input_lengths=10, batch  , 2048) 
hidden * encoder_outputs can multiply the hidden vector (2048 numbers) with the corresponding position to the 10 vectors (2048 numbers) of encoder_outputs. And the inner product needs to add up these 2048 products. So use sum in the second dimension.   The final result is (10, batch). 

In fact, this should be equivalent to using bmm.

alanakbik: Ah ok, thanks for clarifying!

alanakbik:
@87302380 thanks for implementing this! The code currently breaks if you initialize with `contextualized_embedding=False`, for instance

tagger = Lemmatization(hidden_size=128,
                       character_dictionary=Dictionary.load('chars'),
                       embeddings=FlairEmbeddings('news-forward-fast'),
                       n_layers=1,
                       `contextualized_embedding=False`,
                       )

Could you check?

87302380:
> @87302380 thanks for implementing this! The code currently breaks if you initialize with `contextualized_embedding=False`, for instance
> 
> ```python
> tagger = Lemmatization(hidden_size=128,
>                        character_dictionary=Dictionary.load('chars'),
>                        embeddings=FlairEmbeddings('news-forward-fast'),
>                        n_layers=1,
>                        contextualized_embedding=False,
>                        )
> ```
> 
> Could you check?

Now it has been fixed!

qanastek:
Hello,

Have you considered merging the branch in the next release ? Contextualized embeddings could be a very interesting improvement for the lemmatization.

Regards

alanakbik:
We have a PR that will be merged in a few minutes: #2531 - this implementation uses a combination of character features and pre-trained embeddings.

qanastek:
Ok, thank you, I stay connected.

djstrong:
@alanakbik So there is no lemmatization in flair? Some time ago you said you are working on this ;)

87302380:
> @alanakbik So there is no lemmatization in flair? Some time ago you said you are working on this ;)
Hi @djstrong !
Actually, Flair does include lemmatization functionality. You can find more information about it in the lemmatizer model：
[https://github.com/flairNLP/flair/blob/master/flair/models/lemmatizer_model.py](lemmatizer_model)

PR #2229, opened 2021-04-18
Stackoverflow NER dataset cleaning

symeneses:
It adds entity mappings, removes sentences with metadata and adds questions/answers counts.
It fixes #2228.

**Expected Results:**
```py
from flair.data import Corpus
from flair.datasets import STACKOVERFLOW_NER

corpus: Corpus = STACKOVERFLOW_NER()
print(corpus)
```
Corpus: 9263 train + 2896 dev + 3108 test sentences

```py
corpus.train[0:3]
```
[Sentence: "If I would have 2 tables"   [− Tokens: 6  − Token-Labels: "If I would have 2 tables <S-Data_Structure>"],
 Sentence: "How do I get this result"   [− Tokens: 6],
 Sentence: "The following query needs to be adjusted , but I dont know how"   [− Tokens: 13]]

**Notes:**
It would be nicer to extend the parameter `comment_symbol` in the `ColumnCorpus` class or create one similar that accepts also a`list`. In this case, the new parameter would take the values in the list `disallowed_list`.

PR #2262, opened 2021-05-02
Add banned sentences parameter in SequenceTagger 

symeneses:
It contains:
- Add banned sentences parameter as mentioned in #2229
- Use the new parameter to load the `STACKOVERFLOW_NER` corpus.
- Ignore tags that were excluded in the paper using `label_name_map`. There was an error when mapping to `O` tag, which was fixed with #2254.

Using the `comment_symbol` was not feasible as it applies at a line level. The new one has to be applied once there is a full sentence.

### DIFF
--- a/flair/datasets/sequence_labeling.py
+++ b/flair/datasets/sequence_labeling.py
@@ -36,6 +36,7 @@ class ColumnCorpus(Corpus):
             skip_first_line: bool = False,
             in_memory: bool = True,
             label_name_map: Dict[str, str] = None,
+            banned_sentences: List[str] = None,
             autofind_splits: bool = True,
             **corpusargs,
     ):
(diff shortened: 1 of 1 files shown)


flair/datasets/sequence_labeling.py
```diff
@@ -1014,49 +1030,26 @@ def __init__(
...
-                    words = []
-                    lines_sentence = []
-                    continue
-                words.append(line_values[0])
-                lines_sentence.append(line)
-            log.info(f"File {file} processed:")
-            log.info(f"The longest sentences has {max_length} words.")
-            log.info(f"Questions: {questions} and Answers: {answers}")
-            log.info(f"Processed sentences: {sentences}.")
+            log.info(f"File {file} has {questions} questions and {answers} answers.")
```
alanakbik, line 1041: Very helpful!

alanakbik:
@symeneses thanks for adding this!

PR #2264, opened 2021-05-02
Changing forward loss variable naming and return values

whoisjones:
Since we're now averaging loss over amount of tokens, this should become clear when looking at the forward loss function in our TARS tagger model.


flair/models/tars_tagger_model.py
```diff
@@ -384,17 +384,17 @@ def _init_model_with_state_dict(state):
 
     def forward_loss(
             self, data_points: Union[List[Sentence], Sentence]
-    ) -> torch.tensor:
+    ) -> tuple[float, int]:
```
alanakbik, line 387: Shouldn't it be `Tuple[torch.tensor, int]`?

PR #2266, opened 2021-05-03
Add Support For WandB Logger

AyushExel:
# This PR adds support for WandB logger. 
## Following notable changes have been introduced:
* Instead of accepting project name and tensorboard log directory as separate variables, this accepts `project_name` and `run_name` and utilizes them to initialize both wandb run and tensorboard directory. ( `self.tensorboard_log_dir = Path(project_name) / run_name`)
* Trainer now accepts `log_model_each_k_epochs` to log models as W&B aritfacts. Assigns aliases-> `epoch`, `best` if the current model is best
* Automatically tracks all metrics
###  Resume runs directly from artifacts: 
There are many ways we can do this as logging model checkpoints is already implemented. Simplest way would be to check if a checkpoint is a W&B link. For example,
```
checkpoint = 'https://wandb.ai/user/project/model_runid:latest'
trainer = ModelTrainer.load_checkpoint(checkpoint, corpus)
```
 But we can go a step further, and add another function that automatically resumes runs from run links like:
```
run_link = `hhtps://wandb.ai/user/project/runid`
trainer.resume_from_artifact(run_link )
```
### Log and version the datasets

cc @alanakbik

### DIFF
--- a/flair/trainers/trainer.py
+++ b/flair/trainers/trainer.py
@@ -18,7 +18,7 @@ except ImportError:
     amp = None
 
 import flair
-import flair.nn
+import flair.nn 
 from flair.data import MultiCorpus, Corpus
 from flair.datasets import DataLoader
 from flair.optim import ExpAnnealLR
@@ -33,6 +33,7 @@ from flair.training_utils import (
 )
 from torch.optim.lr_scheduler import OneCycleLR
 from flair.models import SequenceTagger, TextClassifier
+from flair.wandb_logger import WandbLogger
 import random
 
 log = logging.getLogger("flair")
@@ -46,7 +47,8 @@ class ModelTrainer:
             optimizer: torch.optim.Optimizer = SGD,
             epoch: int = 0,
             use_tensorboard: bool = False,
-            tensorboard_log_dir = None,
+            project_name: str = "flair",
+            run_name: str = '',
             metrics_for_tensorboard = []
     ):
         """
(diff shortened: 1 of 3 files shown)


flair/trainers/trainer.py
```diff
@@ -674,13 +711,37 @@ def train(
...
+                    )
 
             if use_swa:
                 optimizer.swap_swa_sgd()
 
             # if we do not use dev data for model selection, save final model
             if save_final_model and not param_selection_mode:
                 self.model.save(base_path / "final-model.pt")
+                if self.wandb_logger.wandb_run:
+                    self.wandb_logger.log_artifact(
```
djstrong, line 738: I guess it uploads the model. It should be optional as model can weight many GBs.

AyushExel: Oh yes. I missed this. I'll add another `log_final_model` flag here and lof the model only id its set. thanks for catching this

AyushExel: @djstrong should be fixed now

AyushExel:
@alanakbik This PR only allows users to log models( either on set epoch intervals or the final model). If this looks good, we can implement similar logging in the language model trainer. We could also discuss enabling dataset logger.

AyushExel:
@alanakbik Heads up- I'm closing this for now. Let me know whenever you're ready to discuss further.

mehadi92:
@AyushExel, @djstrong, @alanakbik  any update regarding this PR. I think wandb is a good feature for experiment tracking. And it would be better if flair support this.

alanakbik:
Hello @mehadi92 @AyushExel this is on our roadmap. We're focusing on core features for the 1.0 release now - once the interfaces are stable, we will look into this again.

PR #2267, opened 2021-05-04
fixed typo in property name

schelv:
fixed typo in property name


alanakbik:
oof - thanks for spotting and fixing this!

PR #2271, opened 2021-05-13
Adding PERSON_NER dataset to flair

OatsProduction:


flair/datasets/sequence_labeling.py
```diff
@@ -732,6 +732,64 @@ def __offset_docstarts(file_in: Union[str, Path], file_out: Union[str, Path]):
...
+        # default dataset folder is the cache root
+        if not base_path:
+            base_path = flair.cache_root / "datasets"
+        data_folder = base_path / dataset_name
+
+        # download data if necessary
+        conll_path = "https://github.com/das-sudeshna/genid"
+
+        # download files if not present locally
+        cached_path(f"{conll_path}.conll", Path("datasets") / dataset_name)
```
alanakbik, line 771: You need to download all four files and concatenate them into one before instantiating the corpus

OatsProduction: Added the download of all the files and the concatenating of them.

flair/datasets/sequence_labeling.py
```diff
@@ -732,7 +732,65 @@ def __offset_docstarts(file_in: Union[str, Path], file_out: Union[str, Path]):
...
+        cached_path(f"{conll_path}wiki-g.conll", data_folder / 'raw')
+
+        self.__concatAllFiles(data_folder)
+
+        super(PERSON_NER, self).__init__(
+            data_folder,
+            columns,
+            tag_to_bioes=tag_to_bioes,
+            in_memory=in_memory,
+            train_file='bigFile.conll"'
```
alanakbik, line 783: Throws error because you have an additional " in the name of the file. 

Should be `train_file='bigFile.conll'` instead

alanakbik (changes requested):

flair/datasets/sequence_labeling.py
```diff
@@ -732,7 +732,65 @@ def __offset_docstarts(file_in: Union[str, Path], file_out: Union[str, Path]):
...
+        cached_path(f"{conll_path}ieer-g.conll", data_folder / 'raw')
+        cached_path(f"{conll_path}textbook-g.conll", data_folder / 'raw')
+        cached_path(f"{conll_path}wiki-g.conll", data_folder / 'raw')
+
+        self.__concatAllFiles(data_folder)
+
+        super(PERSON_NER, self).__init__(
+            data_folder,
+            columns,
+            tag_to_bioes=tag_to_bioes,
```
alanakbik, line 781: Also, this line needs to be removed, since it is not in BIO format

flair/datasets/sequence_labeling.py
OatsProduction: Fixed in the new commit.

flair/datasets/sequence_labeling.py
OatsProduction: Fixed in the new commit

alanakbik:
@OatsProduction thanks for adding this!

alanakbik (approved):

PR #2280, opened 2021-05-19
Proposal to allow users to push models to the 🤗 Hub

osanseviero:
**Hi Flair Team! I wanted to propose extending the 🤗 Hub integration by allowing your users to upload their models.**

Your users would benefit by being able to easily share their models and try them out with the Hub widgets. This PR is just a demo to show how it can be done. This allows users to simply share their model with a single line.

```
model = SequenceTagger.load(...)
model.push_to_hub('my-own-ner-english')
```

[Here is an example](https://huggingface.co/osanseviero/flair-ner-english) of a repo created with this method. I kept the changes minimal to showcase how this part of the integration could look like, but we would be happy to discuss further in Slack if you prefer.

As a follow-up, there are two things that could be nice

1. Allow trainer to push to Hub

```
trainer.train('resources/taggers/example-pos',
              learning_rate=0.1,
              mini_batch_size=32,
              max_epochs=150,
              push_to_hub=True)
```

2. Automatically create the model card information from training, such as datasets, metrics, and more!

Let us know what you think :) I'm looking forward for a more complete integration with Flair, the library is awesome!

Cheers!
Omar & 🤗 team

### DIFF
--- a/flair/models/sequence_tagger_model.py
+++ b/flair/models/sequence_tagger_model.py
@@ -1,11 +1,13 @@
 import logging
 import sys
+import os
 
 from pathlib import Path
 from typing import List, Union, Optional, Dict, Tuple
 from warnings import warn
 
 import numpy as np
+import tempfile
 import torch
 import torch.nn
 import torch.nn.functional as F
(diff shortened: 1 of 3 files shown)


osanseviero:
Hi @stefan-it! 

Is this a feature that still interests you or should I close this PR?

Thank you :D

PR #2286, opened 2021-05-27
Log saving of best model instead of printing it

MLDLMFZ:


alanakbik:
@MLDLMFZ thanks for improving this!

PR #2297, opened 2021-06-07
Reading algorithm for semantic roles for the Universal Propositions Bank dataset

Dabendorf:
The code contains the new algorithm which reads in the semantic roles, currently working for the seven UP-datasets in flair (can be extended). It also includes the new Frame-class which are the objects the semantic roles get read into.

### DIFF
--- a/flair/data.py
+++ b/flair/data.py
@@ -594,6 +594,8 @@ class Sentence(DataPoint):
         # some sentences represent a document boundary (but most do not)
         self.is_document_boundary: bool = False
 
+        self.frames : List[Frame] = list()
+
     def get_token(self, token_id: int) -> Token:
         for token in self.tokens:
             if token.idx == token_id:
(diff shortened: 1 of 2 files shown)


alanakbik:
@Dabendorf thanks for adding this! However, when I try: 

```python
corpus = UP_ENGLISH()

sentence = corpus.test[1]
print(sentence)
print(sentence.frames)
```

it doesn't load the frames. I think the problem is that the `is_srl` parameter is not automatically set and passed through in `UD_ENGLISH`, `UD_GERMAN`, etc.

Dabendorf:
Thats a very good question, bad behaviour that I loaded the data directly from my disk.
I am actually running into a new question due to your question.

The Universal Proposition Banks are ColumnCorpus objects, while the integration of my methods works with ColumnDatasets, therefore I cannot directly put that into the __init__ functions of those datasets. Is there a better connection between those two objects? There are derived from FlairDataset respectively Corpus. But if I load them directly from my disk its interchangeable. 

Is there a smart inter-flair-solution for that?

alanakbik:
Yes, the ColumnCorpus internally calls thee ColumnDataset objects for the new splits. So you can extend ColumnCorpus with this parameter and pass it through. Then in UD_ENGLISH etc you call ColumnCorpus with the new parameter.

Dabendorf:
Thank you for that explaination, that has been a nice lecture of object orientation. I have now changed that in both the eight UP-objects (I guess you mean UP instead of UD) and inside the ColumnCorpus. It works on my computer now

PR #2303, opened 2021-06-11
Fix typo.

carlosfranzreb:
The word "label" should be in singular, not plural, as there is only one language label.


alanakbik:
@carlosfranzreb thanks for correcting this!

PR #2304, opened 2021-06-12
Remove space before '.' in sample sentences

juanmirocks:
The extra unnecessary space led to wrongly think that the input sentences must already be tokenized.
Moreover, there were already other sample sentences that did not include that extra space.


alanakbik:
@juanmirocks good point! Thanks for fixing this!

Issue #2311, opened 2021-06-20
SegtokSentenceSplitter Incorrect Sentence Position Attributes


dobbersc:
**Describe the bug**
Flair's implementation of the segtok splitter poorly handles input texts with whitespaces before and between the sentences and produces incorrect `start_pos` and `end_pos` attributes of the split sentences.

**To Reproduce**
The following script runs some test cases, reproduces the bug and illustrates the bug's occurrences. For a more in-depth explanation, please see the "Additional context" section.

Huge thanks to @weyaaron for helping me debug this issue.

```python
from typing import List

from flair.data import Sentence
from flair.tokenization import SegtokSentenceSplitter


def test_sentence_splitter(text: str, test_case_header: str) -> None:
    print(f'{test_case_header:-^30}')
    print(f'{text=}')

    sentences: List[Sentence] = SegtokSentenceSplitter().split(text)

    last_start_pos: int = -1
    for i, sentence in enumerate(sentences, start=1):
        print('----')
        print(f'{sentence=}')
        print(f'\tStart of sentence {i}: {sentence.start_pos}')
        print(f'\tEnd of sentence {i}: {sentence.end_pos}')
        if last_start_pos == sentence.start_pos:
            print(f'\t{last_start_pos=} equals {sentence.start_pos=}. They should be unique!')
        last_start_pos = sentence.start_pos

    print()


if __name__ == '__main__':

    # Test-Case 1: No Error
    input_text: str = 'Frau Schmidt hatte den Namen immer gemocht und nie Gelegenheit gehabt, ihn auszusprechen. Ozelot. Ozelot. Ozelot.'
    test_sentence_splitter(input_text, test_case_header='Test-Case 1: No Error')

    # Test-Case 2: Error
    input_text: str = "      Frau Schmidt hatte den Namen immer gemocht und nie Gelegenheit gehabt, ihn auszusprechen. Ozelot. Ozelot. Ozelot."
    test_sentence_splitter(input_text, test_case_header='Test-Case 2: Error')

    # Test-Case 3: Error
    input_text: str = "Frau Schmidt hatte den Namen immer gemocht und nie Gelegenheit gehabt, ihn auszusprechen.      Ozelot. Ozelot. Ozelot."
    test_sentence_splitter(input_text, test_case_header='Test-Case 3: Error')

    # Test-Case 4: Error
[...]

PR #2312, opened 2021-06-20
GH-2311: Fix SegtokSentenceSplitter Incorrect Sentence Position Attributes

dobbersc:
This PR fixes GH-2311.

My fix requires a sentence lookahead when iterating over the plain sentences straight from the segtok library.
This is probably not the best solution. I appreciate any feedback.

### DIFF
--- a/flair/tokenization.py
+++ b/flair/tokenization.py
@@ -1,10 +1,12 @@
 import logging
 
 from abc import ABC, abstractmethod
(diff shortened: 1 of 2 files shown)


alanakbik:
@dobbersc thanks for fixing this and documenting the issue so well! Great solution too with `stagger` (did not know this one) - makes the code very readable!

PR #2327, opened 2021-07-01
Update url for BioSemantics corpus

mariosaenger:
The corpus has been rehosted.


alanakbik:
@mariosaenger thanks for fixing this!

PR #2340, opened 2021-07-15
Relation extraction for DrugProt [WIP]

ChristophAlt:
- add DrugProt corpus to datasets
- CoNLLUDataset now uses SpanLabels for sentence-level entity annotations


alanakbik:
@ChristophAlt thanks for adding this!

PR #2343, opened 2021-07-19
Parameterize CoNLL-U (Plus) Fields for Token-Level Annotation

dobbersc:
This PR enhances the `CoNLLUCorpus` and `CoNLLUDataset` to accept an optional `token_annotation_fields` parameter. It's used to explicitly specify the annotation fields of the CoNLL-U (Plus) file, which should be transcribed to token-level labels. Per default the LEMMA, UPOS and XPOS will be used. 

#### Example:

Consider the following CoNLL-U Plus snippet contained in a `train.conllup` file:

```
# global.columns = id form upos ner misc
# text = Larry Page and Sergey Brin founded Google.
# relations = 7;7;1;2;founded_by|7;7;4;5;founded_by
1	Larry	PROPN	B-PER	_
2	Page	PROPN	I-PER	_
3	and	CCONJ	O	_
4	Sergey	PROPN	B-PER	_
5	Brin	PROPN	I-PER	_
6	founded	VERB	O	_
7	Google	PROPN	B-ORG	_
8	.	PUNCT	O	SpaceAfter=No
```

We can now load a `CoNLLUDataset` annotated with NER and UPOS tags:

```python
from flair.datasets.conllu import CoNLLUDataset

dataset: CoNLLUDataset = CoNLLUDataset('train.conllup', token_annotation_fields=['upos', 'ner'])
```

Next, we print the sentence:

```python
print(dataset.sentences[0])

>>> Sentence: "Larry Page and Sergey Brin founded Google ."   [− Tokens: 8  − Token-Labels: "Larry <PROPN/B-PER> Page <PROPN/I-PER> and <CCONJ> Sergey <PROPN/B-PER> Brin <PROPN/I-PER> founded <VERB> Google <PROPN/B-ORG> . <PUNCT>"  − Sentence-Labels: {'relation': [founded_by from Google (7) -> Larry Page (1,2) (1.0), founded_by from Google (7) -> Sergey Brin (4,5) (1.0)]}]
```

### DIFF
--- a/flair/datasets/conllu.py
+++ b/flair/datasets/conllu.py
@@ -1,21 +1,18 @@
 import logging
+
 from pathlib import Path
 from typing import List, Union, Optional, Sequence, Dict, Tuple
 
+import conllu
+
 from flair.data import Sentence, Corpus, Token, FlairDataset, Span, RelationLabel
 from flair.datasets.base import find_train_dev_test_files
-import conllu
 
 log = logging.getLogger("flair")
 
-DEFAULT_FIELDS = ("id", "form", "lemma", "upos", "xpos", "feats", "head", "deprel", "deps", "misc")
+DEFAULT_FIELDS: Tuple[str, ...] = ("id", "form", "lemma", "upos", "xpos", "feats", "head", "deprel", "deps", "misc")
 
-DEFAULT_FIELD_PARSERS: Dict[str, conllu._FieldParserType] = dict(
-    conllu.parser.DEFAULT_FIELD_PARSERS,
-    **{
-        "ner": lambda line, i: conllu.parser.parse_nullable_value(line[i]),
-    },
-)
+DEFAULT_TOKEN_ANNOTATION_FIELDS: Tuple[str, ...] = ("lemma", "upos", "xpos")
 
 DEFAULT_METADATA_PARSERS: Dict[str, conllu._MetadataParserType] = dict(
     conllu.parser.DEFAULT_METADATA_PARSERS,
@@ -51,6 +48,7 @@ class CoNLLUCorpus(Corpus):
             dev_file=None,
             in_memory: bool = True,
             fields: Optional[Sequence[str]] = None,
+            token_annotation_fields: Optional[Sequence[str]] = None,
             field_parsers: Optional[Dict[str, conllu._FieldParserType]] = None,
             metadata_parsers: Optional[Dict[str, conllu._MetadataParserType]] = None,
     ):
@@ -62,6 +60,7 @@ class CoNLLUCorpus(Corpus):
         :param test_file: the name of the test file
         :param dev_file: the name of the dev file, if None, dev data is sampled from train
         :param in_memory: If set to True, keeps full dataset in memory, otherwise does disk reads
+        :param token_annotation_fields: A subset of the fields parameter for token level annotations
         :return: a Corpus with annotated train, dev and test data
         """
 
@@ -73,6 +72,7 @@ class CoNLLUCorpus(Corpus):
             train_file,
             in_memory=in_memory,
             fields=fields,
+            token_annotation_fields=token_annotation_fields,
             field_parsers=field_parsers,
             metadata_parsers=metadata_parsers,
         )
@@ -83,6 +83,7 @@ class CoNLLUCorpus(Corpus):
                 test_file,
                 in_memory=in_memory,
                 fields=fields,
+                token_annotation_fields=token_annotation_fields,
                 field_parsers=field_parsers,
                 metadata_parsers=metadata_parsers,
             )
@@ -96,6 +97,7 @@ class CoNLLUCorpus(Corpus):
                 dev_file,
                 in_memory=in_memory,
                 fields=fields,
+                token_annotation_fields=token_annotation_fields,
                 field_parsers=field_parsers,
                 metadata_parsers=metadata_parsers,
             )
@@ -112,6 +114,7 @@ class CoNLLUDataset(FlairDataset):
             path_to_conllu_file: Union[str, Path],
             in_memory: bool = True,
             fields: Optional[Sequence[str]] = None,
+            token_annotation_fields: Optional[Sequence[str]] = None,
             field_parsers: Optional[Dict[str, conllu._FieldParserType]] = None,
             metadata_parsers: Optional[Dict[str, conllu._MetadataParserType]] = None,
     ):
@@ -120,6 +123,7 @@ class CoNLLUDataset(FlairDataset):
 
         :param path_to_conllu_file: Path to the CoNLL-U formatted file
         :param in_memory: If set to True, keeps full dataset in memory, otherwise does disk reads
+        :param token_annotation_fields: A subset of the fields parameter for token level annotations
         """
         if type(path_to_conllu_file) is str:
             path_to_conllu_file = Path(path_to_conllu_file)
@@ -134,7 +138,17 @@ class CoNLLUDataset(FlairDataset):
                 fields = conllu.parser.parse_conllu_plus_fields(file)
 
         self.fields = fields or DEFAULT_FIELDS
-        self.field_parsers = field_parsers or DEFAULT_FIELD_PARSERS
+        self.token_annotation_fields = token_annotation_fields or DEFAULT_TOKEN_ANNOTATION_FIELDS
+
+        augmented_default_field_parsers: Dict[str, conllu._FieldParserType] = dict(
+            conllu.parser.DEFAULT_FIELD_PARSERS,
+            **{
+                field: lambda line_, i: conllu.parser.parse_nullable_value(line_[i])
+                for field in self.token_annotation_fields
+            }
+        )
+
+        self.field_parsers = field_parsers or augmented_default_field_parsers
         self.metadata_parsers = metadata_parsers or DEFAULT_METADATA_PARSERS
 
         self.total_sentence_count: int = 0
@@ -200,11 +214,8 @@ class CoNLLUDataset(FlairDataset):
         for conllu_token in token_list:
             token = Token(conllu_token["form"])
 
-            if "ner" in conllu_token:
-                token.add_label("ner", conllu_token["ner"])
-
-            if "lemma" in conllu_token:
-                token.add_label("lemma", conllu_token["lemma"])
+            for field in self.token_annotation_fields:
+                token.add_label(typename=field, value=str(conllu_token[field]))
 
             if "misc" in conllu_token and conllu_token["misc"] is not None:
                 space_after = conllu_token["misc"].get("SpaceAfter")
--- a/flair/datasets/conllu.py
+++ b/flair/datasets/conllu.py
@@ -26,10 +26,10 @@ DEFAULT_METADATA_PARSERS: Dict[str, conllu._MetadataParserType] = dict(
 
 
 def parse_relation_tuple_list(
-    key: str,
-    value: Optional[str] = None,
-    list_sep: str = "|",
-    value_sep: str = ";",
+        key: str,
+        value: Optional[str] = None,
+        list_sep: str = "|",
+        value_sep: str = ";",
 ) -> Optional[List[Tuple[int, int, int, int, str]]]:
     if value is None:
         return value
@@ -44,15 +44,15 @@ def parse_relation_tuple_list(
 
 class CoNLLUCorpus(Corpus):
     def __init__(
-        self,
-        data_folder: Union[str, Path],
-        train_file=None,
-        test_file=None,
-        dev_file=None,
-        in_memory: bool = True,
-        fields: Optional[Sequence[str]] = None,
-        field_parsers: Optional[Dict[str, conllu._FieldParserType]] = None,
-        metadata_parsers: Optional[Dict[str, conllu._MetadataParserType]] = None,
+            self,
+            data_folder: Union[str, Path],
+            train_file=None,
+            test_file=None,
+            dev_file=None,
+            in_memory: bool = True,
+            fields: Optional[Sequence[str]] = None,
+            field_parsers: Optional[Dict[str, conllu._FieldParserType]] = None,
+            metadata_parsers: Optional[Dict[str, conllu._MetadataParserType]] = None,
     ):
         """
         Instantiates a Corpus from CoNLL-U (Plus) column-formatted task data
@@ -108,12 +108,12 @@ class CoNLLUCorpus(Corpus):
 
 class CoNLLUDataset(FlairDataset):
     def __init__(
-        self,
-        path_to_conllu_file: Union[str, Path],
-        in_memory: bool = True,
-        fields: Optional[Sequence[str]] = None,
-        field_parsers: Optional[Dict[str, conllu._FieldParserType]] = None,
-        metadata_parsers: Optional[Dict[str, conllu._MetadataParserType]] = None,
+            self,
+            path_to_conllu_file: Union[str, Path],
+            in_memory: bool = True,
+            fields: Optional[Sequence[str]] = None,
+            field_parsers: Optional[Dict[str, conllu._FieldParserType]] = None,
+            metadata_parsers: Optional[Dict[str, conllu._MetadataParserType]] = None,
     ):
         """
         Instantiates a column dataset in CoNLL-U (Plus) format.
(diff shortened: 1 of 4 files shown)


alanakbik:
The ID and form fields are always mandatory I think, so it is a bit counterintuitive that you always need to specify them as fields. 

Currently this works: 

```python
corpus = CoNLLUCorpus("tests/resources/tasks/conllu",
                      train_file="train.conllup",
                      fields=["id", "form", "ner", "misc"],
                      )
```

but not this: 

```python
corpus = CoNLLUCorpus("tests/resources/tasks/conllu",
                      train_file="train.conllup",
                      fields=['ner', 'misc']
                      )
```

The latter is a bit more succinct as you only define the additional fields you're interested in.

ChristophAlt:
@dobbersc: Looks good! :+1: 
One thing I want to put up for discussion: What is the desired behavior when ``token_annotation_fields`` is None? Currently, only "lemma", "upos", and "xpos" are considered token annotation fields. I'd argue that in this case we should consider all fields to be token annotations except "id", "form", and "misc" to be more flexible.

@alanakbik: I think it depends on how we intend to use CoNLLUCorpus / CoNLLUDataset -- as an abstract base class for CoNLLU-based corpora or a class intended for direct use. The ``fields`` argument specifies the name of each field in the file, i.e. the column headers. If none is given, we try to infer it from the file by looking for ``global.columns`` as defined in the CoNLLU-P specification and otherwise fall back to the default columns (universal dependencies). This is not intended to select from the set of available fields (if I understood your comment correctly). If we enforce the mandatory "id" and "form" columns, we may prevent datasets that derive from CoNLLUDataset to deviate from the CoNLLU format.

dobbersc:
> 
> 
> The ID and form fields are always mandatory I think, so it is a bit counterintuitive that you always need to specify them as fields.
> 
> Currently this works:
> 
> ```python
> corpus = CoNLLUCorpus("tests/resources/tasks/conllu",
>                       train_file="train.conllup",
>                       fields=["id", "form", "ner", "misc"],
>                       )
> ```
> 
> but not this:
> 
> ```python
> corpus = CoNLLUCorpus("tests/resources/tasks/conllu",
>                       train_file="train.conllup",
>                       fields=['ner', 'misc']
>                       )
> ```
> 
> The latter is a bit more succinct as you only define the additional fields you're interested in.

@alanakbik: Added to the comment of @ChristophAlt, I think it depends whether the user passes a CoNLL-U or a CoNNL-U Plus file. The [CoNNL-U Plus documentation](https://universaldependencies.org/ext-format.html) states 

> There are ten predefined column names that identify the original columns of the core CoNLL-U files (see specification). A CoNLL-U Plus file may contain any subset of them (or none of them, although for stand-off annotation of UD files, at least the ID column is typically needed).

Therefore, always forcing the parsing of the ID and FORM fields may lead to problems for such edge-cases.

In general, if the user passes a CoNLL-U Plus file, the fields to be parsed are taken from the `global.columns`. If the user passes a conform CoNLL-U file, the fallback default fields work fine. With this, the only scenario where the user has to specify the fields parameter is when a passed CoNLL-U (not Plus) file deviates from the official format. 

Moreover, did you use the same `train.conllup` file as provided by the PR description? Actually, both examples should fail, since the default `token_annotation_fields` are not a subset of the specified `fields`.

dobbersc:
> 
> 
> @dobbersc: Looks good! 👍
> One thing I want to put up for discussion: What is the desired behavior when `token_annotation_fields` is None? Currently, only "lemma", "upos", and "xpos" are considered token annotation fields. I'd argue that in this case we should consider all fields to be token annotations except "id", "form", and "misc" to be more flexible.

Then the default token annotations would be LEMMA, UPOS, XPOS, FEATS, HEAD, DEPREL and DEPS. I see your point that including these fields is more flexible long term. 

The problem is that currently, all the specified fields of the `token_field_annotations` parameter are parsed with the `parse_nullable_value` function. But the FEATS, HEAD, DEPREL and DEPS fields need to be parsed differently. Since these fields are part of the default field parser of the `conllu` package, my suggestion is that we do not overwrite the default field parsers' values from the `token_field_annotations` parameter. 

As an example, the FEATS field is parsed as a dictionary value. This raises the question of how we want to encode the parsed information into the token's labels.

An alternative would be to save the annotation as plaintext even for the "special" fields. The drawback of this approach would be that the user has to do the rest of the parsing themselves.

alanakbik:
Ah I think I got confused since now there is `fields` and `token_annotation_fields` and still am a bit confused. So `fields` is an optional parameter that defines the names of each column in case global names aren't defined. And `token_annotation_fields` then tells us which ones to add to the `Sentence`?

Generally, we want FEATs to be handled correctly out-of-the-box so they shouldn't be stored as plain text.

dobbersc:
> Ah I think I got confused since now there is `fields` and `token_annotation_fields` and still am a bit confused. So `fields` is an optional parameter that defines the names of each column in case global names aren't defined. And `token_annotation_fields` then tells us which ones to add to the `Sentence`?

Yes. 
- The `fields` parameter defines the fields of the CoNLL file which will be parsed. If it's `None` the `global.columns` will be used if available. If not we fall back to the default CoNNL-U fields. 
- The `token_annotation_fields` parameter is a subset of the `fields` parameter and defines which fields should serve as an annotation for the sentence's tokens.

I get that its kind of confusing having so many parameters related to the "fields". I thought about combining the two but that really restricts the parser's flexibility. Maybe more descriptive names would help?

> Generally, we want FEATs to be handled correctly out-of-the-box so they shouldn't be stored as plain text.

I agree. Consider a token with annotated FEATS field as `Case=Nom|Gender=Fem|Number=Sing|Polarity=Pos`. The connlu-parser is able to parse this as the dictionary `{'Case': 'Nom', 'Gender': 'Fem', 'Number': 'Sing', 'Polarity': 'Pos'}`. Then we have the problem that we cannot attach a value any other than a string to a token label.

ChristophAlt:
If a field value is a dictionary (or mapping in general) its key-value pairs should be added to the token as the label name and label value. That would be consistent with the behavior of [UniversalDependenciesDataset (line 197)](https://github.com/flairNLP/flair/blob/5e372d1d9b15da7c1d3ab9d8ae4292a16d8e56a9/flair/datasets/treebanks.py#L197), we add the features as labels to the respective token.

dobbersc:
I'm not sure whether it is desired but another possibility would be to relax the types that a `Label.value` can take. Currently, the label's values are restricted to strings. I assume this is to support the BIO tagging format ([related function](https://github.com/flairNLP/flair/blob/c298a35bb0972c4f91fe6ff9be121242e1dcd10f/flair/data.py#L714)). If we allow all types as a label value, when constructing the spans we may assume that tokens with `Label.value` types other than `str` form their own span. I.e. tokens with such labels form a singleton span.

With this, we can use the parsed dictionary directly instead of the `key=value` string notation. Then the user can access the tokens label's values more conveniently and reuse them in their programming logic.

I'm not sure if this idea is reasonable but since there's a huge refactoring going on I might as well ask.

dobbersc:
I've thought about this again and implemented the changes suggested by @ChristophAlt. I don't know why I had a problem with this before, my mind was kind of blocked. Now the `default_field_annotations` have been extended to 
> `("lemma", "upos", "xpos", "feats", "head", "deprel")`.

The problem with the `deps` field is that it does not contain key-value annotations. In fact, it is an extended version of the `deprel` field with lists of heads and relation annotations. Here, I'm not sure how to convert this field to token annotations. Consider this example:

image

With the `deprel` field, the first token has the head "2" and the relation "nsubj". Now, the `deps` field has two different heads and relation annotations. Here the implementation problem is: How to distinguish between both heads and how to preserve the option to match the `deps` heads with their original intended relation? Without a new label type that supports pair annotations, I don't have a solution right now. This is the reason I've implemented the relation annotations on a sentence level, using the relation label. In my opinion, this is a nice additional feature.

Usage example:

```
# text = They buy and sell books.
1    They     they    PRON    PRP    Case=Nom|Number=Plur               2    nsubj    2:nsubj|4:nsubj
2    buy      buy     VERB    VBP    Number=Plur|Person=3|Tense=Pres    0    root     0:root
3    and      and     CONJ    CC     _                                  4    cc       4:cc
4    sell     sell    VERB    VBP    Number=Plur|Person=3|Tense=Pres    2    conj     0:root|2:conj
5    books    book    NOUN    NNS    Number=Plur                        2    obj      2:obj|4:obj
6    .        .       PUNCT   .      _                                  2    punct    2:punct
```

```python
from flair.datasets.conllu import CoNLLUDataset

dataset: CoNLLUDataset = CoNLLUDataset('universal_dependencies.conllu')  # The default token annotation fields are sufficient.

print(dataset[0].get_labels("deps"))
```

Shortened output: `[nsubj from buy (2) -> They (1) (1.0), nsubj from sell (4) -> They (1) (1.0), ...]`

flair/data.py
```diff
@@ -1062,7 +1062,7 @@ def is_context_set(self) -> bool:
     def get_labels(self, label_type: str = None):
 
         # TODO: crude hack - replace with something better
-        if label_type:
+        if label_type and label_type != "deprel":
```
dobbersc, line 1065: Without this change the `get_labels` function fails:

```python
from flair.datasets.conllu import CoNLLUDataset

dataset: CoNLLUDataset = CoNLLUDataset('universal_dependencies.conllu')  # same example file as in the comment above

print(dataset[0].get_labels("deprel"))
```

```
Traceback (most recent call last):
  File "<input>", line 1, in <module>
  File "/vol/fob-vol7/mi18/dobbersc/PyCharmProjects/flair_fork/flair/data.py", line 1068, in get_labels
    self.add_complex_label(label_type, label=SpanLabel(span, span.tag, span.score))
  File "/vol/fob-vol7/mi18/dobbersc/PyCharmProjects/flair_fork/flair/data.py", line 300, in add_complex_label
    if typename in self.annotation_layers and label in self.annotation_layers[typename]:
  File "/vol/fob-vol7/mi18/dobbersc/PyCharmProjects/flair_fork/flair/data.py", line 257, in __eq__
    and self.head.id_text == other.head.id_text \
AttributeError: 'SpanLabel' object has no attribute 'head'
```

My guess is that inside the `add_complex_label` function the created `deprel` relation labels are incorrectly overwritten with span labels. The reason behind this is that the tokens have an individual `deprel` annotation in addition to the sentence labels.

I'm not sure how to fix this properly, since I'm not familiar with the "crude hack" that is mentioned here. I would appreciate it if you, @alanakbik, could take a look at this.

alanakbik: @dobbersc I think recording deprels as relation labels is a great idea! But the refactoring involving complex labels is not yet complete and is currently blocked by some other refactorings that need to happen first. Since we currently do not yet predict this type of extended dependency trees in Flair anyway, I would wait with recording the deps until the rest is sorted out.

alanakbik:
I tested the PR using the English UP and got an error, plus an inconsistency. It involves this sentence: 

```
# sent_id = answers-20111108072305AAPJTjj_ans-0005
# text = It's more compact, ISO 6400 capability (SX40 only 3200), faster lens at f/2 and the SX40 only f/2.7.
1-2	It's	_	_	_	_	_	_	_	_
1	It	it	PRON	PRP	Case=Nom|Gender=Neut|Number=Sing|Person=3|PronType=Prs	4	nsubj	4:nsubj	_
2	's	be	AUX	VBZ	Mood=Ind|Number=Sing|Person=3|Tense=Pres|VerbForm=Fin	4	cop	4:cop	_
3	more	more	ADV	RBR	_	4	advmod	4:advmod	_
4	compact	compact	ADJ	JJ	Degree=Pos	0	root	0:root	SpaceAfter=No
5	,	,	PUNCT	,	_	8	punct	8:punct	_
6	ISO	iso	NOUN	NN	Number=Sing	8	compound	8:compound	_
7	6400	6400	NUM	CD	NumType=Card	6	nummod	6:nummod	_
8	capability	capability	NOUN	NN	Number=Sing	4	list	4:list	_
9	(	(	PUNCT	-LRB-	_	10	punct	10:punct|10.1:punct	SpaceAfter=No
10	SX40	SX40	PROPN	NNP	Number=Sing	8	parataxis	8:parataxis|10.1:nsubj	_
10.1	has	have	VERB	VBZ	_	_	_	8:parataxis	CopyOf=-1
11	only	only	ADV	RB	_	12	advmod	12:advmod	_
12	3200	3200	NUM	CD	NumType=Card	10	orphan	10.1:obj	SpaceAfter=No
13	)	)	PUNCT	-RRB-	_	10	punct	10:punct|10.1:punct	SpaceAfter=No
14	,	,	PUNCT	,	_	8	punct	8:punct	_
15	faster	fast	ADJ	JJR	Degree=Cmp	16	amod	16:amod	_
16	lens	lens	NOUN	NN	Number=Sing	4	list	4:list	_
17	at	at	ADP	IN	_	18	case	18:case	_
18	f/2	f/2	NOUN	NN	Number=Sing	16	nmod	16:nmod:at	_
19	and	and	CCONJ	CC	_	21	cc	21:cc|21.1:cc	_
20	the	the	DET	DT	Definite=Def|PronType=Art	21	det	21:det	_
21	SX40	SX40	PROPN	NNP	Number=Sing	16	conj	16:conj:and|21.1:nsubj	_
21.1	has	have	VERB	VBZ	_	_	_	16:conj:and	CopyOf=-1
22	only	only	ADJ	JJ	Degree=Pos	23	amod	23:amod	_
23	f	f	NOUN	NN	Number=Sing	21	orphan	21.1:obj	SpaceAfter=No
24	/	/	PUNCT	,	_	23	punct	23:punct	SpaceAfter=No
25	2.7	2.7	NUM	CD	NumType=Card	23	nummod	23:nummod	SpaceAfter=No
26	.	.	PUNCT	.	_	4	punct	4:punct	_
```

It throws an error trying to parse the `deps` field. 

But even when commenting out the deps-logic, I get the following sentence: 

~~~
Sentence: "It's It 's more compact , ISO 6400 capability ( SX40 has only 3200 ) , faster lens at f/2 and the SX40 has only f / 2.7 ."   [− Tokens: 29  − Token-Labels: "It's <None/None/None/None/None/None> It <it/PRON/PRP/Nom/Neut/Sing/3/Prs/4/nsubj> 's <be/AUX/VBZ/Ind/Sing/3/Pres/Fin/4/cop> more <more/ADV/RBR/None/4/advmod> compact <compact/ADJ/JJ/Pos/0/root> , <,/PUNCT/,/None/8/punct> ISO <iso/NOUN/NN/Sing/8/compound> 6400 <6400/NUM/CD/Card/6/nummod> capability <capability/NOUN/NN/Sing/4/list> ( <(/PUNCT/-LRB-/None/10/punct> SX40 <SX40/PROPN/NNP/Sing/8/parataxis> has <have/VERB/VBZ/None/None/None> only <only/ADV/RB/None/12/advmod> 3200 <3200/NUM/CD/Card/10/orphan> ) <)/PUNCT/-RRB-/None/10/punct> , <,/PUNCT/,/None/8/punct> faster <fast/ADJ/JJR/Cmp/16/amod> lens <lens/NOUN/NN/Sing/4/list> at <at/ADP/IN/None/18/case> f/2 <f/2/NOUN/NN/Sing/16/nmod> and <and/CCONJ/CC/None/21/cc> the <the/DET/DT/Def/Art/21/det> SX40 <SX40/PROPN/NNP/Sing/16/conj> has <have/VERB/VBZ/None/None/None> only <only/ADJ/JJ/Pos/23/amod> f <f/NOUN/NN/Sing/21/orphan> / <//PUNCT/,/None/23/punct> 2.7 <2.7/NUM/CD/Card/23/nummod> . <./PUNCT/./None/4/punct>"  
~~~

Note how the sentence now begins with "**It's It 's** more compact , ISO 6400 capability ( SX40 has only 3200 ) , faster lens at f/2 and the SX40 has only f / 2.7 .", i.e. the duplication of the word "it's". 

This is a particularity of UD formatting that words  like "it's" may get expanded, and this is something which we already handle in the `UniversalDependencyDataset` object (see #1981).

Generally, I see an expanding problem of redundancy in Flair datasets that deal with column data, since we now have three/four objects: 
- the `ColumnDataset` as the standard dataset we use for NER and UP tasks
- the `HunerDataset` class (subclass of `ColumnDataset`) that we use for all biomedical NER tasks
- the `UniversalDependenciesDataset` that we use for all UD data
- the `CoNLLUDataset` that we use for relation extraction datasets (and I guess now also UD?)

I'd say that's too many top-level classes for essentially the same thing. Ideally, they should be refactored into a single top-level class from which other classes perhaps inherit, like `HunerDataset` inherits from `ColumnDataset` and just adds a few extra things. 

To move forward, I'd suggest removing the handling of deps from the `CoNLLUDataset` for now, so that we can merge the PR, and use this object only for relation extraction datasets for now. I would then begin a larger refactoring to try to merge all four classes above into one.

dobbersc:
I agree. Due to the redundancies, the different datasets need to be refactored. The parsing with the conllu module and the handling of the multi-word tokens need some work. Then this is a task for future refactorings. 

With regard to the current use case, I've removed the "complex label annotation" of the `deps` and `deprel` fields and adjusted the tests accordingly. I've also added a docstring to reflect the yet unsupported formats.

flair/data.py
dobbersc: Thank you for the review. I've implemented the requested changes and hope that we can include the deps relation labels after the refactorings again.

PR #2349, opened 2021-07-23
Fix error with whitespaces at the end of last token

djstrong:
Test:
```
import torch

from flair.data import Token, Sentence

ff = FlairEmbeddings('news-forward-fast', tokenized_lm=False)
s1 = Sentence()
s1.add_token(Token('1 '))
ff.embed(s1)
```

### DIFF
--- a/flair/data.py
+++ b/flair/data.py
@@ -477,7 +477,7 @@ class Span(DataPoint):
             plain += token.text
             if token.whitespace_after:
                 plain += " "
-        return plain.rstrip()
+        return plain
 
     def to_dict(self):
         return {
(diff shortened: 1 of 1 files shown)


djstrong:
The problem is with SegtokTokenizer, which wronlgy set `whitespace_after` for last token.
```
from flair.tokenization import SegtokTokenizer
tokens = SegtokTokenizer.run_tokenize("I love Berlin.")
for token in tokens:
  print(token, token.whitespace_after)

Token: I True
Token: love True
Token: Berlin False
Token: . True
```

PR #2363, opened 2021-08-03
Added text classification datasets from Glue benchmark 

lukasgarbas:
Following datasets from GLUE were added:

- Text classification: CoLA and SST2 (which already exists as SENTEVAL_SST_BINARY so I just checked that train/dev/test splits are kept in original form and not sampled)
- Text-pair classification: MNLI, QNLI, QQP, MRPC, WNLI

The only missing dataset is Semantic Textual Similarity STS-B, which is a text-pair regression task and would probably require a TextPairRegressor model.

### DIFF
--- a/flair/datasets/text_text.py
+++ b/flair/datasets/text_text.py
@@ -590,10 +590,10 @@ class GLUE_MNLI(DataPairCorpus):
                       str(data_folder / "MNLI/eval_dataset_mismatched.tsv"))
 
         matched_suffix = "matched" if evaluate_on_matched else "mismatched"
-        
-        dev_dataset = "MNLI/dev_" + matched_suffix + ".tsv"
-        eval_dataset = "MNLI/eval_dataset_" + matched_suffix + ".tsv"
-        
+
+        dev_dataset = "dev_" + matched_suffix + ".tsv"
+        eval_dataset = "eval_dataset_" + matched_suffix + ".tsv"
+
         self.evaluate_on_matched = evaluate_on_matched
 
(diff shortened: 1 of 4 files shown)


flair/datasets/text_text.py
```diff
@@ -520,6 +520,476 @@ def tsv_from_eval_dataset(self, folder_path: Union[str, Path]):
...
+
+            # rename test file to eval_dataset, since it has no labels
+            os.rename(str(data_folder / "MNLI/test_matched.tsv"),
+                      str(data_folder / "MNLI/eval_dataset_matched.tsv"))
+            os.rename(str(data_folder / "MNLI/test_mismatched.tsv"),
+                      str(data_folder / "MNLI/eval_dataset_mismatched.tsv"))
+
+        matched_suffix = "matched" if evaluate_on_matched else "mismatched"
+        
+        dev_dataset = "MNLI/dev_" + matched_suffix + ".tsv"
```
marcelmmm, line 594: I think there is aproblem with the path here. I get an assertion error when first initializing the corpus. Maybe one 'MNLI' too much.

marcelmmm (commented):
Looks good :+1:

flair/datasets/text_text.py
lukasgarbas: Indeed! Thanks for pointing that out. I fixed the path for MNLI and trained the small electra model with it. MNLI seems to work now : )

alanakbik:
@lukasgarbas thanks for adding this!

PR #2366, opened 2021-08-10
Add per-label thresholds

### LINKED ISSUE #2274: Allow per-label threshold for multi-label classifiers
Currently there is a single threshold for multi-label classifiers, meaning that the same threshold is applied to all labels. It is sometimes beneficial to have different thresholds for different labels. 

Model class has `multi_label_threshold` member. Currently it's a float. We could make it dictionary:
```
model.multi_label_threshold = {'label_1': label_1_threshold, ..., 'default': default_threshold}
```
so when the label is explicitly mentioned in the dictionary, the corresponding threshold is applied, otherwise `default_threshold` is applied.

yosipk:
Closes #2274

### DIFF
--- a/flair/nn/model.py
+++ b/flair/nn/model.py
@@ -385,8 +385,8 @@ class DefaultClassifier(Classifier):
     def multi_label_threshold(self):
         return self._multi_label_threshold
 
-    @multi_label_threshold.setter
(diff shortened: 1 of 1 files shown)


alanakbik:
@yosipk thanks for adding this - very helpful!

PR #2375, opened 2021-08-17
Entity linking

marcelmmm:
EntityLinker and Entity Linking Datasets


alanakbik:
I think the dependency 'wikipediaapi' needs to be added to requirements.txt

alanakbik:
I've added the missing requirements and also changed the default sentence splitters to Sektok since SpaCy requires additional modules to be installed that don't come with default Flair. In the `AQUAINT_EL` and `IITB_EL` corpora, you can now pass the `sentence_splitter` as an argument in case you want to use SpaCy instead.

However, unfortunately the change to SegTok broke `AQUAINT_EL` since it now sometimes splits a sentence at the wrong place and breaks up an entity. You can reproduce this error with: 

```python
corpus = AQUAINT_EL()
```

Could you take a look?

marcelmmm:
Thank you for all this!!
I changed the AQUAINT_EL corpus. Now it works when using SegTok as well.

alanakbik:
@marcelmmm thanks a lot for adding this!

PR #2391, opened 2021-08-23
Allow lower-cased Flair Embeddings

stefan-it:
Hi,

this PR introduces a new `is_lower` option for the Flair Embeddings class, that allows to use Flair Embeddings, that are trained on lower-cased text.

This option is e.g. needed for the CLEF HIPE Flair Embeddings (that were trained on lower-cased historic corpora):

```python

embeddings_historic = FlairEmbeddings("de-impresso-hipe-v1-forward")
embeddings_historic.is_lower  # returns True

embeddings_german = FlairEmbeddings("de-forward")
embeddings_german.is_lower  # returns False
```


alanakbik (approved):
Thanks for adding this!

PR #2392, opened 2021-08-23
add dict for unknown labels to model.evaluate and trainer.train and fix error in AIDA_CONLL_EL corpus

marcelmmm:


alanakbik:

PR #2403, opened 2021-08-28
Introduce new Flair Examples Folder

stefan-it:
Hi,

this PR introduces a new examples folder, where scripts for various downstream tasks (such as NER) will be located in near future :rocket: 

For this PR an example script for fine-tuning models on NER datasets is added. It also shows how to use the recently introduced FLERT approach for NER.

The script itself uses the awesome Hugging Face Argument Parser wrapper. This allows us to easily define dataclasses for commandline options.


stefan-it:
As briefly discusses with @alanakbik , this should be merged after 0.9 release (so marking it as draft here for now).

examples/ner/run_ner.py
```diff
@@ -0,0 +1,183 @@
...
+    set_seed(training_args.seed)
+
+    flair.device = training_args.device
+
+    corpus = get_flair_corpus(data_args)
+
+    logger.info(corpus)
+
+    tag_type: str = "ner"
+    tag_dictionary = corpus.make_tag_dictionary(tag_type=tag_type)
```
stefan-it, line 143: Oh, it seems that this method is outdated:

```bash
run_ner.py:143: DeprecationWarning: Call to deprecated method make_tag_dictionary. (Use 'make_label_dictionary' instead.) -- Deprecated since version 0.8.
```

stefan-it:
@alanakbik I could also add an example for PoS Tagging so that we have at least two working examples for the first start :)

alanakbik:
@stefan-it thanks a lot for adding this, many people will surely find this useful! We'll add more examples in future PRs!

Issue #2396, opened 2021-08-25
Support for linear scheduler with warmup (finetune like BERT does)


ciaochiaociao:
Hello,

I would like to run some experiments based on transformer-like architecture, e.g., BERT + Linear + CRF, and **finetune** the BERT layer. 

The common practice for finetune it is using AdamW as optimizer and a learning rate scheduler with warmup at first and linear decay afterwards.

I know in trainer, it can customize optimizer like

```
from torch.optim import AdamW

trainer = ModelTrainer(...,optimizer=AdamW)  # takes as input a class

trainer.train(...)
```

But when it comes to customize **scheduler**, things get weird. For example, using huggingface's `transformers` package to get a linear scheduler with warmup:

```
from transformers import get_linear_schedule_with_warmup
scheduler = get_linear_schedule_with_warmup(optimizer=AdamW(...))  # optimizer takes as input an instance and this function returns a scheduler INSTANCE
```

Note one has to **instantiate** an optimizer beforehand.

And then to customize scheduler during training
```
trainer.train(scheduler=scheduler)  # this seems not to work for it only consumes a class instead of an instance
```

So to make this work, maybe we can make 
 1. `ModelTrainer` accept **optimizer object** AND `trainer.train()` accept **scheduler object** OR
 2. Create a `LinearSchedulerWithWarmup` class, so it can be input to `.train()`

Do you plan to do this or have any suggestions?

alanakbik:
@ciaochiaociao good point and something we should really add to Flair! @lukasgarbas can you take a look at this?

PR #2415, opened 2021-09-07
Linear scheduler with warmup

lukasgarbas:
Hi,

as suggested in [#2396](https://github.com/flairNLP/flair/issues/2396), a linear learning rate schedule can work nicely when tuning transformer models. One way to add this is to create a `LinearSchedulerWithWarmup` class that can be passed to `trainer.train()` together with the number of warmup steps as a fraction (`warmup_fraction=0.1` by default). The learning rate linearly increases from zero during warmup and decays for the rest of the training. Here’s an example:

```python
from flair.optim import LinearSchedulerWithWarmup

trainer = ModelTrainer(...,optimizer=AdamW)

trainer.train(learning_rate=2e-5,
              mini_batch_size=16,
              max_epochs=10,
              scheduler=LinearSchedulerWithWarmup,
              warmup_fraction=0.1) # warmup for 0.1 of total train steps (1 epoch) 
```

Note on using other schedulers than AnnealOnPlateau:
- Current logging in the trainer prints ‘patience’, ‘anneal_factor’, and 'bad epochs' that are not used by other lr schedulers. I guess the only way to change it is to adjust the trainer.

Let me know what do you think. I'm also open to other suggestions on how to add this : )

### DIFF
--- a/flair/optim.py
+++ b/flair/optim.py
@@ -4,7 +4,7 @@ import math
 import torch
 from torch.optim import Optimizer
 from torch.optim.optimizer import required
-from torch.optim.lr_scheduler import _LRScheduler, ReduceLROnPlateau
+from torch.optim.lr_scheduler import _LRScheduler, ReduceLROnPlateau, LambdaLR
 
 
 log = logging.getLogger("flair")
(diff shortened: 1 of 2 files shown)


ciaochiaociao:
Thank you for a prompt support for this feature! 

I will also leave some of my thoughts here:
The other repositories like `huggingface` actually implements it with the API of `Trainer.train(optimizers=(optimizer_instance, scheduler_instance))`, taking as input the optimizer instance and scheduler instance, which I believe has the following benefits:
 a. allowing for users to use their own customized optimizer and scheduler, which normally can be made by `LambdaLR` from `torch.optim`. (see https://huggingface.co/transformers/main_classes/trainer.html#transformers.Trainer)
 b. allowing for saving the current state of the optimizer and scheduler and making easier recovering their states back to where they are finished/interrupted. This helps when one wants to continue previously terminated training like `trainer.train(resume_from_checkpoint='folder_containing_model_optimizer_and_scheduler')` as in https://huggingface.co/transformers/main_classes/trainer.html#transformers.Trainer.train

P.S. Currently in flair, to resume training from flair, the tutorial has it in https://github.com/flairNLP/flair/blob/master/resources/docs/TUTORIAL_7_TRAINING_A_MODEL.md#resuming-training, where the code is https://github.com/flairNLP/flair/blob/c5bed58b024ebc2c323d405882ef5ad6c465e651/flair/trainers/trainer.py#L759. Flair only loads the model state but not for scheduler and optimizer.

alanakbik:
@lukasgarbas thanks for adding this - works well!

@ciaochiaociao good points - ideally, I'd like to find an abstraction such that for most users, a default fine-tuning setup is used without needing to do many modifications. For instance, next to the current `trainer.train()` which does SGD with learning_rate=0.1, 150 epochs and annealing against dev score (i.e. good default values for standard training), I've been thinking of adding a `trainer.fine_tune()` that uses AdamW, linear scheduler with warmup, a small learning rate and only 10 epochs by default (good default values for fine-tuning transformers). Like it currently is in `trainer.train()`, users can overwrite pre-set variables like setting a different mini-batch size or learning rate. Then, for "power users" that would define their own schedulers and optimizers, we'd need to add a different way perhaps like you suggest. 

I'll merge this PR for now but will keep this in mind as we work on the interfaces for the next version!

PR #2421, opened 2021-09-10
make it optional to create logfiles and loss files

helpmefindaname:
When using cloud services (aws, gcloud, azure, ...) the logs get saved automatically in their respective logging service.
Hence, creating a logfile is redundant and only (very slightly) slows down training speed or increases storage costs. Therefore, I'd like to add an option to disable the creation of such files.

This also fixes small logging mistakes, like not logging the momentum when using cyclic momentum with adam or using print statements instead of logging properly


flair/trainers/trainer.py
```diff
@@ -227,7 +230,10 @@ def train(
         if type(base_path) is str:
             base_path = Path(base_path)
 
-        log_handler = add_file_handler(log, base_path / "training.log")
+        if create_file_logs:
```
alanakbik, line 233: This is problematic. If you set create_file_logs=False and create_loss_file=False and use a new folder, the execution crashes since the folder is never created. 

You need to add the following snippet above to make sure the folder is always created: 


```python
        # create folder if it doesn't exist
        if not os.path.exists(base_path):
            os.makedirs(base_path)
```

alanakbik (changes requested):

flair/trainers/trainer.py
```diff
@@ -227,7 +230,10 @@ def train(
         if type(base_path) is str:
             base_path = Path(base_path)
```
alanakbik, line 232:
```suggestion
        # create folder if it doesn't exist
        if not os.path.exists(base_path):
            os.makedirs(base_path)
```

flair/trainers/trainer.py
helpmefindaname: I see, I didn't test it with non-existing folders, now it should work, thanks

alanakbik (approved):
Thanks!

alanakbik:

Issue #2434, opened 2021-09-14
start_pos and end_pos attributes print 'None'


rohitjajee:
**Describe the bug**
I am trying to get the start position and end position of the entities in the craft dataset, but the start_pos and end_pos atrribites of the entity object print “None”. 

**To Reproduce**
import flair 
from flair.datasets import biomedical 
craft = biomedical.CRAFT_V4() 

for sentence  in craft.get_all_sentences(): 
	for entity in sentence.get_spans(): 
		print(entity.text, entity.labels, entity.start_pos, entity.end_pos) 

**Expected behavior**
It should the start position and end positions of the entity

**Output**
DRG [uberon (1.0)] None None  
Neurons [cl (1.0)] None None  
Pea3 [pr (1.0)] None None  
dorsal root ganglion [uberon (1.0)] None None  
DRG [uberon (1.0)] None None 
. 
. 

**Environment (please complete the following information):**
 - OS [e.g. iOS, Linux]:
 - Version [e.g. flair-0.3.2]:

**Additional context**

PR #2440, opened 2021-09-17
set token positions on column corpus

helpmefindaname:
This is supposed to fix https://github.com/flairNLP/flair/issues/2434
besides that, it removes the whitespacing for japanese tokenisation, as there is no space between words


helpmefindaname:
status update:
this pr is back to a mergeable state/ready for review.

alanakbik:
@helpmefindaname thanks for fixing this! And sorry for taking so long to merge!

PR #2462, opened 2021-10-03
GH-1529 fix infinite loop in Span::to_original_text

mhoyer:
With https://github.com/flairNLP/flair/issues/1529 there was already a fix for a potential endless loop in the `Sentence` class. Now we also got trapped in the same situation, but within `Span` class this time.


alanakbik:
@mhoyer thanks for fixing this!

PR #2486, opened 2021-10-22
Dependency parser

73minerva:
This PR adds dependency parser model based on [Deep biaffine attention for neural dependency parsing](https://arxiv.org/pdf/1611.01734.pdf). The main model in paper uses a static word embedding and POS tags vector as input, while it only uses StackedEmbeddings. 
The first trained model can be downloaded from [here](https://drive.google.com/file/d/1knTksQjnflvpWz-EC7St_7ox99GG5AIJ/view?usp=sharing). It was trained with Glove and Flair embeddings on flair.datasets.UD_ENGLISH dataset. If I understand correctly, the UD_ENGLISH dataset in flair library, only contains the "EWT" version, not the other ones mentioned in [this](https://universaldependencies.org). In the end, the following results were obtained:
**UAS : 0.9032 - LAS : 0.9243**
In the near future, I will train the model on PTB dataset to have a comparison with paper results.

For better printing of parsed sentences, I add the "print_tree" param to predict method:
```
sentence = Sentence('I prefer the morning flight through Denver.')
dep_parser_model: DependencyParser = DependencyParser.load(path_to_model)
dep_parser_model.predict(sentence, print_tree=True)
```
This should print something like this:
image

It is not done yet and needs some refactoring. Until then, it will be great if you share your comments and suggestions.

### DIFF
--- a/flair/models/dependency_parser_model.py
+++ b/flair/models/dependency_parser_model.py
@@ -183,3 +183,42 @@ class DependencyParser(flair.nn.Model):
             rel_loss += self.loss_function(score_relation, rel_labels)
 
         return arc_loss, rel_loss
+    
+    def predict(self,
+                sentences: Union[List[Sentence], Sentence],
+                mini_batch_size: int = 32,
+                num_workers: int = 8,
+                print_tree: bool = False,
+                embedding_storage_mode="none",
+                ) -> None:
+        """
+        Predict arcs and tags for Dependency Parser task
+        :param sentences: a Sentence or a List of Sentence
+        :param mini_batch_size: mini batch size to use
+        :param print_tree: set to True to print dependency parser of sentence as tree shape
+        :param embedding_storage_mode: default is 'none' which is always best. Only set to 'cpu' or 'gpu' if
+        you wish to not only predict, but also keep the generated embeddings in CPU or GPU memory respectively.
+        'gpu' to store embeddings in GPU memory.
+        """
+
+        if not sentences:
+            return sentences
+        sentences = SentenceDataset(sentences)
+        data_loader = DataLoader(sentences, batch_size=mini_batch_size, num_workers=num_workers)
+
+        for batch in data_loader:
+            with torch.no_grad():
+                score_arc, score_rel = self.forward(batch)
+                arc_prediction, relation_prediction = self._obtain_labels_(
+                    score_arc, score_rel)
+
+            for sentnce_index, (sentence, sent_tags, sent_arcs) in enumerate(zip(batch, relation_prediction, arc_prediction)):
+                for token_index, (token, tag, head_id) in enumerate(zip(sentence.tokens, sent_tags, sent_arcs)):
+                    token.add_tag(self.tag_type, tag,
+                                  score_rel[sentnce_index][token_index])
+                    token.head_id = int(head_id)
+
+                if print_tree:
+                    tree_printer(sentence, self.tag_type)
+                    print("-" * 50)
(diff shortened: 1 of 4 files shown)


alanakbik:
@73minerva really cool, thanks for adding this and sorry for not getting around to reviewing this sooner.  

I'm trying to train a model, but its throwing errors if the mini-batch size is higher than 1. See script below - am I instantiating it wrong? 

```python
corpus = UD_ENGLISH()

dictionary = corpus.make_label_dictionary("dependency")

model = DependencyParser(token_embeddings=WordEmbeddings("turian"), relations_dictionary=dictionary)

trainer.train("resources/taggers/dependency", mini_batch_size=2)
```

73minerva:
@alanakbik, The recent commit should fix the mentioned mini-batch error. I also added word dropout.
[Here](https://drive.google.com/file/d/1sBYFsRQkvC5wwE5kzXJSZzZjylC3dibt/view?usp=sharing) is a trained dependency parser model with bert that achieved **UAS : 0.9210 - LAS : 0.9354**.

stefan-it:
Hi @73minerva ,

thanks for that PR! I wanted to test it with Transformer-based models, but this failed due to an error that was I think fixed a while ago on Flair master. So could you please rebase your PR on latest master branch :thinking: Many thanks!

flair/models/dependency_parser_model.py
```diff
@@ -0,0 +1,624 @@
...
+                    token: Token = token
+                    token.add_tag_label("predicted", Label(tag))
+                    token.add_tag_label("predicted_head_id", Label(str(arc)))
+
+                    # append both to file for evaluation
+                    eval_line = "{} {} {} {} {}\n".format(token.text,
+                                                          token.get_tag(gold_label_type).value,
+                                                          str(token.head_id),
+                                                          tag,
+                                                          str(arc))
```
alanakbik, line 279: this is sometimes a tensor and sometimes an int, leading to uneven printouts

alanakbik (approved):
A few minor comments and some questions on parameter choices.

If you prefer, I can merge this PR now and do the rebase to master @stefan-it mentioned and we can do further improvements in more PRs.

flair/models/dependency_parser_model.py
```diff
@@ -0,0 +1,624 @@
...
+    def __init__(
+            self,
+            token_embeddings: TokenEmbeddings,
+            relations_dictionary: Dictionary,
+            tag_type: str = 'dependency',
+            lstm_hidden_size: int = 400,
+            mlp_arc_units: int = 500,
+            mlp_rel_units: int = 100,
+            lstm_layers: int = 3,
+            mlp_dropout: float = 0.33,
```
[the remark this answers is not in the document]
73minerva: All default hyperparameters including dropout, were set according to the [paper](https://arxiv.org/pdf/1611.01734.pdf). As mentioned in the paper, to reduce label classifier overfitting, increasing dropout is necessary. More details and why label classifier suffers from overfitting are explained in the 4.2.1 section of the paper. I haven't tried lower values and unfortunately, there is no experiment about it.

flair/models/dependency_parser_model.py
73minerva: Thanks, It's fixed on the recent commit. Also, I add CSV header to prevent confusion. it's better now.


Screenshot from 2021-11-28 11-38-57

alanakbik:
@73minerva another thing: the LAS should only count positives if both attachment and deprel are correctly predicted. So the LAS should always be lower than the UAS. I have it fixed in my local branch and will do a PR soon!

Issue #2489, opened 2021-10-23
Error with saving OneHotEmbeddings with CoNLLUCorpus


djstrong:
**Describe the bug**
Saving model with OneHotEmbeddings and corpus CoNLLUCorpus causes `_pickle.PicklingError: Can't pickle <function <lambda> at 0x7fa069ae2f28>: attribute lookup <lambda> on conllu.parser failed`.

**To Reproduce**
https://colab.research.google.com/drive/1tAQcAGkdJSApCcW-DxTbx_kGHQtdenUD?usp=sharing

PR #2490, opened 2021-10-24
refactor onehot embeddings

helpmefindaname:
a few changes to the OneHotEmbeddings:

- allows creating OneHotEmbeddings by a Dictionary directly, in case you don't want to gather it from a corpus.
- still allows creating OneHotEmbeddings by a Corpus, using `OneHotEmbeddings.from_corpus(..)` 
- should fix https://github.com/flairNLP/flair/issues/2489 as the instance_parameters cover the Dictionary and not the Corpus -> it won't store the whole dataset and therefore reduces model size.
- allows usage of "stable embeddings" -> https://arxiv.org/abs/2110.02861 which seem to "reduce gradient variance that comes from the highly non-uniform distribution of input tokens"
- makes the `_add_embeddings_internal` method more beautiful, by removing nested loops (iterates over all tokens instead of all sentences and their tokens) and removing unneeded indexing/enumerations

### DIFF
--- a/flair/embeddings/token.py
+++ b/flair/embeddings/token.py
@@ -1431,24 +1431,88 @@ class OneHotEmbeddings(TokenEmbeddings):
 
     def __init__(
             self,
-            corpus: Corpus,
+            vocab_dictionary: Dictionary,
             field: str = "text",
             embedding_length: int = 300,
-            min_freq: int = 3,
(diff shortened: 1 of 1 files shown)


alanakbik:
@helpmefindaname this is awesome, thanks for fixing this!

PR #2491, opened 2021-10-24
allow finetuning of word-embeddings

helpmefindaname:
This PR reworks the `WordEmbeddings` to add a few features:

- embeddings are stored as a torch Embedding instead of a gensim keyedvector. That way it will never come to version issues, if gensim doesn't ensure backwards compatibility (happened to me a few times over the last 2 years)
- embeddings can be put on gpu, if wanted. For that, `WordEmbeddings('<name>', force_cpu=False)` can be used
- embeddings can be fine-tuned. For that, `FordEmbeddings('<name>', fine_tune=True)` can be used.
- the lru_cache will store indices per word instead of vector lookups. This is because the lookups shouldn't take long if calculating all indices at once (especially if using gpu). However, the regex operations and multiple lookups will remain cached.
- lru_cache is increased, as it takes (<embedding_length> - 1)* 4 bytes less per lookup.
- allows usage of "stable embeddings" → https://arxiv.org/abs/2110.02861 which seem to "reduce gradient variance that comes from the highly non-uniform distribution of input tokens"

### DIFF
--- a/flair/embeddings/token.py
+++ b/flair/embeddings/token.py
@@ -192,17 +192,25 @@ class WordEmbeddings(TokenEmbeddings):
         self.static_embeddings = True
 
         if str(embeddings).endswith(".bin"):
-            self.precomputed_word_embeddings = gensim.models.KeyedVectors.load_word2vec_format(
+            precomputed_word_embeddings = gensim.models.KeyedVectors.load_word2vec_format(
                 str(embeddings), binary=True
             )
         else:
-            self.precomputed_word_embeddings = gensim.models.KeyedVectors.load(
+            precomputed_word_embeddings = gensim.models.KeyedVectors.load(
                 str(embeddings)
             )
 
+        self.__embedding_length: int = precomputed_word_embeddings.vector_size
+        self.vectors = np.row_stack(
+            (precomputed_word_embeddings.vectors, np.zeros(self.__embedding_length, dtype="float"))
+        )
+        self.vocab = {
+            k: v.index
+            for k, v in precomputed_word_embeddings.vocab.items()
+        }
+
         self.field = field
 
-        self.__embedding_length: int = self.precomputed_word_embeddings.vector_size
         super().__init__()
 
     @property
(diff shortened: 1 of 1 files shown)


flair/embeddings/token.py
```diff
@@ -233,18 +261,25 @@ def get_cached_vec(self, word: str) -> torch.Tensor:
...
+                word = token.text
+            else:
+                word = token.get_tag(self.field).value
+            word_indices.append(self.get_cached_token_index(word))
 
-                if "field" not in self.__dict__ or self.field is None:
-                    word = token.text
-                else:
-                    word = token.get_tag(self.field).value
+        embeddings = self.embedding(torch.tensor(word_indices, dtype=torch.int, device=self.device))
```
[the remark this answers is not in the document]
helpmefindaname: thanks for clarifying, yes that totally makes sense, I'l change it

flair/embeddings/token.py
```diff
@@ -119,16 +120,21 @@ def get_named_embeddings_dict(self) -> Dict:
...
         :param embeddings: one of: 'glove', 'extvec', 'crawl' or two-letter language code or custom
         If you want to use a custom embedding file, just pass the path to the embeddings as embeddings variable.
+        set stable=True to use the stable embeddings as described in https://arxiv.org/abs/2110.02861
         """
         self.embeddings = embeddings
 
         self.instance_parameters = self.get_instance_parameters(locals=locals())
 
+        if fine_tune and force_cpu and flair.device.type != "cpu":
+            raise Exception("Cannot train WordEmbeddings on cpu if the model is trained on gpu")
```
alanakbik, line 136: And one more minor thing: Perhaps in the error message you could point to either setting `force_cpu=False` to enable training on GPU?

helpmefindaname: sounds good

alanakbik:
@helpmefindaname thanks a lot for adding this, really cool features! 

I ran some experiments with the vanilla Flair configuration on CoNLL-03 (glove+flair embeddings) with three settings: no-finetune (i.e. same as before), finetune-normal (with finetuning) and finetune-stable (with finetuning and stable=True). Each configuration ran 5 times, here are averages and standard deviation: 

| Approach                      | Test F1 |
|----|----|
| no-finetune | 93.01 +- 0.12 |
| finetune-normal | 93.03 +- 0.08 |
| finetune-stable | 92.76 +- 0.04 |

alanakbik:
@helpmefindaname we just realized that `gensim` is pinned to 3.8. max in Flair, which does not work with Python 3.9. 

So I was thinking of upgrading to `gensim` version 4, but there's [lots of changes](https://github.com/RaRe-Technologies/gensim/wiki/Migrating-from-Gensim-3.x-to-4) in the syntax, so I guess that would mean that 3 no longer works. Do you think requiring gensim > 4 makes sense in Flair?

PR #2492, opened 2021-10-30
Add pretrained relation extraction models

alanakbik:
This PR adds a few relation extraction models, trained over a modified version of TACRED. Two models are added: `relations` and `relations-fast`.

To use these models, you also need an entity tagger. The tagger identifies entities, then the relation extractor possible entities. For instance use this code: 

```python
from flair.models import RelationExtractor, SequenceTagger

# 1. make example sentence
sentence = Sentence("George was born in Washington")

# 2. load entity tagger and predict entities
tagger = SequenceTagger.load('ner-fast')
tagger.predict(sentence)

# check which entities have been found in the sentence
entities = sentence.get_labels('ner')
for entity in entities:
    print(entity)

# 3. load relation extractor
extractor: RelationExtractor = RelationExtractor.load('relations-fast')

# predict relations
extractor.predict(sentence)

# check which relations have been found
relations = sentence.get_labels('relation')
for relation in relations:
    print(relation)
```

### DIFF
--- a/flair/models/relation_extractor_model.py
+++ b/flair/models/relation_extractor_model.py
@@ -205,12 +205,14 @@ class RelationExtractor(flair.nn.DefaultClassifier):
 
                 relation_embeddings.append(embedding)
 
-            # stack and drop out
-            all_relations = torch.stack(relation_embeddings)
+            # stack and drop out (squeeze and unsqueeze)
+            all_relations = torch.stack(relation_embeddings).unsqueeze(1)
 
             all_relations = self.dropout(all_relations)
             all_relations = self.locked_dropout(all_relations)
             all_relations = self.word_dropout(all_relations)
+
+            all_relations = all_relations.squeeze(1)
 
             # send through decoder
             if self.non_linear_decoder:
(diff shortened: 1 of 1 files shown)


alejandrojcastaneira:
Thank you very much for adding this, can you provide some insights or a short description of the architecture of the relations extraction models?

alanakbik:
Hello @alejandrojcastaneira it we used a very simple architecture here: the `RelationExtractor` goes through all pairwise combinations of entities found in a sentence. For each entity pair, it takes the embeddings of both entities and concatenates them. This concatenated entity pair embedding is then classified.

SohaK:
@alanakbik @djstrong in my question I meant positional context not semantic context. Is there any paper or blog that shed some light to the FLAIR Relation Extraction?

djstrong:
Semantic context is usually positional. Flair RE is extracting embeddings of the spans, concats them and pass through fully connected neural network.

SohaK:
@djstrong is the 'embedding' the dynamic embedding of the span in the current sentence, or the pretrained embedding?

djstrong:
You can choose embeddings (dynamic or static).

Issue #2210, opened 2021-04-15
OverflowError: int too big to convert


gunturbudi:
Hello, 
I'm trying to train a named entity recognition model, with this embedding: `TransformerWordEmbeddings('emilyalsentzer/Bio_ClinicalBERT')`. However, it always failed with `OverflowError: int too big to convert`. This is also happening in some other transformer word embedding such as `XLNet`. However, `BERT` and `RoBERTa` works fine.

Here is the full traceback of the error:

```
2021-04-15 09:34:48,106 ----------------------------------------------------------------------------------------------------
2021-04-15 09:34:48,106 Corpus: "Corpus: 778 train + 259 dev + 260 test sentences"
2021-04-15 09:34:48,106 ----------------------------------------------------------------------------------------------------
2021-04-15 09:34:48,106 Parameters:
2021-04-15 09:34:48,106  - learning_rate: "0.1"
2021-04-15 09:34:48,106  - mini_batch_size: "32"
2021-04-15 09:34:48,106  - patience: "3"
2021-04-15 09:34:48,106  - anneal_factor: "0.5"
2021-04-15 09:34:48,106  - max_epochs: "200"
2021-04-15 09:34:48,106  - shuffle: "True"
2021-04-15 09:34:48,106  - train_with_dev: "False"
2021-04-15 09:34:48,106  - batch_growth_annealing: "False"
2021-04-15 09:34:48,107 ----------------------------------------------------------------------------------------------------
2021-04-15 09:34:48,107 Model training base path: "/home/xxx/data/xxx-clinical-bert"
2021-04-15 09:34:48,107 ----------------------------------------------------------------------------------------------------
2021-04-15 09:34:48,107 Device: cuda:0
2021-04-15 09:34:48,107 ----------------------------------------------------------------------------------------------------
2021-04-15 09:34:48,107 Embeddings storage mode: gpu
2021-04-15 09:34:48,116 ----------------------------------------------------------------------------------------------------
Traceback (most recent call last):
  File "train_medical_2.py", line 144, in <module>
    train_ner(d + '-base-ent',corpus_base)
  File "train_medical_2.py", line 136, in train_ner
    max_epochs=200)
  File "/home/d111199102201607101/flair/lib/python3.6/site-packages/flair/trainers/trainer.py", line 381, in train
    loss = self.model.forward_loss(batch_step)
[...]

Issue #2501, opened 2021-11-03
add default model_max_length for HF Tokenizer


whoisjones:
Due to the increasing amount of issue for HF Tokenizer where model_max_length is not set, add a default processing for it.

Related issues: gh-2210, gh-2476, gh-2305, gh-2500

Issue #2500, opened 2021-11-02
using `TransformerWordEmbeddings`


zingxy:
**Describe the bug**
when I used Huggingface model ''dmis-lab/biobert-base-cased-v1.2'' to embed a sentence, it raises  `OverflowError: int too big to convert`. And model `bert-base-cased`  works.

**To Reproduce**

```
bert = TransformerWordEmbeddings('bert-base-cased')
# biobert was trained on pubmed and pmc corpus.
biobert = TransformerWordEmbeddings('dmis-lab/biobert-base-cased-v1.2')
bert.embed(Sentence('hello'))
biobert.embed(Sentence('hello'))
```

**Expected behavior**

**Screenshots**
```
>>>bert.tokenizer
PreTrainedTokenizerFast(name_or_path='bert-base-cased', vocab_size=28996, model_max_len=512, is_fast=True, padding_side='right', special_tokens={'unk_token': '[UNK]', 'sep_token': '[SEP]', 'pad_token': '[PAD]', 'cls_token': '[CLS]', 'mask_token': '[MASK]'})
>>>biobert.tokenizer
PreTrainedTokenizerFast(name_or_path='dmis-lab/biobert-base-cased-v1.2', vocab_size=28996, **model_max_len=1000000000000000019884624838656**, is_fast=True, padding_side='right', special_tokens={
```
The biobert tokenizer's **model_max_len** may  cause this error, but I don't know how to fix this bug.

whoisjones:
hi @zingxy,

it seems that BioBert has no model_max_length set. If you set the respective parameter `model_max_length` of huggingface's Tokenizer class to `512`, it works. However, this should be fixed in the config file of BioBert I guess.

`biobert = TransformerWordEmbeddings('dmis-lab/biobert-base-cased-v1.2', model_max_length=512)`

See also related issues: [hf-9317](https://discuss.huggingface.co/t/cannot-create-an-identical-pretrainedtokenizerfast-object-from-a-tokenizer-created-by-tokenizers-library/9317) and gh-2305

zingxy:
@whoisjones Thank you for your detailed answer. According to your method, I solved the problem. Thank you.

Issue #2476, opened 2021-10-12
OverflowError: int too big to convert


MelaniaNitu:
**Describe the bug**
I'm using flair to import existing embedding from Huggingface to train BERTopic as described [here](https://maartengr.github.io/BERTopic/tutorial/embeddings/embeddings.html#flair). As per documentation, we should be able to select any transformer model available on [huggingface](https://huggingface.co/models). For some of the pre-trained models I get this error: **"OverflowError: int too big to convert"**.

**To Reproduce**
```
from flair.embeddings import TransformerDocumentEmbeddings
robert = TransformerDocumentEmbeddings('readerbench/RoBERT-base')
model = BERTopic(embedding_model=robert,verbose=True) 
docs = df.text.to_list()
topics, probabilities = model.fit_transform(docs)
```
Output:
```
0it [00:00, ?it/s]
---------------------------------------------------------------------------
OverflowError                             Traceback (most recent call last)
<ipython-input-17-4da31a0e0fc5> in <module>()
      9 
     10 #train model
---> 11 topics, probabilities = model.fit_transform(docs)

10 frames
/usr/local/lib/python3.7/dist-packages/transformers/tokenization_utils_fast.py in set_truncation_and_padding(self, padding_strategy, truncation_strategy, max_length, stride, pad_to_multiple_of)
    352             target = {"max_length": max_length, "stride": stride, "strategy": truncation_strategy.value}
    353             if _truncation != target:
--> 354                 self._tokenizer.enable_truncation(**target)
    355 
    356         if padding_strategy == PaddingStrategy.DO_NOT_PAD:

**OverflowError: int too big to convert**
```

**Expected behavior**

Similar issue was reported for TransformerWordEmbeddings. Solution for that one is described here: https://githubmemory.com/repo/flairNLP/flair/issues/2210 

class TransformerWordEmbeddings(TokenEmbeddings):
--->    NO_MAX_SEQ_LENGTH_MODELS=[XLNetModel, TransfoXLModel, **BertModel**]

**Environment**
Google Colab (Jupyter Notebook)

**Additional context**
Similar issue was reported for **TransformerWordEmbedding** for XLNet model which was fixed by @tiagokv. It looks like the PR fix made by @tiagokv is needed as well for the other transformers models besides XLNet so TransformerWordEmbeddings can work properly. 

Now, in the case of **TransformerDocumentEmbeddings**, I guess we would have to "port" that fix to make it work with TransformerDocumentEmbeddings as well.

[...]

PR #2502, opened 2021-11-03
default handling when Tokenizer's model_max_length is not set

whoisjones:
fixes and closes gh-2210, gh-2305, gh-2476, gh-2500


whoisjones:
closes gh-2501

alanakbik:

PR #2507, opened 2021-11-07
add MultiCoNer Dataset

helpmefindaname:
Add a loader for the new Dataset of SemEval2022 Task 11: https://multiconer.github.io/competition
notice, that the dataset has to be manually applied for and downloaded here: https://multiconer.github.io/dataset

after setting it up, the corpus can be loaded for any of the tasks, e.g.
`corpus: Corpus = MultiCoNer("en")` for english monolingual, or `corpus: Corpus = MultiCoNer("multi")` for multilingual of all datasets


alanakbik:
@helpmefindaname this is really cool, thanks for adding this! Once Flair 0.10.0 is released, perhaps we could even notify the organizers that the dataset is supported in Flair to easily train some baseline models?

PR #2510, opened 2021-11-10
Lemmatizer

marcelmmm:


flair/models/lemmatizer_model.py
```diff
@@ -0,0 +1,849 @@
...
+        self.hidden_size = hidden_size
+        if self.token_embedding:
+            self.emb_to_hidden = nn.Linear(token_embedding.embedding_length, hidden_size)
+
+        # decoder: dictionary of characters to create one-hot embeddings for decoder inputs
+        self.path_to_char_dict = path_to_char_dict
+        if path_to_char_dict is None and path_to_corpus_for_creating_char_dict is None:
+            self.char_dictionary = Dictionary.load("common-chars-lemmatizer")
+        elif path_to_char_dict:
+            self.char_dictionary = Dictionary.load_from_file(path_to_char_dict)
```
alanakbik, line 81: This does not guarantee that the requires special characters will be part of the dictionary.

flair/models/lemmatizer_model.py
```diff
@@ -0,0 +1,849 @@
...
+
+
+        # encoder: embed tokens and map embedding to hidden size of RNN
+        self.token_embedding = token_embedding
+        self.hidden_size = hidden_size
+        if self.token_embedding:
+            self.emb_to_hidden = nn.Linear(token_embedding.embedding_length, hidden_size)
+
+        # decoder: dictionary of characters to create one-hot embeddings for decoder inputs
+        self.path_to_char_dict = path_to_char_dict
```
alanakbik, line 77: Why is this a class variable? Isn't it enough to load the dictionary at this point?

flair/models/lemmatizer_model.py
```diff
@@ -0,0 +1,849 @@
...
+        if not seq_length:
+            # sequence length of each word is equal to max length of a word in the sentence plus one (start character <S>)
+            sequence_length = max_length
+        else:
+            sequence_length=max(seq_length,max_length)
+        # batch size is length of sentence
+        tensor = torch.ones(len(words), sequence_length, dtype=torch.long).to(flair.device)
+        for i in range(len(words)):
+            if for_input:
+                tensor[i][0] = 2  # start character <S>
```
alanakbik, line 139: This is not robust to other dictionaries where the <S> token might have another index. Better never hard-code indices.

alanakbik (commented):
Some comments regarding dictionary (see upcoming PR)

flair/models/lemmatizer_model.py
marcelmmm: True, I changed it now.

flair/models/lemmatizer_model.py
marcelmmm: Done

flair/models/lemmatizer_model.py
marcelmmm: Replaced by variables

alanakbik:
Closed, see #2531

PR #2518, opened 2021-11-15
Removes hyperparameter features

alanakbik:
Removes the old and long-deprecated hyperparameter search functionality from Flair.


fabero:
Just out of curiosity: what is the reason for deprecating and ultimately removing the hyperparameter search functionality?

Is it because the implementation was not considered good enough, not important enough or something else?

Thanks in advance!

tadejmagajna:
@alanakbik +1 to the comment above

PR #2525, opened 2021-11-18
readme: minor fixes in sota models section

stefan-it:
This PR introduces minor fixes in the sota models section :)


alanakbik:
@stefan-it thanks for correcting this!

PR #2531, opened 2021-11-30
Lemmatizer

alanakbik:
This PR adds a lemmatizer to Flair based on Word Embeddings plus a character-level sequence-to-sequence model with attention.

Future PRs will add pre-trained models.


alanakbik:
@marcelmmm thanks for implementing this!

abdullah-alnahas:
Thanks for adding the lemmatizer and all the great work.
I could not find any info about using the lemmatizer. Is it available in v 0.10? What are the supported languages?
Thanks!

marcelmmm:
The Lemmatizer model itself is available but we do not provide any pre-trained models so far. You could train a model yourself.

Issue #2504, opened 2021-11-05
Add `__eq()__` to `Dictionary` class


sgraaf:
**Describe the solution you'd like**
I would like to be able to directly compare to `Dictionary` objects, such that when fine-tuning a pre-trained NER tagger, I can write some logic to see whether or not I need to manipulate the tagger's `state`, per a comment in #1540.

The solution is very simple, just add the following code to the `Dictionary` class:
```Python
def __eq__(self, o: object) -> bool:
    if not isinstance(o, Dictionary):
        return False
    return self.item2idx == o.item2idx and self.idx2item == o.idx2item and self.add_unk == o.add_unk
```

alanakbik:
@marcelmmm can you take a look?

PR #2532, opened 2021-11-30
GH-2504 add equality method to Dictionary class

marcelmmm:
Sensible addtition to Dictionary. Closes GH-2504


alanakbik:

PR #2533, opened 2021-11-30
Implements a RegexpTagger

MaxDall:


alanakbik:
@MaxDall works well! Thanks for adding this! 

You can call the tagger and define regular expressions that match to span annotations. I.e.: 

```python
    # sentence with a number and two quotes
    sentence = Sentence('Figure 11 is both "too colorful" and "not informative enough".')

    # instantiate regex tagger with a quote matching pattern
    tagger = RegexpTagger(mapping=(r'(["\'])(?:(?=(\\?))\2.)*?\1', 'QUOTE'))

    # also add a number mapping
    tagger.register_labels(mapping=(r'\b\d+\b', 'NUMBER'))

    # tag sentence
    tagger.predict(sentence)

    # check out matches
    for entity in sentence.get_labels():
        print(entity)
```

Issue #2508, opened 2021-11-10
Multitask Learning in flair


whoisjones:
We would like to enhance flair with multitask learning, e.g. combine various models and share specific layers across several tasks.

**Is your feature/enhancement request related to a problem? Please describe.**
New feature.

**Describe the solution you'd like**
A modular, easy-to-use multitask framework solution.

djstrong:
I would be interested in multitask training NER and relation extraction using the same dataset. So it should compute embeddings once for both heads. I guess it would be complicated to implement?

Issue #2445, opened 2021-09-21
Cannot load NER model from local file


DieseKartoffel:
Hallo, I am trying to use NER-German-Large in an environment without internet connection. I downloaded the model to a local directory from [here](https://huggingface.co/flair/ner-german-large) but I can not initialize it, since the corresponding TransformerWordEmbeddings seem to be missing locally.

How do I find out what files are missing exactly, where can I download them and how do I load and use them in my script?

Here is the code I am running:
```
from flair.models import SequenceTagger
tagger = SequenceTagger.load("../my_path/ner-german-large/pytorch_model.bin")
```

Output without internet connection:
```
Traceback (most recent call last):
  File "my_file.py", line 104, in <module>
    tagger = SequenceTagger.load("../my_path/ner-german-large/pytorch_model.bin")
  File "/opt/conda/lib/python3.8/site-packages/flair/nn/model.py", line 101, in load
    state = torch.load(f, map_location='cpu')
  File "/opt/conda/lib/python3.8/site-packages/torch/serialization.py", line 592, in load
    return _load(opened_zipfile, map_location, pickle_module, **pickle_load_args)
  File "/opt/conda/lib/python3.8/site-packages/torch/serialization.py", line 851, in _load
    result = unpickler.load()
  File "/opt/conda/lib/python3.8/site-packages/flair/embeddings/token.py", line 1332, in __setstate__
    embedding = TransformerWordEmbeddings(
  File "/opt/conda/lib/python3.8/site-packages/flair/embeddings/token.py", line 838, in __init__
    self.tokenizer: PreTrainedTokenizer = AutoTokenizer.from_pretrained(model, **kwargs)
  File "/opt/conda/lib/python3.8/site-packages/transformers/models/auto/tokenization_auto.py", line 445, in from_pretrained
    return tokenizer_class_fast.from_pretrained(pretrained_model_name_or_path, *inputs, **kwargs)
  File "/opt/conda/lib/python3.8/site-packages/transformers/tokenization_utils_base.py", line 1672, in from_pretrained
    resolved_vocab_files[file_id] = cached_path(
  File "/opt/conda/lib/python3.8/site-packages/transformers/file_utils.py", line 1329, in cached_path
    output_path = get_from_cache(
  File "/opt/conda/lib/python3.8/site-packages/transformers/file_utils.py", line 1552, in get_from_cache
    raise ValueError(
[...]

PR #2558, opened 2021-12-16
Bf/combine transformer embeddings

helpmefindaname:
creates a transformer embedding that combines both `TransformerWordEmbedding` and `TransformerDocumentEmbedding`

it should be able to:
* still load all embeddings saved with previous versions
* should not need to load the tokenizer -> https://github.com/flairNLP/flair/issues/2445 should be fixed
* allow using document embeddings and token embeddings at once (might be usefull for https://github.com/flairNLP/flair/issues/2508 ?)
* allow using long sequences / document context for DocumentClassification too (if the pooling is `max` or `mean`)
* make code better maintainable, as there is no duplicated code

current state:

- [x] first implementation
- [x] run test
- [x] manually test training with all kinds of features
- [x] test loading old embeddings / test backwardscompability

### DIFF
--- a/flair/embeddings/base.py
+++ b/flair/embeddings/base.py
@@ -270,7 +270,6 @@ class TransformerEmbedding(Embeddings[Sentence]):
 
         self.cls_pooling = cls_pooling
         self.subtoken_pooling = subtoken_pooling
-
         self.layer_mean = layer_mean
         self.fine_tune = fine_tune
         self.static_embeddings = not self.fine_tune
@@ -347,9 +346,11 @@ class TransformerEmbedding(Embeddings[Sentence]):
         with tempfile.TemporaryDirectory() as temp_dir:
             files = self.tokenizer.save_pretrained(temp_dir)
             zip_data = BytesIO()
-            zip_obj = zipfile.ZipFile(zip_data, "w+")
+            zip_obj = zipfile.ZipFile(zip_data, "w")
             for f in files:
-                zip_obj.write(f, os.path.relpath(f, temp_dir))
+                # transformers returns the "added_tokens.json" even if it doesn't create it
+                if os.path.exists(f):
+                    zip_obj.write(f, os.path.relpath(f, temp_dir))
 
         zip_data.seek(0)
         return zip_data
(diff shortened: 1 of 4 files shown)


flair/embeddings/base.py
```diff
@@ -149,3 +165,616 @@ def forward(self, tensors: List[torch.Tensor]) -> torch.Tensor:
...
+                    )[1].tolist(),
+                )
+            else:
+                sentence_hidden_states = list(hidden_states.permute((1, 0, 2, 3)))
+
+            # remove padding tokens
+            sentence_hidden_states = [
+                sentence_hidden_state[:, : len(subtokens) + 1, :]
+                for (subtokens, sentence_hidden_state) in zip(tokenized_sentences, sentence_hidden_states)
+            ]
```
alanakbik, line 719: This causes problems if a sentence has more subtokens than tokens. Padding tokens should be removed to subtoken length:

```suggestion
            sentence_hidden_states = [
                sentence_hidden_state[:, : sum(token_subtoken_lengths) + 1, :]
                for (token_subtoken_lengths, sentence_hidden_state) in zip(all_token_subtoken_lengths, sentence_hidden_states)
            ]
```

alanakbik (changes requested):
Thanks for adding this! Still testing, but found an error that appears with the following code: 

```python
embeddings = TransformerWordEmbeddings(model='xlm-roberta-base',
                                       layers="-1",
                                       subtoken_pooling="first",
                                       fine_tune=True,
                                       use_context=False,
                                       )
text = "."
sentence = Sentence(text)
embeddings.embed(sentence)
```

Suggestion to solve this (I think) added in-line.

flair/embeddings/base.py
helpmefindaname: Very good catch, thank you.
My fix is slightly different: `all_token_subtoken_lengths` is only filled if it is a token embedding, as document embeddings don't need the subtoken_token mapping and shouldn't create that overhead.
besides that, I noticed that if the last subtoken doesn't map to a token it will be discarded, so I rather store the full subtoken length and take that one. Hence we can ensure that TransformerDocumentEmbeddings with pad_last always have the right token as last.

alanakbik (approved):
Thanks a lot for refactoring this!

alanakbik:
@helpmefindaname I found another error. It seems the fix for the previous error now broke sentences that are too long (over 512 subtokens).

Reproducible with this script: 

```python

# example transformer embeddings
embeddings = TransformerWordEmbeddings(model='distilbert-base-uncased')

# create sentence with more than 512 subtokens
long_sentence = Sentence('a ' * 513)

# embed
embeddings.embed(long_sentence)
```

Throws the same assertion error as previously, i.e.: 

```console
  File ".../flair/flair/embeddings/base.py", line 769, in _add_embeddings_internal
    self._add_embeddings_to_sentences(expanded_sentences)
  File ".../flair/flair/embeddings/base.py", line 728, in _add_embeddings_to_sentences
    self._extract_token_embeddings(sentence_hidden_states, sentences, all_token_subtoken_lengths)
  File ".../flair/flair/embeddings/base.py", line 656, in _extract_token_embeddings
    assert subword_start_idx < subword_end_idx <= sentence_hidden_state.size()[1]
AssertionError
```

Any ideas how to fix this?

isaac47:
Hi,
this merged request is done? How can use it?

Issue #2560, opened 2021-12-17
predict method in SequenceTagger cannot read flair Dataset


whoisjones:
**Describe the bug**
New version of SequenceTagger only accept list of sentences. If you pass a flair dataset such as corpus.test, this will be understood as a sentence.

**To Reproduce**
Load any tagger, call tagger.predict() with a flair.dataset object.

Issue #2559, opened 2021-12-16
Compute each score for each token in SequenceTagger


whoisjones:
Previous model was able to do, current refactoring of SequenceTagger class does not support it.

PR #2561, opened 2021-12-17
GH-2559: Return all tag probabilities + GH-2560: predict() method can handle flair datasets

whoisjones:
closes GH-2559 and closes GH-2560


alanakbik:

Issue #2582, opened 2022-01-05
TransformerEmbeddings should process long sequences by default


whoisjones:
In TransformerEmbedding refactoring, the constructor default argument has changed. Default should be `allow_long_sequences=True`.

PR #2584, opened 2022-01-05
TransformerWordEmbeddings to process long sequences by default

whoisjones:
closes gh-2582


alanakbik:

Issue #2577, opened 2021-12-29
'Expected all tensors to be on the same device' while training LanguageModel


Tarotis:
**Describe the bug**
While training my own FlairEmbeddings, I receive the error message that not all tensors are on the same device. This is both for custom and loaded dictionaries.

**Expected behavior**
All tensors are on the same device.

**The training script:**
```
from flair.data import Dictionary
from flair.models import LanguageModel
from flair.trainers.language_model_trainer import LanguageModelTrainer, TextCorpus
from helper_functions import make_dictionary

dictionary: Dictionary = make_dictionary('data/Shipibo/embeddings/train/train.txt')
# dictionary: Dictionary = Dictionary.load('chars')

is_forward_lm = True

corpus = TextCorpus('data/Shipibo/embeddings', dictionary, is_forward_lm, character_level=True)

language_model = LanguageModel(dictionary, is_forward_lm, hidden_size=512,nlayers=1)

trainer = LanguageModelTrainer(language_model, corpus)
trainer.train('models/resources/embeddings/sk_word', sequence_length=11, learning_rate=0.3, mini_batch_size=2, max_epochs=30)
```

**The error message:**
```
2021-12-29 16:43:52,094 Sequence length is 11
2021-12-29 16:43:52,107 Split 1  - (16:43:52)
2021-12-29 16:43:52,536 | split   1 /  1 |   100/65123 batches | ms/batch  4.28 | loss  3.46 | ppl    31.66
Traceback (most recent call last):
  File "models/flair_embeddings.py", line 38, in <module>
    max_epochs=30)
  File "~/flair/trainers/language_model_trainer.py", line 324, in train
    total_loss += loss.data
RuntimeError: Expected all tensors to be on the same device, but found at least two devices, cuda:0 and cpu!
```
**Environment (please complete the following information):**
 - flair master branch
 
The error does **not** reproduce with the most recent published version.

PR #2590, opened 2022-01-06
small lm training improvements

helpmefindaname:
fix lm training (fix https://github.com/flairNLP/flair/issues/2577 )
nice output in lm training
allow usage of lstm and gru in flair lm
remove unnecessary imports in flair/__init__.py

@alanakbik this is basically some modifications I did while training the SRU, however the SRU itself was not sucessfull enough to be merged (not only that it is a bit more complicated to install it due to special cuda requirements)

### DIFF
--- a/flair/models/language_model.py
+++ b/flair/models/language_model.py
@@ -4,10 +4,12 @@ from typing import List, Optional, Tuple, Union
 
 import torch
 import torch.nn as nn
+from torch import logsumexp
 from torch.optim import Optimizer
 
 import flair
 from flair.data import Dictionary
+from flair.nn.recurrent import create_recurrent_layer
 
 
 class LanguageModel(nn.Module):
@@ -23,6 +25,7 @@ class LanguageModel(nn.Module):
         nout=None,
         document_delimiter: str = "\n",
         dropout=0.1,
+        recurrent_type="LSTM",
     ):
 
         super(LanguageModel, self).__init__()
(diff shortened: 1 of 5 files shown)


flair/models/language_model.py
```diff
@@ -82,20 +83,19 @@ def forward(self, input, hidden, ordered_sequence_lengths=None):
...
 
     def init_hidden(self, bsz):
         weight = next(self.parameters()).detach()
-        return (
-            weight.new(self.nlayers, bsz, self.hidden_size).zero_().clone().detach(),
-            weight.new(self.nlayers, bsz, self.hidden_size).zero_().clone().detach(),
-        )
+        w1 = weight.new(self.nlayers, bsz, self.hidden_size).zero_().clone().detach()
+        w2 = weight.new(self.nlayers, bsz, self.hidden_size).zero_().clone().detach()
+        return w1, w2
```
alanakbik, line 98: in a GRU, only a single state is needed

flair/trainers/trainer.py
```diff
@@ -509,7 +510,6 @@ def train(
 
                     # depending on memory mode, embeddings are moved to CPU, GPU
                     # or deleted
```
alanakbik, line 512: I guess the comments on storing embeddings should move too ;)

alanakbik (changes requested):
Thanks for improving this and sorry for the late review! It mostly looks good, but the GRU is currently not working, i.e. 

```python
language_model = LanguageModel(
    dictionary,
    is_forward_lm=True,
    hidden_size=128,
    nlayers=1,
    recurrent_type="GRU",
)

# train your language model

# train your language model

trainer.train(
    "resources/LMs/test",
...
```

due to the hidden state being created for LSTM.

helpmefindaname:
my bad, I removed the states part again, as when I removed the SRU again,
now it should work

alanakbik (approved):

alanakbik:
@helpmefindaname looks great, thanks for improving this!

PR #2599, opened 2022-01-17
long sequence transformers of any kind

helpmefindaname:
apparently the logic of the Transformers tokenizers is that if the sequence lenght exceeds a very large number, it counts as infinitive. 
this includes models like DebertaV3 but also XLNetModel.

For TransfoXLModel I couldn't test if this logic is valid, as I didn't find any model when searching for it in the HF modelhub


alanakbik:

PR #2616, opened 2022-02-02
Bug import

marcelmmm:
Directories flair.models.diagnosis and flair.models.sequence_tagger_utils were not imported due to missing __init__.py
Solution: Add (empty) init file in both directories.

### DIFF
--- a/flair/models/lemmatizer_model.py
+++ b/flair/models/lemmatizer_model.py
@@ -564,7 +564,7 @@ class Lemmatizer(flair.nn.Classifier[Sentence]):
 
                         # reshape to vector of size (batch_size, beam_size*beam_size),
                         # each row contains beam_size*beam_size scores of the new possible hypothesis
-                        hypothesis_scores_per_token = hypothesis_scores.view(number_tokens, self.beam_size ** 2)
+                        hypothesis_scores_per_token = hypothesis_scores.view(number_tokens, self.beam_size**2)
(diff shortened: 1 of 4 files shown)


flair/models/lemmatizer_model.py
```diff
@@ -560,7 +562,8 @@ def predict(
                         hypothesis_scores = torch.cat(self.beam_size * [scores], dim=1) + log_probabilities
                         # print(hypothesis_scores)
 
-                        # reshape to vector of size (batch_size, beam_size*beam_size), each row contains beam_size*beam_size scores of the new possible hypothesis
+                        # reshape to vector of size (batch_size, beam_size*beam_size),
+                        # each row contains beam_size*beam_size scores of the new possible hypothesis
                         hypothesis_scores_per_token = hypothesis_scores.view(number_tokens, self.beam_size ** 2)
```
helpmefindaname, line 567:
```suggestion
                        hypothesis_scores_per_token = hypothesis_scores.view(number_tokens, self.beam_size**2)
```
@alanakbik  with the new release, of black, this line changes.
I see this is annoying when the build changes just because a dev library changes behaivour. I will think of a solution in the weekend.

alanakbik: Ah wow thanks for spotting this. I guess everyone then needs to update to the newest version of black.

PR #2620, opened 2022-02-04
biaffine_model

87302380:
Using the biaffine model as decoder to solve the NER problem. The code based on the SequenceTagger model. 
But the accuracy and precision of the model is very bad， the cause of the problem has not been found yet.

The code about biaffine model is referenced ：https://github.com/juntaoy/biaffine-ner


stefan-it:
Hi @87302380 ,

thanks for adding this! I can have a look at it :)

One question: have you already implemented the context functionality? For a given `t` the context for generating the embedding for `t` will be calculated from the `x` left (and right?) tokens, and `x` == 64 tokens? But I have to look at the paper again.

alanakbik:
Yes, setting `use_context` to True should do it!

whoisjones (commented):
Hey @87302380, cool PR thanks for that! Looking forward to integrate this. However, I have some comments:
1. Let's simplify your code such that your BiaffineTagger does only this and not also other things such as being an LSTM CRF ;) you can stick to this implementation and copy code from there but be sure your model is a BiaffineTagger and nothing else.
2. If this is done, let's get it running and try to reproduce the results of the original paper. I saw that the conflict handling part is a complicated one. Maybe skip it in the first iteration and think about about a better abstraction?
3. I also try to get it running this week, I already saw that in your Bilinear / Biaffine module, you use the same linear projection for start and end. Maybe this solves your issues already.

bratao:
@87302380 , any chance of this merging with flair? I got great results with Biaffine for NER and would be a great addition to Flair 😄

87302380:
> 
Thank you for your attention！I am working towards this, but it will take some time. :wink:

flair/models/sequence_tagger_utils/biaffine.py
```diff
@@ -0,0 +1,156 @@
...
+        shun                                                                                                                O      O       O   O
+        British                                                                                                                   MISC     O   O
+        lamb                                                                                                                               O   O
+        .                                                                                                                                      O
+        :return: 1-D  gold labels
+        """
+
+        gold_labels = []
+        for sentence in sentences:
+            ner = {(label.data_point.tokens[0].idx-1, label.data_point.tokens[-1].idx-1):label.value for label in sentences[0].get_labels("ner")}
```
helpmefindaname, line 130:
```suggestion
            ner = {(label.data_point.tokens[0].idx-1, label.data_point.tokens[-1].idx-1):label.value for label in sentence.get_labels("ner")}
```
I think this line contains a bug, which extracts labels of the first sentence for all sentences

87302380: You are right. Thank you very much！

Issue #2614, opened 2022-02-01
Loading a model that uses TransformerWordEmbeddings as embeddings，retraining it and saving will encounter a save error


87302380:
hey guys,  
I am testing my model with the latest development version. I used the following TransformerWordEmbeddings as the embedding layer of the model.
```
from flair.embeddings import TransformerWordEmbeddings

embeddings = TransformerWordEmbeddings(
    model='xlm-roberta-large',
    layers="-1",
    subtoken_pooling="first",
    fine_tune=True,
    use_context=True,
)
```
The training of the initial model everything is in order， until I load the saved model, and try to continue training, at the end of the training save the model, I get the error.
```
/usr/local/lib/python3.7/dist-packages/flair/nn/model.py in save(self, model_file, checkpoint)
    117 
    118         # save model
--> 119         torch.save(model_state, str(model_file), pickle_protocol=4)
    120 
    121         # restore optimizer and scheduler to model card if set

/usr/local/lib/python3.7/dist-packages/torch/serialization.py in save(obj, f, pickle_module, pickle_protocol, _use_new_zipfile_serialization)
    377         if _use_new_zipfile_serialization:
    378             with _open_zipfile_writer(opened_file) as opened_zipfile:
--> 379                 _save(obj, opened_zipfile, pickle_module, pickle_protocol)
    380                 return
    381         _legacy_save(obj, opened_file, pickle_module, pickle_protocol)

/usr/local/lib/python3.7/dist-packages/torch/serialization.py in _save(obj, zip_file, pickle_module, pickle_protocol)
    482     pickler = pickle_module.Pickler(data_buf, protocol=pickle_protocol)
    483     pickler.persistent_id = persistent_id
--> 484     pickler.dump(obj)
    485     data_value = data_buf.getvalue()
    486     zip_file.write_record('data.pkl', data_value, len(data_value))

/usr/local/lib/python3.7/dist-packages/flair/embeddings/base.py in __getstate__(self)
    374         config_dict = self.model.config.to_dict()
    375 
--> 376         tokenizer_data = self._tokenizer_bytes()
    377 
    378         model_state = {

/usr/local/lib/python3.7/dist-packages/flair/embeddings/base.py in _tokenizer_bytes(self)
    343     def _tokenizer_bytes(self):
    344         with tempfile.TemporaryDirectory() as temp_dir:
--> 345             files = self.tokenizer.save_pretrained(temp_dir)
    346             zip_data = BytesIO()
[...]

PR #2622, opened 2022-02-05
fix reloading fast tokenizers

helpmefindaname:
Solves https://github.com/flairNLP/flair/issues/2614
context:
The slow variant of tokenizers relies on sentencepiece which are saved in the respective files defined in `vocab_files_names`, when a fast tokenizer is created and will be saved, it creates a `tokenizer.json` file which basically contains all information, however it still tries to persist the `vocab_file_names` even if they are not required anymore. It cannot recreate the `vocab_file_names` without rebuilding the sentencepiece objects, so it rather tries to copy the file, if one is specified.
Excluding those files makes those not be specified → it won't try to copy the file

### DIFF
--- a/flair/embeddings/base.py
+++ b/flair/embeddings/base.py
@@ -340,7 +340,10 @@ class TransformerEmbedding(Embeddings[Sentence]):
 
     def _tokenizer_bytes(self):
         with tempfile.TemporaryDirectory() as temp_dir:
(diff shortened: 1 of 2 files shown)


alanakbik:
@helpmefindaname thanks a lot for fixing this!

87302380:
@alanakbik  @helpmefindaname 
Great, now it can be saved successfully！
thanks a lot for the fix.

PR #2627, opened 2022-02-09
Prototypical networks in Flair

alanakbik:
Thanks to the work of @plonerma, this PR adds a prototype decoder to Flair. 

Prototype networks learn prototypes for each target class. For each data point to be classified, the network predicts a vector in class-prototype-space, which is then compared to all class prototypes.The prediction is then the closest class prototype. See paper [Prototypical Networks for Few-shot Learning](https://arxiv.org/abs/1703.05175) for more info. 

@plonerma implemented a custom decoder that can be added to any Flair model that inherits from `DefaultClassifier` (i.e. early all Flair models). For instance, use this script: 

```python
from flair.data import Corpus
from flair.datasets import UP_ENGLISH
from flair.models import WordTagger
from flair.nn import PrototypicalDecoder

# what tag do we want to predict?
tag_type = 'frame'

# get a corpus
corpus: Corpus = UP_ENGLISH().downsample(0.1)

# make the tag dictionary from the corpus
tag_dictionary = corpus.make_label_dictionary(label_type=tag_type)

# initialize simple embeddings
embeddings = TransformerWordEmbeddings(model="distilbert-base-uncased",
                                       fine_tune=True,
                                       layers='-1')

# initialize prototype decoder
decoder = PrototypicalDecoder(num_prototypes=len(tag_dictionary),
                              embeddings_size=embeddings.embedding_length,
                              distance_function='euclidean',
                              normal_distributed_initial_prototypes=True,
                              )

# initialize the WordTagger, but pass the prototype decoder
tagger = WordTagger(embeddings,
                    tag_dictionary,
                    tag_type,
                    decoder=decoder)

# initialize trainer

# run training
trainer.fine_tune('resources/taggers/prototypical_decoder')
```

However, this is still in **beta** and will likely still change a lot until the next Flair release.


flair/models/pairwise_classification_model.py
```diff
@@ -150,5 +150,4 @@ def _init_model_with_state_dict(cls, state, **kwargs):
             else state["multi_label_threshold"],
             loss_weights=state["weight_dict"],
             embed_separately=state["embed_separately"],
-            **kargs,
```
plonerma: If a class inherits from TextPairClassifier, it might need to pass other kwargs ("kargs" is of course wrong).

alanakbik: good point!

Issue #2600, opened 2022-01-18
StackedEmbeddings modifies original embeddings' name which causes unwanted behaviour and potential crashes


tadejmagajna:
**Description**
When `StackedEmbeddings` are initialised, they modify the input embeddings by renaming their `name` property as part of this code:

https://github.com/flairNLP/flair/blob/579c7b70dfcc1d184bd47069f722d7a9ae8b78d7/flair/embeddings/token.py#L49-L50

The above code modifies the original embeddings' name. But several parts of Flair rely on embedding names being consistent.

For example the [`_everything_embedded()`](https://github.com/flairNLP/flair/blob/b774774752c8338aab3d620f7e5062f66ec7a69d/flair/embeddings/base.py#L67) method relies on embedding names when checking whether all embeddings were applied. When an embedding's name changes, its methods like `_everything_embedded()` become broken and this potentially results in the same embedding being added twice to the same token.

The adverse conesquences of this bug and the consequences follow this pattern:

1. The original embeddding's (let's call it `Char`) gets renamed to `0-Char` the first time we call StackedEmbeddings. If we pass the same embedding object to another StackedEmbeddings class it'll be renamed to `0-0-Char` and so on...
2. `_everything_embedded()` method breaks and **even if all tokens in a `Sentence` are embedded, the method will think they aren't** because the embedding now has a new name and no tokens were embedded with an embedding that has this new name. This results in the token being embedded again the next time we call `.embed()` on the same token resulting in duplicated embedding dimensionality.
3. The tokens are now embedded with double embedding dimensionality which causes a number of downstream issues and crashes

**To Reproduce**
```python
from flair.embeddings import StackedEmbeddings, CharacterEmbeddings
from flair.data import Sentence

# setup
s = Sentence('a')
e = CharacterEmbeddings(hidden_size_char=2)
e.embed(s)
print("name of the embedding is", e.name)  # name is "Char"
print("embedding len is:", len(s[0].embedding))  # embedding is 4-dimentional
e.embed(s)
print("embedding len is:", len(s[0].embedding))  # embedding is 4-dimentional regardless of how many times we call embed()

# trigger the bug
stack = StackedEmbeddings([e])
e.embed(s)

# unwanted results
print("name of the embedding is now:", e.name)  # name is "0-Char"
print("embedding len is now:", len(s[0].embedding))  # embedding is now 8-dimentional
```

**Expected behaviour**
[...]

PR #2633, opened 2022-02-14
Reintroduce hyperparameter tuning

### LINKED ISSUE #2632: Reintroduce hyperparameter tuning in Flair
Hyperparameter tuning used to be a feature in Flair<0.10. There's still some remains of this feature in [the documentation](https://github.com/flairNLP/flair/blob/master/resources/docs/KOR_docs/TUTORIAL_8_MODEL_OPTIMIZATION.md) though the code for it has been removed.

My understanding is that the feature was removed because noone was prepared to maintain it. Some users are still asking about why it was removed and are probably surprised to see it go: https://github.com/flairNLP/flair/pull/2518

The work required to get hyperparameter tuning working again with changes from Flair v0.10 is pretty minor.
I'd be happy to reintroduce hyperparameter tuning in Flair in its basic form.

tadejmagajna:
This PR reintroduces hyperparameter tuning in Flair
closes #2632

A number of small changes was required so I split this PR into appropriately named commits where each commit is an isolated change to the code base. For example, there is a separate commit that reverts the deletion or hyperparameter tuning in Flair - this helps us quickly see which code changes are new and which simply reintroduce the old code.

The changes and their corresponding commits include:
1. [Revert "Removes hyperparameter features"](https://github.com/flairNLP/flair/commit/c9591b9482afcd3971497fe88afb14bc394bd516)[](https://github.com/tadejmagajna)  Reverts the removal of all hyperparameter runing code
2. [Updating the param selection docs for the v0.10 syntax](https://github.com/flairNLP/flair/commit/32ebc4c13ddfcebf60a0c8487000d1359a497e86). This updates the README tutorial for English and Korean (I did my best with Korean here) to fit the new v0.10 compliant syntax.
3. [Adding hyperopt back to requirements.txt](https://github.com/flairNLP/flair/commit/f0c9f42a31468b4ee76497dd18a9d02f277b3479) adds the most recent version of hyperopt back to requirements
4. [Fixing paramselection code to work with changes in Flair v0.10](https://github.com/flairNLP/flair/commit/2995cc43153516f9709f58c33c449899dd7bf5f1) This updates the original hyperparameter syntax to work with changes applied in v0.10 such as, for example, the `make_label_dictionary(label_type)` requiring the `label_type` argument
5. [Fixing bug where embeddings got added twice on multiple training runs](https://github.com/flairNLP/flair/commit/0a140b8a7803c44228b4943a22dab3a67f6d87b8)[](https://github.com/tadejmagajna). This provides a temporary solution for #2600 where embeddings were added to tokens multiple times and this caused crashes (can be reproduced by running the TUTORIAL_8 code). Note that the fix probably does probably slow the execution down a bit, but it only affects runs where `training_runs>1`
6. [Enabling and fixing tests for param selection](https://github.com/flairNLP/flair/commit/a63f988a5db2c8471947d24f77cd302f803b5ed6). Here I reintroduce hyperparameter tests, apply some fixes to them and enable them again because they were previously disabled for some reason. There was also a deprecation warning related to some dependency of Hyperopt. To fix it I applied a change to to `pyproject.toml` where I swapped the `-W` parameter for the more-readable `filterwarnings` parameter and made it ignore that specific `DeprecationWarning`.
7. [Automatic code formatting of param selection tests](https://github.com/flairNLP/flair/commit/ebaa9fe4ca37f91964bc84c5c7244ba1a41c06e7) auto code formatting

Please do me know if there are any other parts of the hyperparameter tuning code I forgot to make compliant with the current Flair syntax

### DIFF
--- a/flair/hyperparameter/param_selection.py
+++ b/flair/hyperparameter/param_selection.py
@@ -15,7 +15,6 @@ from flair.hyperparameter.parameter import (
     SEQUENCE_TAGGER_PARAMETERS,
     TRAINING_PARAMETERS,
     DOCUMENT_EMBEDDING_PARAMETERS,
-    MODEL_TRAINER_PARAMETERS,
 )
 from flair.models import SequenceTagger, TextClassifier
 from flair.trainers import ModelTrainer
@@ -97,13 +96,8 @@ class ParamSelector(object):
             training_params = {
                 key: params[key] for key in params if key in TRAINING_PARAMETERS
             }
-            model_trainer_parameters = {
-                key: params[key] for key in params if key in MODEL_TRAINER_PARAMETERS
-            }
 
-            trainer: ModelTrainer = ModelTrainer(
-                model, self.corpus, **model_trainer_parameters
-            )
+            trainer: ModelTrainer = ModelTrainer(model, self.corpus)
 
             result = trainer.train(
                 self.base_path,
@@ -226,6 +220,7 @@ class TextClassifierParamSelector(ParamSelector):
     def __init__(
         self,
         corpus: Corpus,
+        label_type: str,
         multi_label: bool,
         base_path: Union[str, Path],
         document_embedding_type: str,
@@ -236,6 +231,7 @@ class TextClassifierParamSelector(ParamSelector):
     ):
         """
         :param corpus: the corpus
+        :param label_type: string to identify the label type ('question_class', 'sentiment', etc.)
         :param multi_label: true, if the dataset is multi label, false otherwise
         :param base_path: the path to the result folder (results will be written to that folder)
         :param document_embedding_type: either 'lstm', 'mean', 'min', or 'max'
@@ -254,9 +250,10 @@ class TextClassifierParamSelector(ParamSelector):
         )
 
         self.multi_label = multi_label
+        self.label_type = label_type
         self.document_embedding_type = document_embedding_type
 
-        self.label_dictionary = self.corpus.make_label_dictionary()
+        self.label_dictionary = self.corpus.make_label_dictionary(self.label_type)
 
     def _set_up_model(self, params: dict):
         embdding_params = {
@@ -264,13 +261,14 @@ class TextClassifierParamSelector(ParamSelector):
         }
 
         if self.document_embedding_type == "lstm":
-            document_embedding = DocumentRNNEmbeddings(**embdding_params)
+            document_embedding = DocumentRNNEmbeddings(rnn_type="LSTM", **embdding_params)
         else:
             document_embedding = DocumentPoolEmbeddings(**embdding_params)
 
         text_classifier: TextClassifier = TextClassifier(
             label_dictionary=self.label_dictionary,
             multi_label=self.multi_label,
+            label_type=self.label_type,
             document_embeddings=document_embedding,
         )
 
--- a/flair/hyperparameter/param_selection.py
+++ b/flair/hyperparameter/param_selection.py
@@ -81,15 +81,15 @@ class ParamSelector(object):
             log.info(f"\t{k}: {str(v)}")
         log_line(log)
 
-        for sent in self.corpus.get_all_sentences():
-            sent.clear_embeddings()
-
         scores = []
         vars = []
 
         for i in range(0, self.training_runs):
             log_line(log)
             log.info(f"Training run: {i + 1}")
+
+            for sent in self.corpus.get_all_sentences():
+                sent.clear_embeddings()
 
             model = self._set_up_model(params)
 
--- a/flair/hyperparameter/param_selection.py
+++ b/flair/hyperparameter/param_selection.py
@@ -1,0 +1,277 @@
+import logging
+from abc import abstractmethod
+from enum import Enum
+from pathlib import Path
+from typing import Tuple, Union
+import numpy as np
+
+from hyperopt import hp, fmin, tpe
+
+import flair.nn
+from flair.data import Corpus
+from flair.embeddings import DocumentPoolEmbeddings, DocumentRNNEmbeddings
+from flair.hyperparameter import Parameter
+from flair.hyperparameter.parameter import (
+    SEQUENCE_TAGGER_PARAMETERS,
+    TRAINING_PARAMETERS,
+    DOCUMENT_EMBEDDING_PARAMETERS,
+    MODEL_TRAINER_PARAMETERS,
+)
+from flair.models import SequenceTagger, TextClassifier
+from flair.trainers import ModelTrainer
+from flair.training_utils import (
+    EvaluationMetric,
+    log_line,
+    init_output_file,
+    add_file_handler,
+)
+
+log = logging.getLogger("flair")
+
+
+class OptimizationValue(Enum):
+    DEV_LOSS = "loss"
+    DEV_SCORE = "score"
+
+
+class SearchSpace(object):
+    def __init__(self):
+        self.search_space = {}
+
+    def add(self, parameter: Parameter, func, **kwargs):
+        self.search_space[parameter.value] = func(parameter.value, **kwargs)
+
+    def get_search_space(self):
+        return hp.choice("parameters", [self.search_space])
+
+
+class ParamSelector(object):
+    def __init__(
+        self,
+        corpus: Corpus,
+        base_path: Union[str, Path],
+        max_epochs: int,
+        evaluation_metric: EvaluationMetric,
+        training_runs: int,
+        optimization_value: OptimizationValue,
+    ):
+        if type(base_path) is str:
+            base_path = Path(base_path)
+
+        self.corpus = corpus
+        self.max_epochs = max_epochs
+        self.base_path = base_path
+        self.evaluation_metric = evaluation_metric
+        self.run = 1
+        self.training_runs = training_runs
+        self.optimization_value = optimization_value
+
+        self.param_selection_file = init_output_file(base_path, "param_selection.txt")
+
+    @abstractmethod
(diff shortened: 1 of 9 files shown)


flair/hyperparameter/param_selection.py
```diff
@@ -0,0 +1,260 @@
...
+        self.multi_label = multi_label
+        self.label_type = label_type
+        self.document_embedding_type = document_embedding_type
+
+        self.label_dictionary = self.corpus.make_label_dictionary(self.label_type)
+
+    def _set_up_model(self, params: dict):
+        embdding_params = {key: params[key] for key in params if key in DOCUMENT_EMBEDDING_PARAMETERS}
+
+        if self.document_embedding_type == "lstm":
```
alanakbik, line 248: it seems the only options are RNN or pooled word embeddings, but the standard approach today would be to simply use a transformer only which far outperforms RNNs trained over word embeddings. Not necessarily something that needs to be fixed in this PR but something that limits the usefulness of our parameter selection for text classifiers at the moment.

flair/hyperparameter/parameter.py
```diff
@@ -0,0 +1,66 @@
...
+    Parameter.HIDDEN_SIZE.value,
+    Parameter.RNN_LAYERS.value,
+    Parameter.USE_CRF.value,
+    Parameter.USE_RNN.value,
+    Parameter.DROPOUT.value,
+    Parameter.LOCKED_DROPOUT.value,
+    Parameter.WORD_DROPOUT.value,
+]
+DOCUMENT_EMBEDDING_PARAMETERS = [
+    Parameter.EMBEDDINGS.value,
```
alanakbik, line 57: one idea for the problem above would be to support param selection only for `TransformerDocumentEmbeddings` and define some parameters that one would like to search over (which transformer model, which layers, fine-tune yes/no) and remove the RNN specific ones.

resources/docs/TUTORIAL_8_MODEL_OPTIMIZATION.md
```diff
@@ -0,0 +1,156 @@
...
+[hyperopt](https://github.com/hyperopt/hyperopt).
+
+First you need to load your corpus. If you want to load the [AGNews corpus](https://www.di.unipi.it/~gulli/AG_corpus_of_news_articles.html)
+used in the following example, you first need to download it and convert it into the correct format. Please
+check [tutorial 6](/resources/docs/TUTORIAL_6_CORPUS.md) for more details.
+```python
+from flair.datasets import TREC_6
+
+# load your corpus
+corpus = TREC_6()
```
alanakbik, line 18: The tutorial is a bit counterintuitive since it defines a text classification corpus and then word embeddings. The fact that there's an LSTM used is buried in the model initialization and not explained.

Perhaps switch to a sequence labeling example? `WNUT_17` for instance could be used as example corpus, and for sequence labeling the current implementation allows some interesting parameter searches.

resources/docs/TUTORIAL_8_MODEL_OPTIMIZATION.md
```diff
@@ -0,0 +1,156 @@
...
+
+# 1. get the corpus
+corpus = WNUT_17().downsample(0.1)
+print(corpus)
+
+# 2. what tag do we want to predict?
+tag_type = 'ner'
+
+# 3. make the tag dictionary from the corpus
+tag_dictionary = corpus.make_tag_dictionary(tag_type=tag_type)
```
alanakbik, line 121: should be `tag_dictionary = corpus.make_label_dictionary(label_type=tag_type)`, otherwise the example won't run

resources/docs/TUTORIAL_8_MODEL_OPTIMIZATION.md
```diff
@@ -0,0 +1,156 @@
...
+                                        embeddings=embeddings,
+                                        tag_dictionary=tag_dictionary,
+                                        tag_type=tag_type,
+                                        use_crf=True)
+
+# 6. initialize trainer
+trainer: ModelTrainer = ModelTrainer(tagger, corpus)
+
+# 7. find learning rate
+learning_rate_tsv = trainer.find_learning_rate('resources/taggers/example-ner', Adam)
```
alanakbik, line 144: this function is broken, it should be changed to the code below, ideally as part of this PR.

```python
    def find_learning_rate(
        self,
        base_path: Union[Path, str],
        optimizer,
        file_name: str = "learning_rate.tsv",
        start_learning_rate: float = 1e-7,
        end_learning_rate: float = 10,
        iterations: int = 100,
        mini_batch_size: int = 32,
        stop_early: bool = True,
        smoothing_factor: float = 0.98,
        **kwargs,
    ) -> Path:
        best_loss = None
        moving_avg_loss = 0

        # cast string to Path
        if type(base_path) is str:
            base_path = Path(base_path)
        learning_rate_tsv = init_output_file(base_path, file_name)

        with open(learning_rate_tsv, "a") as f:
            f.write("ITERATION\tTIMESTAMP\tLEARNING_RATE\tTRAIN_LOSS\n")

        optimizer = optimizer(self.model.parameters(), lr=start_learning_rate, **kwargs)

        train_data = self.corpus.train

        scheduler = ExpAnnealLR(optimizer, end_learning_rate, iterations)

        model_state = self.model.state_dict()
        self.model.train()

        step = 0
        while step < iterations:
            batch_loader = DataLoader(train_data, batch_size=mini_batch_size, shuffle=True)
            for batch in batch_loader:
                step += 1

                # forward pass
                loss = self.model.forward_loss(batch)
                if isinstance(loss, tuple):
                    loss = loss[0]

                # update optimizer and scheduler
                optimizer.zero_grad()
                loss.backward()
                torch.nn.utils.clip_grad_norm_(self.model.parameters(), 5.0)
                optimizer.step()
                scheduler.step(step)

                learning_rate = scheduler.get_lr()[0]

                loss_item = loss.item()
                if step == 1:
                    best_loss = loss_item
                else:
                    if smoothing_factor > 0:
                        moving_avg_loss = smoothing_factor * moving_avg_loss + (1 - smoothing_factor) * loss_item
                        loss_item = moving_avg_loss / (1 - smoothing_factor ** (step + 1))
                    if loss_item < best_loss:
                        best_loss = loss

                if step > iterations:
                    break

                if stop_early and (loss_item > 4 * best_loss or torch.isnan(loss)):
                    log_line(log)
                    log.info("loss diverged - stopping early!")
                    step = iterations
                    break

                with open(str(learning_rate_tsv), "a") as f:
                    f.write(f"{step}\t{datetime.datetime.now():%H:%M:%S}\t{learning_rate}\t{loss_item}\n")

            self.model.load_state_dict(model_state)
            self.model.to(flair.device)

        log_line(log)
        log.info(f"learning rate finder finished - plot {learning_rate_tsv}")
        log_line(log)

        return Path(learning_rate_tsv)
```

alanakbik (changes requested):
@tadejmagajna Thanks for bringing this back! But some issues regarding the examples, see comments.

flair/hyperparameter/param_selection.py
tadejmagajna: As per your suggestion (which makes perfects sense - it felt really weird readding this outdated RNN code) I refactored `TextClassifierParamSelector` so that it uses `TransformerDocumentEmbeddings` now. I documented it to a reasonable extent in the tutorial.

flair/hyperparameter/parameter.py
tadejmagajna: Did just that and refactored `TextClassifierParamSelector` so that it uses `TransformerDocumentEmbeddings`. Also renamed `DOCUMENT_EMBEDDING_PARAMETERS ` to `TEXT_CLASSIFICATION_PARAMETERS`

resources/docs/TUTORIAL_8_MODEL_OPTIMIZATION.md
tadejmagajna: I agree - makes sense to give precedence to sequence labelling in Flair.

As per your suggestion I added an example to the top that uses `WNUT_17 ` and documents the use of `SequenceTaggerParamSelector `

However, I also kept the example that shows how to use `TextClassifierParamSelector`. I feel like that if a feature isn't documented, people are unlikely to dig through the source and the value of the feature will be diminished.

So what I did was I created two sections (`Selecting hyperparameters for sequence labelling` and `Selecting hyperparameters for text classification`) and documented both use-cases accordingly.

resources/docs/TUTORIAL_8_MODEL_OPTIMIZATION.md
tadejmagajna: I addressed this as part of this PR, but note that I made tiny adjustment to the source code you pasted above. With the above code, we were getting the following warining from Torch:
> UserWarning: The epoch parameter in `scheduler.step()` was not necessary and is being deprecated where possible. Please use `scheduler.step()` to step the scheduler.

this warning broke the CI build.

So I replaced the line `scheduler.step(step)` with  `scheduler.step()`. If my understanding is correct, this shouldn't badly affect the execution of the code. But if it does, please do let me know.

tadejmagajna:
@alanakbik Thank you for the feedback!

I addressed your PR comments and I incorporated all of the suggestions including the refactoring of `TextClassifierParamSelector` so that it uses `TransformerDocumentEmbeddings` (instead of RNN) as well as fixing of the `find_learning_rate()`. I left more detailed comments in each individual PR feedback thread.

tadejmagajna:
@alanakbik I made `fune_tune` a normal (non-tunable) param now and set it to default to True. Updated all the docs as well

alanakbik (approved):

alanakbik:
@tadejmagajna thanks a lot - looks great!

PR #2641, opened 2022-02-21
Allow trainer to work with mutliple learning rates

plonerma:
This PR enables the train function of ModelTrainer to work with multiple learning rates.

Prior, the train function assumed the usage of a single learning rate. When using a LR schedule (like OneCycleLR), group specific learning rates were overwritten.
With this PR, it is possible to pass an optimizer instance with multiple parameter groups and learning rates and still use scheduling without overwriting group specific learning rates.

See [the pytorch documentation](https://pytorch.org/docs/1.10/optim.html#per-parameter-options) for details on how to initialize an optimizer with multiple learning rates.


alanakbik:

Issue #2640, opened 2022-02-20
ONNX compatible models


helpmefindaname:
To be removed, once it is done: Please add the appropriate label to this ticket, e.g. feature or enhancement.

**Is your feature/enhancement request related to a problem? Please describe.**
ONNX support is a frequently requested feature, some issues mention it (https://github.com/flairNLP/flair/issues/2625, https://github.com/flairNLP/flair/issues/2451, https://github.com/flairNLP/flair/issues/2317, https://github.com/flairNLP/flair/issues/1936, https://github.com/flairNLP/flair/issues/1423, https://github.com/flairNLP/flair/issues/999)
so I think there is a big desire for the community to support it.
I suppose the usual ONNX compatibility would also make the models compatible to torch.jit (https://github.com/flairNLP/flair/issues/2528) or AWS Neutron (https://github.com/flairNLP/flair/issues/2443)

ONNX provides large enhancements in terms of production readiness, it creates a static computational graph which can be quantized and optimized towards specific hardware, see https://onnxruntime.ai/docs/performance/tune-performance.html (it claims to be 17x faster)

**Describe the solution you'd like**
I'd suggest iterative progression as multiple architecture changes are required:

1. split the `forward`/`forward_pass` methods, such that all models have a method `_prepare_tensors` which converts all DataPoints to tensors and a `forward` which takes in tensors and outputs tensors (e.g. for the SeqeuenceTagger we the forward has the signature `def forward(self, sentence_tensor: torch.Tensor, lengths: torch.LongTensor)` and returns a single tensor `scores`)
this change allows conversion to ONNX models, however the logic (like decoding crf scores, filling up sentence results, extracting tensors) won't be implemented. Also embeddings won't be part of the ONNX model.
2. create the same `forward`/`_prepare_tensors` architecture for embeddings, such that those could be converted too.
This would allow converting embeddings to ONNX models, but again without logic.
3. change the architecture, that both embeddings and models have the logic part (creating inputs, adding outputs to data points) and the pytorch part be split, such that the pytorch model part can be replaced by a converted ONNX model.
4. create an end-to-end model wrapper, that both embeddings & the model can be converted to a single ONNX model and used as such.

Notice that this would be **4 different PRs** and probably all of them would be very large and should be tested a lot before moving to the next PR,
I would offer to do the first one and then see how much effort this is/how much time I have for this.

kieron-guinamard-privitar:
[...]

PR #2643, opened 2022-02-21
GH-2640: tensor forward

helpmefindaname:
First PR about https://github.com/flairNLP/flair/issues/2640: refactoring models, such that each model has a `_prepare_tensors` and a `forward` method, where one extracts all tensors out of the data points and the latter only does tensor computations.

That way, JIT tracing and ONNX conversation of the models should be possible.
**Notice:** this is not expected to give a huge speedup, as most computation is within the embeddings, which won't be affected. Due to that, the ONNX conversion is also not documented. 

This PR also adds unit scaling for model downloads when a non-huggingface model is downloaded.
This PR also fixes a bug, that relation extraction models without weight_dict set can be loaded.
This PR also fixes an encoding error, if glue-mnli is loaded on a windows machine
This PR also adds an option to not add a unk token to labels.

### DIFF
--- a/flair/models/entity_linker_model.py
+++ b/flair/models/entity_linker_model.py
@@ -91,7 +91,7 @@ class EntityLinker(flair.nn.DefaultClassifier[Sentence, Span]):
         return bool(data_point.get_labels(self.label_type))
 
     def _embed_prediction_data_point(self, prediction_data_point: Span) -> torch.Tensor:
-        return self.aggregated_embedding(prediction_data_point, self.word_embeddings.get_names()).unsqueeze(0)
+        return self.aggregated_embedding(prediction_data_point, self.word_embeddings.get_names())
 
     def _get_state_dict(self):
         model_state = {
--- a/flair/models/entity_linker_model.py
+++ b/flair/models/entity_linker_model.py
@@ -1,11 +1,11 @@
 import logging
-from typing import Callable, Dict, List, Tuple
+from typing import Callable, Dict, List
 
 import torch
 
 import flair.embeddings
 import flair.nn
-from flair.data import DataPoint, Dictionary, Sentence, Span
+from flair.data import Dictionary, Sentence, Span
 
 log = logging.getLogger("flair")
 
--- a/flair/models/entity_linker_model.py
+++ b/flair/models/entity_linker_model.py
@@ -6,6 +6,7 @@ import torch
 import flair.embeddings
 import flair.nn
 from flair.data import Dictionary, Sentence, Span
+from flair.embeddings import Embeddings
 
 log = logging.getLogger("flair")
 
@@ -38,7 +39,6 @@ class EntityLinker(flair.nn.DefaultClassifier[Sentence, Span]):
 
         super(EntityLinker, self).__init__(
             label_dictionary=label_dictionary,
-            embeddings=word_embeddings,
             final_embedding_size=word_embeddings.embedding_length * 2
             if pooling_operation == "first_last"
             else word_embeddings.embedding_length,
@@ -76,6 +76,10 @@ class EntityLinker(flair.nn.DefaultClassifier[Sentence, Span]):
 
     def emb_mean(self, span, embedding_names):
         return torch.mean(torch.cat([token.get_embedding(embedding_names) for token in span], 0), 0)
+
+    @property
+    def _inner_embeddings(self) -> Embeddings[Sentence]:
+        return self.word_embeddings
 
     def _get_prediction_data_points(self, sentences: List[Sentence]) -> List[Span]:
         entities: List[Span] = []
--- a/flair/models/entity_linker_model.py
+++ b/flair/models/entity_linker_model.py
@@ -117,9 +117,9 @@ class EntityLinker(flair.nn.DefaultClassifier[Sentence]):
         if len(embedding_list) > 0:
             embedded_entity_pairs = torch.cat(embedding_list, 0)
 
-            return embedded_entity_pairs,
+            return (embedded_entity_pairs,)
         else:
-            return torch.zeros(0, self.word_embeddings.embedding_length, device=flair.device),
+            return (torch.zeros(0, self.word_embeddings.embedding_length, device=flair.device),)
 
     def _get_state_dict(self):
         model_state = {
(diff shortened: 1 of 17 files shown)


helpmefindaname:
Already working is the TextClassifier, see the following example code:
```
import torch


def convert():
    model = TextClassifier.load("de-offensive-language")
    example_sentence = Sentence("This is a sentence.")

    tensors = model._prepare_tensors([example_sentence])

    torch.onnx.export(model, tensors, "texclassifier.onnx", input_names=["text_embedding_tensor"], output_names=["scores"], opset_version=7, verbose=True)


if __name__ == "__main__":
    convert()
```

bratao:
Any chance of this making into main tree? I love flair, and ONNX models are a must for production code!

alanakbik:
hello @bratao yes we are looking into this now - sorry @helpmefindaname for taking so long, I first wanted to release Flair 0.11 (done yesterday) before making bigger changes.

flair/nn/model.py
```diff
@@ -582,42 +613,19 @@ def multi_label_threshold(self, x):  # setter method
         else:
             self._multi_label_threshold = {"default": x}
 
-    def forward_loss(self, sentences: Union[List[DT], DT]) -> Tuple[torch.Tensor, int]:
-
-        # make a forward pass to produce embedded data points and labels
-        embedded_data_points, labels = self.forward_pass(sentences)  # type: ignore
-
-        # no loss can be calculated if there are no labels
-        if not any(labels):
```
alanakbik, line 591: This check is currently important for models that require candidates like the EntityLinker or RelationExtractor.

alanakbik:
@helpmefindaname I started reviewing, but it will take some time to think through all the changes. 

The logic of splitting out the tensor and non-tensor stuff in the forward pass makes sense for ONNX but I worry about code readability with the logic now distributed across many methods and different parent classes (`forward_loss` now calls `forward` which in turn calls `forward_pass`) and the risk of tensor and label preparation diverging. Before with the `forward_pass` there was a single method that encapsulates all model-specific logic for each class that inherits from DefaultClassifier, albeit with the drawback of an overly complex return type. I wonder if the structure can be somehow more simplified but have to think about this for a bit. 

There also is a small problem with models that require candidates and get passed sentences without candidates, see:

```python
# init a random linker for testing
linker: EntityLinker = EntityLinker(TransformerWordEmbeddings(model='distilbert-base-uncased'),
                                    label_dictionary=Dictionary())

# sentence with candidate label - works
sentence = Sentence("I live in Berlin")
sentence[3:4].add_label('nel', 'LOC')
linker.predict(sentence)
print(sentence)

# sentence without candidate - fails
sentence = Sentence("I live in Berlin")
linker.predict(sentence) 
```

alanakbik:
Hello @helpmefindaname from a first look-through I like this structure a lot!

Some suggestions: 
- `forward_pass` could be renamed into something that communicates that this is now only _optional_ intermediary layers between the embeddings and the decoder.  (Since it always takes an embedding Tensor as input and the output goes to the decoder.) Maybe something that does not contain "forward" to avoid confusion with "forward_loss" and "forward". Unfortunately, I cannot come up with a good name now... 
- `DefaultClassifier` could have a standard implementation of  `_get_label_of_datapoint` (i.e. make it non-abstract). In most (all?) cases `return [datapoint.get_label(self._label_type).value]` should already do the trick, I think. That would be one less method to implement by default. 

That would reduce the code needed to 6 methods for a standard default classifier with no extras (for instance the `TextPairClassifier`): 

1. an `__init__`
2. a `_get_prediction_data_points` 
3. a `_embed_prediction_data_point` 
4. a `label_type` property
5. & 6. and the two load/save methods `_get_state_dict` and `_init_model_with_state_dict`

with all actual important logic in the first three methods, which is cool.

helpmefindaname:
A small change to the interface:
Instead of providing the embedding as a parameter, I had to create a property for it, to not double the amount of parameters per model. Also the embedding is optional so `TextPairClassifier` can choose to not have pre-embedded sentences but do the embedding on its own.

The RelationExtractor logic could be simplified, as the logic for extracting labels was the same as the logic created in `_PartOfSentence` thus `Relation(span1, span2).get_label(self.label_type)` was sufficient.

flair/nn/model.py
```diff
@@ -628,7 +673,24 @@ def _calculate_loss(self, scores, labels) -> Tuple[torch.Tensor, int]:
...
+
+        # make a forward pass to produce embedded data points and labels
+        predict_data_points = self._get_prediction_data_points(sentences)
+        labels = self._prepare_label_tensor(predict_data_points)
+
+        # no loss can be calculated if there are no labels
+        if labels.size(0) == 0:
+            return torch.tensor(0.0, requires_grad=True, device=flair.device), 1
+
+        embedded_tensor = self._prepare_tensors(sentences)
```
alanakbik, line 686: This could cause some problems I think. 4 lines above the `predict_data_points` are extracted using `self._get_prediction_data_points(sentences)` and labels extracted from them. 

But here now the `sentences` are passed into `self._prepare_tensors` which first applies a filter, then again calls `self._get_prediction_data_points(sentences)`. So it is possible that the datapoints extracted after the filter diverge from the datapoints used to get the labels. 

Maybe this `_prepare_tensors` could take as input `predict_data_points` instead of `sentences`? The filtering could be done beforehand as first line in `forward_loss`.

flair/models/entity_linker_model.py
```diff
@@ -62,75 +63,35 @@ def __init__(
...
+        return entities
 
-        if for_prediction:
-            return embedded_entity_pairs, span_labels, data_points
+    def _filter_data_point(self, data_point: Sentence) -> bool:
+        return bool(data_point.get_labels(self.label_type))
 
-        return embedded_entity_pairs, span_labels
+    def _embed_prediction_data_point(self, prediction_data_point: Span) -> torch.Tensor:
+        return self.aggregated_embedding(prediction_data_point, self.word_embeddings.get_names()).unsqueeze(0)
```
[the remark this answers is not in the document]
helpmefindaname: fixed and added a test

flair/nn/model.py
helpmefindaname: Very good catch!
Although the filtering is usually for sentences that do not have any labels, it could be misused and create some divergence.

I moved the filtering to be the first thing to be done to a batch. Even before calling `_prepare_tensors`.

Sadly, I cannot change the signature of `_prepare_tensors`, as otherwise it wouldn't be in line with the general `flair.nn.Model` and would make it very complicated to use jit or onnx exports.

alanakbik:
@helpmefindaname thanks for fixing the rebase conflicts!

alanakbik (approved):
Thanks a lot for adding this to Flair!

PR #2651, opened 2022-03-03
🐛 Fix .pre-commit-config.yaml

AnotherStranger:
While trying to set up pre-commit, I got an indentation error.
Moreover, pycqa/isort does not have a stable rev. I set it to the most recent release tag.


.pre-commit-config.yaml
```diff
@@ -1,11 +1,11 @@
...
+  - repo: https://github.com/ambv/black
     rev: stable
     hooks:
-    - id: black
-      language_version: python3.6 
+      - id: black
+        language_version: python3.6 
   - repo: https://github.com/pycqa/isort
-    rev: stable
+    rev: 5.10.1
```
tadejmagajna, line 8: Minor note: My only concern here is that you're pinning `isort` to `5.10.1` here whereas in the `Pipfile` we take the `*` version from PyPi.

This could mean that once isort 6 is released on PyPi, the two isort verisons (the one used by Pipfile and the one used by pre-commit) will become incompatible and start reporting different suggestions.

It may be worth pinning isort to `~= 5.10` (or `5.*` - same thing) in the Pipfile to make sure the two never get out of sync and start flagging different sort issues

tadejmagajna (approved):
Looks and works great!

One very minor suggestion though I don't think it should be a blocker for merge

alanakbik:
@AnotherStranger thanks for fixing this!

Issue #2554, opened 2021-12-15
 PicklingError: Could not serialize object: AttributeError: 'LanguageModel' object has no attribute 'document_delimiter'


MichSteczko:
Hello, 
I tried to use Flair with PySpark, but when I am using UDF (user defined function) to iterate over column to extract dates from text, I am getting the following error:
 PicklingError: Could not serialize object: AttributeError: 'LanguageModel' object has no attribute 'document_delimiter'.
 My code:
`from flair.data import Sentence`
`from flair.models import SequenceTagger`
`from pyspark.sql.functions import udf`
 `tagger = SequenceTagger.load('ner-ontonotes-fast') `

 `def extract_dates(sent):  ` 
   `    sentence = Sentence(sent)`
    `   tagger.predict(sentence)`
    `   return [entity.text if entity.tag == 'DATE' and entity.score >= 0.7 else None for entity in sentence.get_spans('ner') ]`
`transform_udf = udf(lambda x: extract_dates(x), ArrayType(StringType()))
df.withColumn('Dates', transform_udf(col('Text')))`

Can I fix this error or is this just a bug?

Tarotis:
I receive a similar error, but in my case while training a (custom) corpus and calling the trainer.train() function:

```
corpus = UD_ENGLISH().downsample(0.01)
tagger = SequenceTagger.load("pos-multi-fast")
trainer = ModelTrainer(tagger, corpus)
trainer.train('resources/taggers/example-upos', learning_rate=0.1, mini_batch_size=8, max_epochs=20)
```

Error is found in: `/flair/trainers/trainer.py", line 711, in train
    self.model.save(base_path / "best-model.pt", checkpoint=save_optimizer_state)`
Error message: `torch.nn.modules.module.ModuleAttributeError: 'LanguageModel' object has no attribute 'document_delimiter'`

The problem is reproduced when using the UD_ENGLISH() corpus instead of my custom data.

Package versions as followed: `flair: 0.10` and `torch: 1.9.1`, also tried `1.7.1` for torch and `0.9` for flair.

rayamoh:
Hi @MichSteczko - Hope all is well. Wondering what was the performance with pyspark FlairNLP. We are running into performance issues , for 1 record it runs for 7 mins. Spark cluster (64 cores , 512GB memory). Can you please share your experience. Thanks!

Issue #2503, opened 2021-11-04
AttributeError: 'LanguageModel' object has no attribute 'document_delimiter'


sgraaf:
**Describe the bug**
Hi! I am trying to "fine-tune" an existing, pre-trained `SequenceTagger` model (specifically `flair/ner-english-ontonotes-fast`) on a proprietary, custom `ColumnCorpus` with completely different entity labels. Inspired by a comment in #1540, for this I manipulate the pre-trained model's `state`, after which I run the training using the custom `Corpus`. 

After training for `X` epochs (I set `X` to 10), when the trainer saves the `final-model.pt`, I am faced with the following error:
```
[...]
2021-11-04 16:10:35,817 EPOCH 10 done: loss 0.2134 - lr 0.0100000
2021-11-04 16:10:36,329 DEV : loss 0.2844020426273346 - f1-score (micro avg)  0.0
2021-11-04 16:10:36,338 BAD EPOCHS (no improvement): 0
Traceback (most recent call last):
  File "C:\Users\sgraaf\Documents\GitHub\NER-Project\src\fine-tune.py", line 116, in <module>
    main()
  File "C:\Users\sgraaf\Documents\GitHub\NER-Project\src\fine-tune.py", line 107, in main
    trainer.train(
  File "C:\Users\sgraaf\Documents\GitHub\NER-Project\.venv\lib\site-packages\flair\trainers\trainer.py", line 690, in train
    self.model.save(base_path / "final-model.pt")
  File "C:\Users\sgraaf\Documents\GitHub\NER-Project\.venv\lib\site-packages\flair\nn\model.py", line 85, in save
    torch.save(model_state, str(model_file), pickle_protocol=4)
  File "C:\Users\sgraaf\Documents\GitHub\NER-Project\.venv\lib\site-packages\torch\serialization.py", line 379, in save
    _save(obj, opened_zipfile, pickle_module, pickle_protocol)
  File "C:\Users\sgraaf\Documents\GitHub\NER-Project\.venv\lib\site-packages\torch\serialization.py", line 484, in _save
    pickler.dump(obj)
  File "C:\Users\sgraaf\Documents\GitHub\NER-Project\.venv\lib\site-packages\flair\models\language_model.py", line 414, in __getstate__
    "document_delimiter": self.document_delimiter,
  File "C:\Users\sgraaf\Documents\GitHub\NER-Project\.venv\lib\site-packages\torch\nn\modules\module.py", line 1130, in __getattr__
[...]

PR #2658, opened 2022-03-07
fix: patch the missing `document_delmiter` for `lm.__get_state__()`

tianjianjiang:
close: #2503
close: #2554

There are already similar special treatments like https://github.com/flairNLP/flair/blob/master/flair/embeddings/token.py#L724 and https://github.com/flairNLP/flair/blob/master/flair/models/language_model.py#L190, but we will still need this for pre-trained `SequenceTagger` or alike to be pickle-able (for Dill and/or multiprocessing).

### DIFF
--- a/flair/models/language_model.py
+++ b/flair/models/language_model.py
@@ -393,8 +393,8 @@ class LanguageModel(nn.Module):
         return perplexity
 
     def __getstate__(self):
-        if "document_delimiter" not in self.__dict__:
-            self.document_delimiter = "\n"
+        # "document_delimiter" property may be missing in some older pre-trained models
+        self.document_delimiter = getattr(self, "document_delimiter", "\n")
 
         # serialize the language models and the constructor arguments (but nothing else)
         model_state = {
--- a/flair/models/language_model.py
+++ b/flair/models/language_model.py
@@ -393,6 +393,8 @@ class LanguageModel(nn.Module):
         return perplexity
 
     def __getstate__(self):
+        if "document_delimiter" not in self.__dict__:
+            self.document_delimiter = "\n"
 
         # serialize the language models and the constructor arguments (but nothing else)
         model_state = {


flair/models/language_model.py
```diff
@@ -393,6 +393,8 @@ def calculate_perplexity(self, text: str) -> float:
         return perplexity
 
     def __getstate__(self):
+        if "document_delimiter" not in self.__dict__:
+            self.document_delimiter = "\n"
```
tadejmagajna, line 397:
```suggestion
        # "document_delimiter" property may be missing in some older pre-trained models
        self.document_delimiter = getattr(self, "document_delimiter", "\n")
```

tadejmagajna (changes requested):
Thank you for your time and for figuring out the solution for this not-so-uncommon problem!

I did some investigation and found out that the root of this problem. The cause of this issue
is that some older serialized `LanguageModel`s don't have the `document_delimiter` property because it didn't exist at the time those models were trained.

For clarity I therefore suggest adding a small comment that explains the reason this workaround is needed.

Also, I suggest the use of `getattr` which is a bit more straight-forward and cleaner (no if statement needed) than the use of `self.__dict__`


Happy to approve otherwise!

alanakbik:
@tianjianjiang thanks for fixing this!

tianjianjiang:
I'm sorry for letting this falling through the cracks.
Really appreciate the review, @tadejmagajna, and I'm gonna provide some context just in case.
And of course, @alanakbik, thank you for the approval.

> For clarity I therefore suggest adding a small comment that explains the reason this workaround is needed.

BLUF: I'm all for it and pardon me for being too late to add that comment.
Context: Besides me been being lazy and careless, I had a vague hesitance because no comments with https://github.com/flairNLP/flair/blob/master/flair/embeddings/token.py#L724 and https://github.com/flairNLP/flair/blob/master/flair/models/language_model.py#L190, such that I was just hoping it could be an excuse to leave no comments about this (since I didn't carefully check whether flair has a convention of inline comment).


> Also, I suggest the use of `getattr` which is a bit more straight-forward and cleaner (no if statement needed) than the use of `self.__dict__`

BLUF: It crossed my mind but I would also like to change other near-identical lines. Yet I didn't want to change too many lines.
Context: I do agree that `getattr` is good for brevity, and I also think it's nice to respect the descriptor protocol. However, I am still unsure the intent behind other pre-existed, near-identical lines, e.g., what if those lines want to fail when a class doesn't have `__dict__`.


Again, thank you guys for all the support. Cheers!

Issue #2654, opened 2022-03-03
printing and logging inconsistencies


Xabilahu:
**Describe the bug**
There are several inconsistencies between printing with the native `print` python function and printing with `logging.getLogger('flair')`.

For example:
https://github.com/flairNLP/flair/blob/9697dc29c03ba7e5513be5a0c4f9902626917616/flair/trainers/trainer.py#L542-L543

https://github.com/flairNLP/flair/blob/9697dc29c03ba7e5513be5a0c4f9902626917616/flair/training_utils.py#L303-L304

**To Reproduce**

```
from flair.datasets import UD_ENGLISH
from flair.embeddings import WordEmbeddings
from flair.models import SequenceTagger
from flair.trainers import ModelTrainer

# 1. get the corpus
corpus = UD_ENGLISH().downsample(0.1)

# 2. what label do we want to predict?
label_type = 'upos'

# 3. make the label dictionary from the corpus
label_dict = corpus.make_label_dictionary(label_type=label_type)

# 4. initialize embeddings
embeddings = WordEmbeddings('glove')

# 5. initialize sequence tagger
tagger = SequenceTagger(hidden_size=256,
                        embeddings=embeddings,
                        tag_dictionary=label_dict,
                        tag_type=label_type,
                        use_crf=True
)

# 6. initialize trainer
trainer = ModelTrainer(tagger, corpus)

# 7. start training
trainer.train('resources/taggers/example-upos',
              learning_rate=0.1,
              mini_batch_size=32,
              max_epochs=10,
              save_model_each_k_epochs=1,
)
```

**Expected behavior**

All messages should be logged using `logger.getLogger('flair')`.

**Screenshots**

Check that the message where learning rate is reduced is not being logged.

```
2022-03-03 15:49:45,287 ----------------------------------------------------------------------------------------------------
2022-03-03 15:49:49,655 epoch 19 - iter 14/141 - loss 0.02138027 - samples/sec: 57.27 - lr: 0.100000
2022-03-03 15:49:53,916 epoch 19 - iter 28/141 - loss 0.02371238 - samples/sec: 53.10 - lr: 0.100000
2022-03-03 15:49:57,994 epoch 19 - iter 42/141 - loss 0.02502675 - samples/sec: 55.68 - lr: 0.100000
2022-03-03 15:50:02,287 epoch 19 - iter 56/141 - loss 0.02439726 - samples/sec: 52.46 - lr: 0.100000
[...]

PR #2665, opened 2022-03-10
GH-2654: Fixed printing and logging inconsistencies.

Xabilahu:
This PR closes https://github.com/flairNLP/flair/issues/2654.


flair/datasets/biomedical.py
```diff
@@ -5089,7 +5089,7 @@ def entity_type_predicate(member):
 
                 self.huner_corpora.append(corpus)
             except Exception:
```
Weyaaron, line 5091: Catching bare Exceptions is considered bad practice, however this code is not changed by this PR.

flair/datasets/entity_linking.py
```diff
@@ -1312,7 +1312,7 @@ def __init__(
                 sample_missing_splits=sample_missing_splits_in_each_corpus,
             )
             corpora.append(corpus)
-        print("...done!")
+        log.info("...done!")
```
Weyaaron, line 1316: This message could be improved upon, since the content of "what has been done" is context-depended.
My suggestion would be to write 
```
log.info("Done with transforming data into column format and creating corpora...")
```

flair/datasets/sequence_labeling.py
```diff
@@ -1154,7 +1154,7 @@ def __init__(
                 with open(data_path / data_file, "w") as file:
                     file.writelines(lines)
 
-                print(data_path / data_file)
+                # print(data_path / data_file)
 
         super(NER_DANISH_DANE, self).__init__(
```
Weyaaron, line 1159: This should be deleted.

flair/embeddings/token.py
```diff
@@ -944,7 +944,7 @@ def __init__(self, embeddings: str, use_local: bool = True, field: str = None):
         self.precomputed_word_embeddings: gensim.models.FastText = gensim.models.FastText.load_fasttext_format(
             str(embeddings)
         )
-        print(self.precomputed_word_embeddings)
+        # print(self.precomputed_word_embeddings)
```
Weyaaron, line 948: Should be deleted as well.

flair/file_utils.py
```diff
@@ -167,7 +167,7 @@ def download_file(url: str, cache_dir: Union[str, Path]):
     filename = re.sub(r".+/", "", url)
     # get cache path to put the file
     cache_path = cache_dir / filename
-    print(cache_path)
+    # print(cache_path)
```
Weyaaron, line 171: This should be deleted.

Weyaaron:
@Xabilahu  Thanks for the work so far, I added some comments. Would you like to include them or shall I take over from here?

Xabilahu:
I can add them in a moment, dw ;)

flair/datasets/biomedical.py
Xabilahu: Refactored to catch each exception type separately.

Weyaaron: I appreciate the effort, thanks a lot!

Weyaaron:
This PR has my approval, well done @Xabilahu

Weyaaron (approved):

Weyaaron:
@Xabilahu Apparently some imports added during the fix break the CI. Could you adress these issues?

Xabilahu:
It seems that `gzip.BadGzipFile` exception was introduced in Python 3.8. In Python 3.6 the library just raises `OSError`, which is already caught, so it is fixed just by deleting the conficting `except` block. My bad 😞

Weyaaron:
> It seems that `gzip.BadGzipFile` exception was introduced in Python 3.8. In Python 3.6 the library just raises `OSError`, which is already caught, so it is fixed just by deleting the conficting `except` block. My bad disappointed

Good to know, no problem, thanks for doing this work :)

Xabilahu:
> > It seems that `gzip.BadGzipFile` exception was introduced in Python 3.8. In Python 3.6 the library just raises `OSError`, which is already caught, so it is fixed just by deleting the conficting `except` block. My bad disappointed
> 
> Good to know, no problem, thanks for doing this work :)

Thanks for the amazing toolkit that you guys made available!

alanakbik:
@Xabilahu thanks for fixing this!

PR #2671, opened 2022-03-13
Update dependencies and use poetry for developers

Lingepumpe:
- removed unused dependencies: more-itertools, tabulate
- removed unused direct dependencies still installed transitivity:
  huggingface-hub, python-dateutil, regex, sentencepiece
- Updated past "requirements.txt version spec" after checking
  compatibility: gdown, flake8, konoha
- Restricted versions of most dependencies more tightly from below
  (require new versions), but avoid capping versions from above unless
  really necessary to allow better co-existence with other libraries
- Conditionally handle python version for numpy, scikit-learn, scipy,
  matplotlib and flake8-black as they each dropped python-3.6 support
  and some even python-3.7 support with newer versions.
- Avoid scipy 1.8.0 (for python>=3.8), because gensim generates
  deprecated warnings with it (when running pytest)

- Tested `poetry install` works and pytest passes for python-3.6.15, python-3.7.12, python-3.8.12, python-3.9.10
- python-3.10.2 works for everything except installing a python-3.10 compatible torch, manually installing a nightly torch all tests pass. Once torch-1.12 is released this will work without further changes.
- Remove requirements.txt and requirements-dev.txt and Pipfile, all requirements are now in pyproject.toml. To publish to pypi `poetry build` and `poetry publish` should work - only build tested by me. See https://python-poetry.org/docs/libraries/ for details.

Side fixes:
- Make sure mypy properly ignores data_fetcher.py and legacy.py (pytest was not passing on newer versions of mypy)
- Fix .pre-commit-config.yaml - the "rev" field should be set to a specific revision (sha256 or tag), not "stable". Updating rev fields works via `pre-commit autoupdate`


Lingepumpe:
Not sure if I should open a ticket for this also, or if we just discuss in here. Let me know.

PR #2674, opened 2022-03-14
Fix the logic for computing validation score

hannah0n:
The code for getting the best validation score didn't seem to match what's written in the comment. I changed the if-statement logic in a couple of spots, mostly by switching on/off `not` so that the min/max mode corresponds to loss/score as well as train/dev loss (or score) to training with/without dev data.

### DIFF
--- a/flair/trainers/trainer.py
+++ b/flair/trainers/trainer.py
@@ -291,8 +291,8 @@ class ModelTrainer:
             optimizer.load_state_dict(optimizer_state_dict)
 
         # minimize training loss if training with dev data, else maximize dev score
-        anneal_mode = "min" if train_with_dev or anneal_against_dev_loss else "max"
-        best_validation_score = 100000000000 if train_with_dev or anneal_against_dev_loss else 0.0
+        anneal_mode = "min" if not train_with_dev or anneal_against_dev_loss else "max"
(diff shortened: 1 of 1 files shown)


alanakbik:
@hannah0n thanks for the PR! 

However, the original version should be correct. `train_with_dev=True` means that we include the dev data in the training data. In this case we can only anneal against training loss which should be minimized.

If `train_with_dev=False` we keep dev as a hold-out dataset for early stopping. In this case we either anneal against dev score (should be maximized, default) or against dev loss (should be minimized). Let me know if you have any questions!

hannah0n:
Thank you for the kind explanations! I was coming from the misunderstanding that `train_with_dev=False` is used for when no dev data exist, i.e. `dev_file=None` and `sample_missing_splits=False` for Corpus.

There is no mention (or did I miss it somewhere?) that dev dataset is required for the system, either by assigning a file to `dev_file` or setting `sample_missing_splits="only_dev"`. Having Corpus set up with `dev_file=None` and `sample_missing_splits=False` and ModelTrainer with `train_with_dev=False`, which is not prevented in the current code, would cause an error since there is no dev score. It would be great if this is handled more carefully for others who might as well be mistaken as me, by writing a comment about the required dev data setting or making the code anneal against train loss when dev is missing. Thanks!

PR #2698, opened 2022-04-01
Refined evaluate

marcelmmm:


alanakbik:
@marcelmmm thanks for fixing this!

PR #2711, opened 2022-04-07
Fix the SciSpacyTokenizer.tokenize() bug

### LINKED ISSUE #2710: SciSpacyTokenizer.tokenize() function is broken
**Describe the bug**
Two minor bugs were introduced to Flair's `SciSpacyTokenizer` as part of #2645
The bugs prevent usage of `SciSpacyTokenizer` and tokenizers that leverage it.

**To Reproduce**
Any use of classes that leverage `SciSpacyTokenizer` will raise errors. For example:

```python
from flair.tokenization import SciSpacySentenceSplitter

splitter = SciSpacySentenceSplitter()

sentences = splitter.split("abcd")
```

Causes
```
[/usr/local/lib/python3.7/dist-packages/flair/tokenization.py](https://localhost:8080/#) in tokenize(self, text)
    257         words: List[str] = []
    258         for word in sentence:
--> 259             word.append(word)
    260         return words
    261 

AttributeError: 'spacy.tokens.token.Token' object has no attribute 'append'
```

**To Fix**

***First issue***
The first problem is that there is a typo here: https://github.com/flairNLP/flair/blob/480d2c9afd66ab8d3bf40a676917e84dba3c4cee/flair/tokenization.py#L259
It should be `words.append`, not `word.append`.

***Second issue***
the `SciSpacyTokenizer.tokenize()` is supposed to return a list of `str` , but instad it returns a list of Spacy `Token` objects.

Happy to open a PR shortly

tadejmagajna:
Fixes #2710 

Makes sure the words are added to the correct list variable and that strings (not SpaCy Token objects) are returned.

### DIFF
--- a/flair/tokenization.py
+++ b/flair/tokenization.py
@@ -256,7 +256,7 @@ class SciSpacyTokenizer(Tokenizer):
         sentence = self.model(text)
         words: List[str] = []
         for word in sentence:
-            word.append(word)
+            words.append(word.text)
         return words
 
     @property


alanakbik:
@tadejmagajna thanks for fixing this!

Issue #2713, opened 2022-04-07
About TransformerWordEmbeddings class


matirojasg:
Hi, I am using the following line of code to create contextualized embeddings from a biomedical roberta model.

```
TransformerWordEmbeddings("PlanTL-GOB-ES/roberta-base-biomedical-clinical-es")
```
However, when executing the code, I get the following error:

 ```
 File "/home/x/work/clinical-nested-ner-mlc/venv/lib/python3.8/site-packages/flair/embeddings/base.py", line 652, in _extract_token_embeddings
    assert subword_start_idx < subword_end_idx <= sentence_hidden_state.size()[1]
AssertionError
```
This does not occur using other hugging face library models, what could be the error in the model? Here is the link to the repository:

https://huggingface.co/PlanTL-GOB-ES/roberta-base-biomedical-clinical-es

Thanks

matirojasg:
https://colab.research.google.com/drive/1Wl4eh26L00MJtSqOfgRbKiqkG3E5X1Oh?usp=sharing

helpmefindaname:
Hey @matirojasg thank you for reporting, @alanakbik as this might be due to my refactoring, I'll look into it soon

alanakbik:
Awesome, thanks @helpmefindaname!

matirojasg:
Thank you guys!

PR #2714, opened 2022-04-08
GH-2713: make transformer offset calculation more robust

helpmefindaname:
tests and fixes the issue with https://github.com/flairNLP/flair/issues/2713


alanakbik:

Issue #2720, opened 2022-04-10
Makes to_plain_string method robust to consecutive spaces


mauryaland:
**Is your feature/enhancement request related to a problem? Please describe.**
When dealing with a text with consecutive spaces, the to_plain_string method is unusable because the parameter whitespace_after is a boolean. We can still use the to_original_text method to keep the consecutive spaces, but when dealing with a long text split in several Sentence objects which keep the index of the original text, we got some performance issues.

**Describe the solution you'd like**
Change the whitespace_after parameter of the Token object from a boolean to an integer referring to the number of whitespaces after the token.
The method to_plain_string will then become something like:

```python
def to_plain_string(self):
        plain = ""
        for token in self.tokens:
            plain += token.text
            if token.whitespace_after > 0:
                plain += token.whitespace_after * " "
        return plain.rstrip()
```

If you are ok with this, I can work on a quick PR. Thanks.

alanakbik:
@mauryaland sure, would appreciate a PR! But what do you mean by performance issues for the to_original_text method?

mauryaland:
It is very specific to my use case. We deal with some very long text that we split into Sentence objects where we keep the original index for postprocessing purposes. Because of that, when using to_original_text_method, we've got some very long text sentences with a lot of whitespaces at the beginning that we need to strip. This takes way much time than using to_plain_string method.

I will work on a PR, thanks for the quick return!

PR #2721, opened 2022-04-10
GH-2720: handle consecutive whitespaces

mauryaland:
Related to the issue #2720.

One quick question, I do not fully understand the purpose of lines 789-791 in data.py:
```python
if token.start_position == 0 and len(self) > 0:
    token.start_pos = len(self.to_original_text()) + self[-1].whitespace_after
    token.end_pos = token.start_pos + len(token.text)
```
Why are you taking the length of the sentence (since self refers to the sentence here) to get start_pos parameter?

### DIFF
--- a/flair/data.py
+++ b/flair/data.py
@@ -468,7 +468,7 @@ class Token(_PartOfSentence):
         self,
         text: str,
         head_id: int = None,
-        whitespace_after: bool = True,
+        whitespace_after: int = 1,
         start_position: int = 0,
         sentence=None,
     ):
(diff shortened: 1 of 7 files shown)


alanakbik:
@mauryaland thanks a lot for adding this and sorry for reviewing so late! 

Regarding your question: I guess it is a bit inefficient to take sentence length. The idea is that tokens get added one after another to the sentence so each time when they get added, the sentence provides its current length as the start position of the new token. But with your change one could probably also use the position of the last token plus its whitespace_after information to get the start position of the new token.

PR #2724, opened 2022-04-11
add exclude labels parameter to trainer.train and minor change in AIDA

marcelmmm:
Add the exclude_labels parameter to ModelTrainer.train. The parameter is already contained in the evaluate of nn.model.classifier. Now every time evaluate is called in ModelTrainer.train the exclude_labels parameter is handed over.

Moreover  there is a minor change in the NEL_ENGLISH_AIDA corpus. Default, the (possibly outdated) wikipedia names will be used as label names. Before the corpus used the wikipedia ids to get the label names. 
One can now work with up-to-date label names and only use existing wikipedia pages by setting the parameter use_ids_and_check_existence to True.


alanakbik:
@marcelmmm what is this PR about? Can you please add a description to this PR?

alanakbik:

PR #2730, opened 2022-04-15
Remove unnecessary more-itertools pin

hauntsaninja:
The dependency and the pin were added in https://github.com/flairNLP/flair/pull/2312/files. more-itertools is a pretty stable library and the one function that is used has been present for the last 6 major versions.


alanakbik:
@hauntsaninja thanks, makes sense! And sorry for taking so long to review!

PR #2731, opened 2022-04-16
fix wrong initialisations of label (where data_type was missing) and …

helpmefindaname:
I noticed that predicting a sentence via a sequence tagger and passing 'return_probabilities_for_all_classes=True' lead to some issues like it works only on cpu and produces corrupted labels that cannot be displayed properly.
To mitigate the issuse with the creation of currpted labels, I updated the type hints and also looked at the distance prediction model as it also had a function.
For the distance prediction model I noticed that the respective functions are not used (and it also doesn't have a predict function, which I find strange) hence I deleted them.

### DIFF
--- a/flair/data.py
+++ b/flair/data.py
@@ -198,7 +198,7 @@ class Label:
     score needs to be between 0.0 and 1.0. Default value for the score is 1.0.
     """
 
-    def __init__(self, data_point, value: Optional[str], score: float = 1.0):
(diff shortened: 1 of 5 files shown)


alanakbik:
@helpmefindaname thanks a lot for fixing this! And also for removing the unnecessary functions from the distance prediction model ;)

Issue #2528, opened 2021-11-23
For productionzing Flair pytorch model using torch.jit.trace 


DevBey:
Feature Request

**Is your feature/enhancement request related to a problem? Please describe.**
Hi I am trying to deploy flair/ner-english-ontonotes-large into production for NER task. For all other pytorch or huggingface transformers models we were able to use torch.jit.trace to trace them. I am not sure how to do the same with this model.

**Describe the solution you'd like**
It would be great if there is a way to trace the model be it using torch.jit.trace https://pytorch.org/docs/stable/generated/torch.jit.trace.html especifally for XLM-Roberta.

sadakmed:
> I am not sure how to do the same with this model.

from the current state of how SequenceTagger and Sentence is implemented, It needs alot of monkey patching, or use inheritance for rewriting, generally u will be forced to use `torch.jit.script` to capture the code flow, remove rewrite imports, lambda, try except, typing (eg Union)....

helpmefindaname:
Hello,
I made a PR which should make it possible to use torch.jit.trace, can you take a look at https://github.com/flairNLP/flair/pull/2739/files#diff-12961c27a5afbdccd88c22a4678b771edbe9a64e512fd5db7b3ac4df4fa4234b and see if that works?

DevBey:
Hi @helpmefindaname,

can you please provide your slack, I have a few questions !!

DevBey:
Hi @helpmefindaname,

here is the same error I got while runnning it manually and using your test_transformer_jit_embeddings with torch.jit.neuron
```
INFO:Neuron:There are 6 ops of 3 different types in the TorchScript that are not compiled by neuron-cc: prim::CallFunction, aten::index, aten::embedding, (For more information see https://github.com/aws/aws-neuron-sdk/blob/master/release-notes/neuron-cc-ops/neuron-cc-ops-pytorch.md)
INFO:Neuron:Number of arithmetic operators (pre-compilation) before = 269, fused = 239, percent fused = 88.85%
WARNING:tensorflow:From /home/ubuntu/anaconda3/envs/tensorflow2_latest_p37/lib/python3.7/site-packages/torch_neuron/ops/aten.py:1530: where (from tensorflow.python.ops.array_ops) is deprecated and will be removed in a future version.
Instructions for updating:
Use tf.where in 2.0, which has the same broadcast rule as np.where
Output exceeds the size limit. Open the full output data in a text editor
[...]

PR #2739, opened 2022-04-23
GH-2640/transformer forward

helpmefindaname:
Another PR as part of https://github.com/flairNLP/flair/issues/2640
resructures the TransformerEmbeddings to be exportable as Onnx model.
This PR also reworks the subtoken2token mapping, as the huggingface tokenizers provide "word_ids" which can be used for this (see their new Book for details)

I consider the following should be part of the PR:
- [x] capabilities of exporting functional Onnx models.
- [x] consider refactoring context for flert 
- [x] extensive testing of all pooling variations and possibilities
- [x] performance testing of the new TransformerEmbeddings compared to the old once
- [x] implementation of OnnxTransformerEmbeddings as drop in replacement for TransformerEmbeddings  
- [x] examples for the standard optimizations in the OnnxRuntime (fusion & dynamic quantisation)
- [x] performance testing of the Onnx models
- [x] refactoring (don't have everything in embeddings.base.py)


helpmefindaname:
results from the performance testing:
grafik
grafik

I tested using the FLERT model on the first 200 (gpu) or 64 (cpu) sentences of the conll03 test-set.
with a batch-size of 1 (gpu) or 32 (cpu).

the speed is measured as "seconds per sentences" (lower is better) and the results show that there are slight but consistent improvements in all modes when comparing the master with this refactored version. (however cpu only is less consistent than others).
as quantisation was always possible, it was easy to compare new vs old.
for jit only the refactoring provides possibilities and that also only with some hacking:

first the embedding needs to be covered by a Jitable Wrapper:
```
class JitWrapper(TokenEmbeddings):
    def __init__(self, emb: TransformerWordEmbeddings):
        super().__init__()
        self.emb = emb

    @property
    def embedding_length(self) -> int:
        return self.emb.embedding_length

    def forward(
        self,
        input_ids: torch.Tensor,
        attention_mask: torch.Tensor,
        overflow_to_sample_mapping: torch.Tensor,
        word_ids: torch.Tensor,
    ):
        return self.emb.forward(
            input_ids=input_ids,
            attention_mask=attention_mask,
            overflow_to_sample_mapping=overflow_to_sample_mapping,
            word_ids=word_ids,
        )["token_embeddings"]

    def get_names(self) -> List[str]:
        return self.emb.get_names()

    def _add_embeddings_internal(self, sentences: List[Sentence]):
        tensors = self.emb._prepare_tensors(sentences)

        with torch.no_grad():
            embeddings = self.forward(**tensors)
            self.emb._extract_token_embeddings(embeddings, sentences)
```
the important part is, that the `forward` method of the JitWrapper, has a similar signature to the `forward` of the `TransformerEmbeddings`, but removes all parameters that are set to None (for example it discards `langs`)


and in the model the jit functionallity needs to be called via `jit.trace`:
```
tensors = model.embeddings._prepare_tensors(sentences[:2])
model.embeddings = JitWrapper(model.embeddings)
params = []
for param in inspect.signature(model.embeddings.forward).parameters.keys():
    params.append(tensors[param])

model.embeddings.forward = torch.jit.trace(model.embeddings, params, strict=False).forward
```

alanakbik:
@helpmefindaname this is great, thanks for sharing these findings. So "main" in the measurements is the current master/main branch, correct? For CUDA, does this mean that you see improvements from 0.054 seconds per sentence to 0.044 seconds per sentence with the refactoring? This would be almost 20% faster, which would be awesome!

helpmefindaname:
> @helpmefindaname this is great, thanks for sharing these findings. So "main" in the measurements is the current master/main branch, correct? For CUDA, does this mean that you see improvements from 0.054 seconds per sentence to 0.044 seconds per sentence with the refactoring? This would be almost 20% faster, which would be awesome!

I think in this case it is more of a absolute improvement than a relative one. I suppose those 0.01s are mainly due to the rework of the subtoken matching. There was a lot of string matching which now is replaced by the word_ids created by the transformer tokenizers

helpmefindaname:
@alanakbik I noticed that I didn't answer your question, yes you are correct about the "main". Btw this branch is now ready for review, would be awesome if we manage to get it merged soon, feel free to approach me on slack if there are any questions

helpmefindaname:
as reported in https://github.com/flairNLP/flair/issues/2528 there is a case where using jit tracing via aws neutron "is it's performing almost 4x faster than a t4 GPU with 70% lower cost" 😄

alanakbik:
@helpmefindaname my mistake, getting the error now. Checking.

helpmefindaname:
@alanakbik as I got a notification that you did rerun the tests: I assume there is some incompatibility with the new torch `1.12.0` as it did work with `1.11.0` before. I will further investigate this

helpmefindaname:
That is easy to explain:

torch.jit doesn't translate the python code to a c++ module, but traces the code by running it once and store which tensor operations were run to generate a computation graph. In that graph we will have plenty of constants, for example the line:
`hidden_states.permute((0, 3, 1, 2)), 0, 1)` makes use of 6 constants. As jit is tracing and not interpreting the python code, it doesn't know that those are actually constants. It could be that we would have code like the following:
```
def fn(hidden_states, a, b, c, d, e, f):
    return hidden_states.permute((a, b, c, d)), e, f)
```
Where a,b,c,d,e,f would be integer parameters. In that case jit would see exactly the same, even if it was intended, that those parameters can change over calls. To prevent us from unintentionally converting such parameters to constants in the graph, we get this warning.

I suppose we get this warnings due to parameters like `stride`, `allow_long_sequences`, `hidden_dimension`, ...

helpmefindaname:
Okay, after long debugging I realized, that the system error was basically only the warning being cached as error by the pytest configuration. Hence the fix was very simple.
However a learning that I got:
the warning is only shown in pytorch 1.12.0 and it is due to the line:
` scores = scores.masked_fill(mask, torch.tensor(-float("inf")))` in the transformers package. Here a tensor is explicitly created with a constant value, which leads to the warning.

resources/docs/TUTORIAL_13_TRANSFORMERS_PRODUCTION.md
```diff
@@ -0,0 +1,145 @@
+# Speed up prediction speed of Transformer Embeddings
```
alanakbik, line 1: Awesome, thanks a lot for writing a tutorial for the feature!

alanakbik (approved):
Thanks a lot for adding this feature - this will surely be useful to many people!

heukirne:
Hello @helpmefindaname. 
First of all, thank you so much for this PR. It will be useful for several people!

I'm trying to follow your [tutorial](https://github.com/flairNLP/flair/blob/master/resources/docs/TUTORIAL_13_TRANSFORMERS_PRODUCTION.md) but I facing this issue:
`_forward_unimplemented() got an unexpected keyword argument 'input_ids'`
at
`script_module = torch.jit.trace(wrapper, parameter_list)`

I have a model trained with TransformerWordEmbeddings Flair 0.11
Should I train it again with last merged version Flair 0.11.3 (master)?

Cheers!

heukirne:
@helpmefindaname , training and tracing with Flair 0.11.3 (master) it works!

PR #2741, opened 2022-04-24
Bf/test dependencies

helpmefindaname:
update pytest and flake8 versions (all plugins are pytest 7.0 compatible by now)

try out https://github.com/utgwkk/pytest-github-actions-annotate-failures which should annotate pytest errors and therefore provide easier feedback for possible errors

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -7,3 +7,4 @@ types-Deprecated
 types-dataclasses
 types-tabulate
 types-requests
+pytest-github-actions-annotate-failures
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,9 +1,8 @@
-pytest==6.2.5
+pytest>=7.0.0
 pytest-mypy
 pytest-isort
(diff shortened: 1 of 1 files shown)


requirements-dev.txt
```diff
@@ -1,10 +1,10 @@
-pytest==6.2.5
+pytest>=7.0.0
 pytest-mypy
 pytest-isort
-pytest-flake8
+pytest-flake8-v2
```
whoisjones, line 4: why do we want to switch to flake8-v2?

helpmefindaname: I usually use flake8-v2 as it lets you specify the config in the `pyproject.toml` file within the `[flake8]` section.

However since I didn't change that config, using v2 has no benefit. As v2 currently has issues with the [stacktrace](https://github.com/VRGhost/pytest-flake8/pull/12), I reverted back to the original version.

alanakbik:
Thanks @helpmefindaname for adding this - looking forward to trying this plugin out.

Issue #2742, opened 2022-04-25
StackedEmbeddings with PooledFlairEmbeddings returning TypeError


Nuveyla:
**Describe the bug**
While running the given code for the best configuration for CONLL-03 NER on English at the following page: [link](https://github.com/flairNLP/flair/blob/b3797096742699997e77d12a66c82310361990a4/resources/docs/EXPERIMENTS.md) I get a TypeError (see screenshot). I have tested out several embeddings (Word and Flair) and saw that this TypeError is only occurring when running with PooledFlair. I have controlled the source code where the error is (token.py), but could not come up with a possible solution.

**Screenshots**
![image](https://user-images.githubusercontent.com/37611243/165096669-942b2c80-7454-4198-a2a4-e22f76535fc1.png)
![image](https://user-images.githubusercontent.com/37611243/165096805-345579fb-c965-4d44-8da9-fde66773db4a.png)

**Environment:**
 - Windows
 - python 3.10
 - flair -0.11

alanakbik:
@Nuveyla thanks for reporting this. @susannaruecker can you take a look?

alanakbik:
Fixed in PR #2744

PR #2744, opened 2022-04-26
#2742 added missing property decorator in PooledFlairEmbeddings

susannaruecker:
As reported in #2742, there was a problem with getting the `embedding_length` from the `PooledFlairEmbeddings`.
The `@proberty` decorator was missing above the `embedding_length` method of the PooledFlairEmbedding class.
I added it in token.py (line 868), so now it's:
```
@property
def embedding_length(self) -> int:
    return self.__embedding_length
```
This should resolve the problem in #2742.

### DIFF
--- a/flair/embeddings/token.py
+++ b/flair/embeddings/token.py
@@ -865,6 +865,7 @@ class PooledFlairEmbeddings(TokenEmbeddings):
 
         return sentences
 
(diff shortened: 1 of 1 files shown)


alanakbik:
@susannaruecker thanks for fixing this!

PR #2753, opened 2022-05-07
github actions cache

helpmefindaname:
cache models/datasets that are used in pytest. This should speed up tests on all parts that involve downloading flair or huggingface specific models/embeddings/datasets.

If at some point the cache leads to incorrect tests (e.g. a model is updated without getting a new name), the cache can be reset by updating `key: cache-v1` to `key: cache-v<n>` (n being an incremental number) such that the next pass downloads all models again. The old cache key will then be available for one more week and then be automatically deleted.

This PR also fixes a small bug, where cached paths have repeated folder structure if the cached path (e.g. setting the cache folder to `cache/flair` lead to the structure being `cache/flair/cache/flair/embeddings/...` instead of `cache/flair`

comparing the last two test runs (first one was uncached, second was cached) there is a difference of 5:37, even when removing the time the uncached needed for uploading the cache) wich leads to a reduction of ~34%

### DIFF
--- a/flair/__init__.py
+++ b/flair/__init__.py
@@ -7,6 +7,7 @@ from transformers import set_seed as hf_set_seed
 
 # global variable: cache_root
 cache_root = Path(os.getenv("FLAIR_CACHE_ROOT", Path(Path.home(), ".flair")))
+cache_root.mkdir(exist_ok=True, parents=True)
 
 # global variable: device
 if torch.cuda.is_available():
--- a/flair/__init__.py
+++ b/flair/__init__.py
@@ -7,7 +7,6 @@ from transformers import set_seed as hf_set_seed
 
 # global variable: cache_root
 cache_root = Path(os.getenv("FLAIR_CACHE_ROOT", Path(Path.home(), ".flair")))
-cache_root.mkdir(exist_ok=True, parents=True)
 
 # global variable: device
 if torch.cuda.is_available():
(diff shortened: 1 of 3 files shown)


alanakbik:
@helpmefindaname thanks a lot for adding this - makes the tests run much faster!

flair/__init__.py
```diff
@@ -7,6 +7,7 @@
 
 # global variable: cache_root
 cache_root = Path(os.getenv("FLAIR_CACHE_ROOT", Path(Path.home(), ".flair")))
+cache_root.mkdir(exist_ok=True, parents=True)
```
alanakbik, line 10: why is this necessary?

flair/file_utils.py
```diff
@@ -82,9 +82,12 @@ def cached_path(url_or_filename: str, cache_dir: Union[str, Path]) -> Path:
     return the path to the cached file. If it's already a local path,
     make sure the file exists and then return the path.
     """
-    if type(cache_dir) is str:
-        cache_dir = Path(cache_dir)
-    dataset_cache = flair.cache_root / cache_dir
+    cache_dir = Path(cache_dir)
+
+    if flair.cache_root not in cache_dir.parents:
```
alanakbik, line 87: what does this check do?

helpmefindaname: This checks if the cache_dir is already part of the cache_root and doesn't duplicate the path. E.g.: makes the cached files being placed in `cache/flair/...` instead of `cache/flair/cache/flair/...`.

flair/__init__.py
helpmefindaname: I ran into problems with the cache folder not being created correctly, but thinking of it again, I think that was due to different errors and this line is not necessary. I'll remove it again

alanakbik: thanks for double-checking!

PR #2775, opened 2022-05-17
Deprecate Python 3.6

### LINKED ISSUE #2707: Deprecate Python 3.6 ?
Python 3.6 is [out of life now](https://www.python.org/downloads/release/python-368/). Maybe consider upgrading to Python 3.7?

The current CI test only runs for Python 3.6. My suggestion is to test on multiple Python version [like here](https://github.com/gradsflow/chitra/blob/master/.github/workflows/main.yml#L16).

**Additional context**
If you decide to upgrade, I would be happy to make a contribution for this 😄 

Thanks

aniketmaurya:
fixes #2707

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -19,7 +19,7 @@ ftfy
 sentencepiece==0.1.95
 konoha<5.0.0,>=4.0.0
 janome
-gdown==3.12.2
+gdown==4.4.0
 huggingface-hub
 conllu>=4.0
 more-itertools
--- a/requirements.txt
+++ b/requirements.txt
@@ -25,4 +25,3 @@ conllu>=4.0
 more-itertools
 wikipedia-api
 pptree
-sacremoses
--- a/requirements.txt
+++ b/requirements.txt
@@ -25,3 +25,4 @@ conllu>=4.0
 more-itertools
 wikipedia-api
 pptree
+sacremoses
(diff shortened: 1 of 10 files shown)


requirements.txt
```diff
@@ -25,3 +25,4 @@ conllu>=4.0
 more-itertools
 wikipedia-api
 pptree
+sacremoses
```
aniketmaurya, line 28: This package was not installed when I changed from Python 3.6 to 3.7

alanakbik: Yes it is no longer a dependency of the `transformers` library which also deprecated support for 3.6. We removed the class that requires `sacremoses` so adding the dependency should not be required.

aniketmaurya: actually, without `sacremoses` I noticed an import error from `transformers` library.

alanakbik: this should be fixed in master branch, see #2768 - is your error on the latest Flair release or on current master branch state?

aniketmaurya: I got [this error](https://github.com/aniketmaurya/flair/runs/6464440062?check_suite_focus=true) in the Github CI, so it would be the current master branch state.

alanakbik: I think your branch is behind current master by a few commits. It was fixed by one of the last commits.

aniketmaurya: oh yes, sorry! Just merged with master.

aniketmaurya:
closing this since CI is already running with 3.7

PR #2778, opened 2022-05-18
Update konoha upper bound

hauntsaninja:
You can find the list of changes in konoha 5.0 here:
https://github.com/himkt/konoha/releases/tag/v5.0.0

None of them appear to affect flair.

(In general, I recommend against potentially unnecessary pins in
libraries. The reason I'm here is actually because konoha in turn
had an unnecessary pin of another library that we needed to update)

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -4,6 +4,7 @@ pytest-isort
 pytest-flake8
 flake8-black
 flake8<4.0.0
+importlib-metadata<4.3
 types-Deprecated
 types-dataclasses
 types-tabulate
--- a/requirements.txt
+++ b/requirements.txt
@@ -17,7 +17,7 @@ langdetect
 lxml
 ftfy
 sentencepiece==0.1.95
-konoha<5.0.0,>=4.0.0
+konoha<6.0.0,>=4.0.0
 janome
 gdown==4.4.0
 huggingface-hub


alanakbik:
Hello @hauntsaninja thanks for updating this! Unfortunately it seems the update is breaking our unit tests.

 I think this may have something to do with some of the dependencies added between 4.0.0 and 5.0.0 (see https://github.com/himkt/konoha/compare/v4.0.0...v5.0.0#diff-f53a023eedfa3fbf2925ec7dc76eecdc954ea94b7e47065393dbad519613dc89) and in particular the importlib-metadata dependency (see https://github.com/celery/kombu/issues/1339 and deprecation info [here](https://github.com/python/importlib_metadata/blob/0c86a02fbbbbdc9a888d9f2a149ec0d2cf5e7d2c/CHANGES.rst#v390))

hauntsaninja:
Yeah, looks like there's some controversy surrounding flake8 and importlib-metadata, e.g. https://github.com/python/importlib_metadata/pull/319 https://github.com/PyCQA/flake8/pull/1438

hauntsaninja:
Adding a pin in requirements-dev.txt to keep flake8 happy seems to have worked!

Lingepumpe:
I would not cap the version from above unless there is a known issue that this fixes. My vote is for `konoha>=4.0.0`, with no upper cap.

himkt:
Hello @hauntsaninja, this is the author of konoha. Thank you so much for the patch of konoha upgrade.
(Sorry for being inactive a long time)

@alanakbik Now the CI looks green with the latest version of konoha ([action](https://github.com/himkt/flair/actions/runs/3182741611/jobs/5189080807). Do we still have some blocker for this?

One possible concern is about Python3.6 (konoha won't work with py36 due to some dependency updates). But I found the installation already fails with the latest master branch w/py36 ([action](https://github.com/himkt/flair/actions/runs/3183007588/jobs/5189690646), simply runs CI with py36). If we can drop the support for py36 (or already dropped it), I think it's possible to upgrade the konoha version as well.

hauntsaninja:
This has conflicts now and I'm not interested in seeing this PR through. Was easier to delete flair from the codebase at work.

himkt:
@alanakbik If we can work on upgrading the package, please let me know. :pray: I'll create a tracking issue or PR.

Issue #2784, opened 2022-05-19
BIO tag format crashes with IndexError


cpalenmichel:
**Describe the bug**
Training an NER model with tag_format="BIO" fails with `The string 'S-ORG' is not in dictionary! Dictionary contains only: ['O', 'B-DATE', 'I-DATE', 'B-LOC', 'I-LOC', 'B-ORG', 'I-ORG', 'B-PER', 'I-PER']` and an IndexError.
It looks like the issue comes from `_get_gold_labels` only converting to BIOES. 

**To Reproduce**
I'm using conll03 English data and my tagger is setup as follows 
```
embeddings = TransformerWordEmbeddings(
                "xlm-roberta-base",
                layers="-1",
                layer_mean=False,
                fine_tune=True
            )
tagger: SequenceTagger = SequenceTagger(
        embeddings=embeddings,
        tag_dictionary=tag_dictionary,
        tag_type=tag_type,
        reproject_embeddings=True,
        use_crf=False,
        use_rnn=False,
        word_dropout=0.0,
        locked_dropout=0.0,
        tag_format="BIO"
    )

```

**Expected behavior**
I expect it to train a model using BIO tags without crashing.

**Screenshots**

**Environment (please complete the following information):**
 - OS [e.g. iOS, Linux]: Linux 5.4.0-100-generic x86_64
 - Version [e.g. flair-0.3.2]: flair==0.11.2

**Additional context**

bashayerm:
I have the same problem, looking for a solution

kaliaanup:
I am having the same issues when I enable tag_format="BIO". It basically gives the following error.

2022-05-25 16:20:11,118 The string 'E-ORG' is not in dictionary! Dictionary contains only: ['O', 'B-LOC', 'I-LOC', 'B-PER', 'I-PER', 'B-ORG', 'I-ORG', 'B-OTHER', 'I-OTHER', '<START>', '<STOP>']

For now I am bypassing the issue by disabling it. The predictions will be based on spans and BIOES format. Spans or BIOES format can be converted into BIO format.

https://github.com/flairNLP/flair/issues/2790

alanakbik:
@kaliaanup this will be fixed in PR https://github.com/flairNLP/flair/pull/2787

salohnana2018:
I have the same issue, but what is PR?

PR #2787, opened 2022-05-21
allow BIO evaulation on sequence tagger

helpmefindaname:
fixes https://github.com/flairNLP/flair/issues/2784

### DIFF
--- a/flair/models/sequence_tagger_model.py
+++ b/flair/models/sequence_tagger_model.py
@@ -86,16 +86,17 @@ class SequenceTagger(flair.nn.Classifier[Sentence]):
                 # without UNK, we cannot evaluate on data that contains labels not seen in test
                 # with UNK, the model learns less well if there are no UNK examples
                 self.label_dictionary = Dictionary(add_unk=allow_unk_predictions)
+                assert self.tag_format in ["BIOES", "BIO"]
                 for label in tag_dictionary.get_items():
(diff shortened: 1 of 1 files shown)


alanakbik:
Thanks for fixing this @helpmefindaname! 

@susannaruecker can you review this PR?

susannaruecker (approved):
Thank you for fixing this @helpmefindaname!

I could reproduce the bug described in #2784 which was indeed coming from the missing BIO formatting in `_get_gold_labels`. With your changes, now the gold label formatting as well as training and evaluating works fine, as tested on both BIO and BIOES format.

bashayerm:
The problem doesn't solve for me, and I still get the same error 

error

helpmefindaname:
hi @bashayerm 
can you please add the full error message and a minimal example that keeps failing?

alanakbik:
@bashayerm this is only fixed in the master branch. If you installed through pip you don't yet have this change. You need to test with current master branch to see if the error is fixed.

helpmefindaname:
Hey @alanakbik as @bashayerm isn't responding any more and the PR is already approved, can we merge this bug fix?

alanakbik:
@helpmefindaname thanks for the bump, yes let's merge! Sorry for the long delays in merging PRs at the moment - this will improve in a few weeks (lots of deadlines until then)!

PR #2684, opened 2022-03-21
Add support for v2.0 of HIPE-2022 data

stefan-it:
Hi,

this PR adds support for v2.0 of the HIPE-2022 dataset.

The following changes were introduced:

* Version number is now part of the (internal, chached) dataset path. Now, `~/.flair/datasets/ner_hipe_2022` gets the version number appended, e.g. `~/.flair/datasets/ner_hipe_2022/v2.0`.
* The branch name argument is added to the `NER_HIPE_2022` constructor. It is now possible to specify the desired branch name for the HIPE upstream data repo.
* New splits (train and dev) for German, English and French for the AJMC dataset are introduced with the v2.0.
* SONAR Dev split is updated.
* NewsEye German train split is filtered (removal of unannotated documents).
* Extensive tests for label dictionary (all datasets/languages/splits) added.

PR in the HIPE upstream data repo: https://github.com/hipe-eval/HIPE-2022-data/pull/3 and [release notes here](https://github.com/hipe-eval/HIPE-2022-data/releases/tag/v2.0).

PR #2675, opened 2022-03-14
Add support for HIPE 2022

stefan-it:
Hi,

this PR adds support for the recently released [HIPE 2022 Shared Tasks](https://hipe-eval.github.io/HIPE-2022/) NER datasets.

HIPE 2022 is a lot more challenging, because of more datasets, languages and different label sets compared to the previous CLEF-HIPE 2020 dataset.

The current released v1.0 version of the dataset includes support for the following datasets and languages:

| Dataset name | Languages              | Labels
| ------------ | ---------------------- | --------------------------------------------------------
| `ajmc`       | `de`, `en`             | `pers`, `work`, `loc`, `object`, `date`, `scope`
| `hipe2020`   | `de`, `en`, `fr`       | `pers`, `org`, `prod`, `time`, `loc`
| `letemps`    | `fr`                   | `pers`, `loc`
| `newseye`    | `de`, `fi`, `fr`, `sv` | `PER`, `LOC`, `ORG`, `HumanProd`
| `sonar`      | `de`                   | `PER`, `LOC`, `ORG`
| `topres19th` | `en`                   | `LOC`, `BUILDING`, `STREET`, `ALIEN`, `OTHER`, `FICTION`

More details can be found in the [dataset documentation](https://github.com/hipe-eval/HIPE-2022-data/blob/main/README.md) from HIPE 2022 repo.

In the current form, there's no "opinionated" corpus implementation. That means: no special normalization (e.g. de-hyphenation) is done: all datasets are just processed as accurate as possible.

Here's a quick example of how to use the Finnish part of the `newseye` dataset:

```python
corpus = NER_HIPE_2022(dataset_name="newseye", language="fi")
```

As the HIPE 2022 is part of an ongoing Shared Task, there's no test data available in the current v1.0 release. It will be added in future versions of HIPE 2022.

# Caveats

Some datasets come with no training split, such as `ajmc`, english part of `hipe2020` and `sonar`.

The `neweye` datasets are really special, because they come with two different development sets: `dev` and `dev2`. For this reason, the `dev_split_name` argument can be used to control which development split should be used:

```python
corpus = NER_HIPE_2022(dataset_name="newseye", language="de", dev_split_name="dev2")
```

# Example

The NER datasets in HIPE 2022 can be used with the `NER_HIPE_2022` implementation. Only two arguments are necessary to initialize a corpus:

* Dataset name (`dataset_name`)
* Language (`language`)

[...]

PR #2807, opened 2022-06-09
docs: mention HIPE-2022 in corpus tutorial

stefan-it:
Hi,

this PR mentions the recently added HIPE-2022 dataset (c.f. #2675 or #2684) in the corpus tutorial.


alanakbik:

PR #2811, opened 2022-06-15
more precise type hint for eval_on_train_fraction

thlor:
The train() method of trainer.py accepts an eval_on_train_fraction parameter, with either a float or "dev" literal string value (see docstring). This commit adds the "dev" literal option to the type hint.

### DIFF
--- a/flair/trainers/trainer.py
+++ b/flair/trainers/trainer.py
@@ -8,7 +8,7 @@ import time
 import warnings
 from inspect import signature
 from pathlib import Path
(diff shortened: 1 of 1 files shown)


thlor:
Looks like Literal typehint is only available from python 3.8 but the tests run on 3.7. I guess we can get back to this when 3.7 support is dropped, and maybe add a plain `str` typehint in the meantime? At least this way IDEs won't complain about incorrect typging when passing "dev" as the argument (though won't complain about any other, invalid strings, either).

alanakbik:
@thlor thanks for adding this!

PR #2812, opened 2022-06-15
Remove "tag_to_bioes" (Sequence) Corpus parameter, as it is not used

Lingepumpe:
After inspecting the current flair code, I see no functionality tied to `tag_to_bioes` parameter, so I removed it.


alanakbik:

stefan-it:
@alanakbik We should mention that in the release notes of the upcoming next release (breaking change, e.g. I've just that argument a lot in my training scripts) :)

PR #2815, opened 2022-06-16
Fix left-over cases of token.get_tag(), which was renamed

Lingepumpe:
Fixes some cases where .get_tag() was still being called on tokens.

Removed Sentence.__copy__ method, which was not working for multiple
reasons (not only get_tag). We could fix __copy__, is it used?


alanakbik:
@Lingepumpe thanks for fixing this! `copy` is currently not used (I think), so removing it is good for now!

PR #2817, opened 2022-06-16
JsonlDataset: Fix code that claims to set "O" labels to actually set them

Lingepumpe:
The current JsonlDataset does not set "O" labels on sentence tokens that are not explicitly labeled otherwise, but it claims to do so in the source code. This fixes the code to actually set "O" on each token.


alanakbik:

PR #2819, opened 2022-06-17
Add deprecation warning for SentenceDataset rename

Lingepumpe:
With 0.11, SentenceDataset was renamed to FlairDatapointDataset. There was
no deprecation or fall-back, this code adds the deprecation fallback with
a message to guide users to the new class name.

Alternatively, the deprecation could also be removed, allowing the
SentenceDataset class to continue without warning.


alanakbik:
@Lingepumpe thanks for fixing this! A deprecation fallback definitely makes sense, perhaps even as you write keeping the `SentenceDataset` object for convenience without deprecation note. I'll merge as-is and think about removing the deprecation note.

PR #2820, opened 2022-06-17
Pytest flake precommit update

Lingepumpe:
For review check the commits, not total changes, as the trailing-whitespace pre-commit hook created a lot of whitespace only changes.

This commit improves setup for contributors, while not touching any non dev dependencies.

- The current setup via pipenv (`pipenv install --dev`) leaves you with incompatible flake8 (3.9.2) and pytest (7.1.x). As fix, we allow pytest and flake8 to use their most recent versions, which are compatible. Also made sure dev dependencies from requirements-dev.txt match those of Pipfile. Pipfile will now also use python-3.7 to setup the environment.
- Updated CONTRIBUTING.md: We now mention pre-commit hook installation in the "Setup" section.
- Update pre-commit hooks `.pre-commit-config.yaml`. Fixes a syntax error, updates a outdated URL.
- Add end-of-file and trailing-whitespace fixer to pre-commit hooks.


Lingepumpe:
Looks like there are some download issues on the Hipe dataset that are failing the tests after rebasing to current master.

Lingepumpe:
I think I fixed the issue with HIPE2022, @stefan-it can you check my fix please? As far as I can tell, for `ner_hipe_2022/v2.1/ajmc/de/with_doc_seperator` the dev.txt file was present in the folder, which the code interpreted as go-ahead to skip running prepare_corpus, but this was in error, as test.txt was missing, which later on lead to an assertion error.

alanakbik:
@Lingepumpe thanks a lot for fixing the HIPE error. Could you separate this out into a separate PR? I would like to merge the Python 3.7-PRs later so we can do a last minor Python 3.6 release before updating to 3.7.

Lingepumpe:
> @Lingepumpe thanks a lot for fixing the HIPE error. Could you separate this out into a separate PR? I would like to merge the Python 3.7-PRs later so we can do a last minor Python 3.6 release before updating to 3.7.

Will do in the evening, I don't have the repo checked out on this machine. Note however that this PR here only bumps the python version for the `pipenv` setup (the Pipfile), it has nothing to do with library users or with our CI/Tests. Only developers that use pipenv would be affected (and looking at the state of pipenv setup, with pytest failing due to flake8 vs pytest issues, thats not a lot of people, or they have fixed it themselves quietly).

Lingepumpe:
Removed:
- The HIPE fix was split out and merged, and removed from the MR.
- Even though it is only for developers using pipenv, hold back on the python 3.7 bump, keeping it on 3.6 for now

Still in:
- Allow more recent pytest and flake8, as the reason they were pinned to an older version is no longer valid
- Update pipenv versions to match requirements-dev versions so developers get the same setup using pipenv vs requirements-dev [in current master pipenv setup is broken, pytest will fail]
- Improve pre-commit hooks, now does flake8 check, end-of-line whitespace and end-of-file newline checks, and uses more recent versions of the hooks

whoisjones (approved):

PR #2821, opened 2022-06-17
Update README.md

chriskamphuis:
Added accolades to the bibtex entries such that the proper capitalization is generated by LaTeX


alanakbik:
@chriskamphuis thanks for adding this!

PR #2831, opened 2022-06-22
Do not remove other labels of sentence for set_label on Token and Span

Lingepumpe:
Currently there is a bug when calling set_label on a Token or Span, this will also call set_label on the surrounding sentence. This will remove any other labels in this label_type, from the point of view of the Sentence.

```
>>> from flair.data import Sentence
>>> s = Sentence("one two three")
>>> s[0].set_label("ner", "B-Number")
>>> s
Sentence: "one two three" → ["one"/B-Number]
>>> s[1].set_label("ner", "I-Number")
>>> s
Sentence: "one two three" → ["two"/I-Number]
```

### DIFF
--- a/flair/data.py
+++ b/flair/data.py
@@ -436,20 +436,19 @@ class _PartOfSentence(DataPoint, ABC):
 
     def add_label(self, typename: str, value: str, score: float = 1.0):
         super().add_label(typename, value, score)
-
-        label = Label(self, value, score)
-        if typename not in self.sentence.annotation_layers:
-            self.sentence.annotation_layers[typename] = [label]
-        else:
-            self.sentence.annotation_layers[typename].append(label)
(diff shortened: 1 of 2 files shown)


flair/data.py
```diff
@@ -436,20 +436,19 @@ def _init_labels(self):
 
     def add_label(self, typename: str, value: str, score: float = 1.0):
         super().add_label(typename, value, score)
-
-        label = Label(self, value, score)
-        if typename not in self.sentence.annotation_layers:
-            self.sentence.annotation_layers[typename] = [label]
-        else:
-            self.sentence.annotation_layers[typename].append(label)
+        self.sentence.annotation_layers.setdefault(typename, []).append(Label(self, value, score))
```
alanakbik, line 439: interesting, much more compact!

alanakbik (approved):
Great catch, thanks a lot for fixing this!

Issue #2833, opened 2022-06-23
Request to update sentencepiece requirement to 0.1.96


tobiasb96:
Dear Flair developers,

we have noticed a minor version complication when installing Flair on our machines. We are using Python 3.10 and encountered the problem of `sentencepiece` not being installed properly due to a missing wheel on [PyPi](https://pypi.org/project/sentencepiece/#files).

They released the version 0.1.96 last year which should contain minor, non-breaking changes according to their release notes. They have only published the builts for 0.1.95 for Python <= 3.9. Therefore we have to use a workaround when installing flair with Python 3.10. 

Maybe it's possible to upgrade the version of `sentencepiece` in the `requirements.txt` or turn it into a "loose" requirement?

Thanks for your help!

bastianwegge:
@alanakbik sorry to bug you with a closed issue, but is there a chance to get a patch-release for this? Maybe something like `0.11.1`?

alanakbik:
Hello @bastianwegge we've made some bigger changes so we'll likely release a 0.12.0 version in a few weeks (that will drop Python 3.6 support). In the meantime, you could install directly from master branch to get this fix.

bastianwegge:
Thanks a lot for the update! We're looking forward to using 0.12.0 🙌

sbwiecko:
Thank you @alanakbik, `pip install git+https://github.com/flairNLP/flair` solved the issue on my installation 👍

Snowpile:
> Dear Flair developers,
> 
> we have noticed a minor version complication when installing Flair on our machines. We are using Python 3.10 and encountered the problem of `sentencepiece` not being installed properly due to a missing wheel on [PyPi](https://pypi.org/project/sentencepiece/#files).
> 
> They released the version 0.1.96 last year which should contain minor, non-breaking changes according to their release notes. They have only published the builts for 0.1.95 for Python <= 3.9. Therefore we have to use a workaround when installing flair with Python 3.10.
> 
> Maybe it's possible to upgrade the version of `sentencepiece` in the `requirements.txt` or turn it into a "loose" requirement?
> 
> Thanks for your help!

What was your work around for installing with Python 3.10?

PR #2835, opened 2022-06-24
use transformers sentencepiece requirement

helpmefindaname:
fixes https://github.com/flairNLP/flair/issues/2833

As flair doesn't use sentencepiece itself, but lets transformers use it indirectly, I propose to let the transformers package handle that dependency. 
This has two advantages:
a) we can use a more broad range of sentencepiece versions
b) if there are any new versions that break something, the transformers guys will handle it

Note: relies on https://github.com/flairNLP/flair/pull/2827 to be merged first for passing tests


alanakbik:
Makes a lot of sense, thanks @helpmefindaname!

PR #2837, opened 2022-06-24
update hf-hub version

helpmefindaname:
The newest release of the huggingface hub releases a new way of downloading models with additional features such as lazy loading and  git-aware cache file layouting.

I suppose there is no good reason to use older versions, so I added a pin of the version to force upgrades


helpmefindaname:
Notice: the tests fail due to the error addressed in https://github.com/flairNLP/flair/pull/2827 but if that PR will be merged, master will still fail due to the hf-hub upgrade. 
**Both** this PR and https://github.com/flairNLP/flair/pull/2827 need to be merged to make the master branch green again.

PR #2838, opened 2022-06-24
tars classifier always predict something on single label

helpmefindaname:
Currently, when TarsClassifier is used in the single-label setting, it can happen, that a prediction results to no label.
I think it makes more sense to have it like in the usual single-label setting, that there is always a class outputted. No matter if the highest logit is positive or not. Simply setting the threshold to 0 for non multi-label sequences should do the trick.


whoisjones:
👍 - just removed an unused variable as well

Issue #2762, opened 2022-05-12
All settings the same, but using "make_label_dictionary" and "make_tag_dictionary" to create "tag_dictionary" for NER model give completely different results.


lixiu911:
All other settings are the same (same data, same code etc), but using "make_label_dictionary" and "make_tag_dictionary" to create "tag_dictionary" for NER model give completely different results. 

I am training a Swedish NER model, I first trained the model using  "make_tag_dictionary" and got the results but then noticed that "make_tag_dictionary" will be deprecated and it is recommended to use "make_label_dictionary" so I changed to "make_label_dictionary". B-U-T they output completely different results. 
The code and output are as below. Is it a bug or is it something I missunderstand? Thanks!

With "make_tag_dictionary":
![8FC7240F-B2DF-4409-A8B9-D3909612287D](https://user-images.githubusercontent.com/16153200/168145401-45af2921-39f7-476b-a855-d9e0777c5579.png)

Results:
- F-score (micro) 0.4605
- F-score (macro) 0.4535
- Accuracy 0.3011

By class:
              precision    recall  f1-score   support

           B     0.5789    0.3894    0.4656       999
           I     0.6849    0.3257    0.4415       307

   micro avg     0.5978    0.3744    0.4605      1306
   macro avg     0.6319    0.3576    0.4535      1306
weighted avg     0.6038    0.3744    0.4599      1306
 samples avg     0.3011    0.3011    0.3011      1306

With "make_label_dictionary":
![5D0A5C6E-1370-410D-855C-626032F8BEB6](https://user-images.githubusercontent.com/16153200/168145670-0b6fba1a-4852-44ca-9bda-49d91dc4631c.png)

Results:
- F-score (micro) 0.0302
- F-score (macro) 0.2986
- Accuracy 0.0156

By class:
              precision    recall  f1-score   support

       <unk>     0.0000    0.0000    0.0000         0
           B     0.4382    0.4965    0.4655       999
           I     0.5000    0.3779    0.4304       307

   micro avg     0.0156    0.4686    0.0302      1306
   macro avg     0.3127    0.2914    0.2986      1306
weighted avg     0.4527    0.4686    0.4573      1306

alanakbik:
Hello @lixiu911 thanks for reporting this - this looks to be the same problem as in #2761 reported by @NicolasVautier. 

Could one of you provide some minimal sample data + script so we can reproduce this issue? On my datasets it seems to not appear.

lixiu911:
[...]

Issue #2761, opened 2022-05-12
UNK tag decrease f1-score


NicolasVautier:
Hello, this is my first time posting ever i will try to be as precise as possible.

i'm training a SequenceTagger model using flair 0.11.2 with 3 tags and i'm having issues with the unwanted add of unk tag to my tag dict which decreases micro f1-score (is it a normal behaviour ?). I've seen other issues about the unk taf but none speaks about f1-score.

Here is my code : 

            columns = {0 : "text", 1 : "ner"}
            corpus: Corpus = ColumnCorpus(path, columns,
                                        train_file = path + '/train.txt',
                                        test_file = path + '/test.txt')
            
            label_type = 'ner'

            tag_dictionary = corpus.make_label_dictionary(label_type=label_type)

            embeddings = StackedEmbeddings([WordEmbeddings('glove')])

            tagger = SequenceTagger(hidden_size=128,
                                                embeddings=embeddings,
                                                tag_dictionary=tag_dictionary,
                                                tag_type=label_type,
                                                use_crf=True)

            trainer = ModelTrainer(tagger, corpus)

            # Start training
            trainer.train(path,
                        learning_rate=0.1,
                        mini_batch_size=32,
                        max_epochs=10)

But when creating label_dictionary <unk> label is added and taken into account for training wich gives low f1-score micro avg despite good overall f1-score for other labels

![image](https://user-images.githubusercontent.com/48012106/168046335-192f5458-0092-4390-b1f4-7b49a522f3d1.png)

A quick fix would be to use the remove_item function of Dictionary class but it seems to dramatically decrease f1-score for other classes, maybe exclude unk from f1 score or add the possibility to edit add_unk directly when calling make_label_dictionary ?

Ps : my train.txt and test.txt files looks like this
`
word1 O
word2 O
word3 attributeA00356
word4 O
word5 attributeA00436
word6 O
word7 attributeA00354

djstrong:
It looks like micro f1 is not correctly calculated.

PR #2839, opened 2022-06-25
make `add_unk` optional and don't use it for ner

helpmefindaname:
The `<unk>` token is a common practice to have a mechanism to add labels that are not common enough for training. This mainly relates to classification and basically simulates an "others" category.
However, for NER, we have a different mechanism with the `O-tag`, hence I don't think `<unk>` has any practical use.
Using `<unk>` for NER can lead to down-weighting the overall score when the NER-model is not trained long enough (random apperances of `<unk>` still happen. See (https://github.com/flairNLP/flair/issues/2762 and https://github.com/flairNLP/flair/issues/2761) and can break pipelines, that do not expect entities marked as `unk` (https://github.com/flairNLP/flair/issues/2224).

I propose making the `<unk>` optional, per default `True`, but allow disableling it for all ner-tasks. As well as disableling it at all ner-related examples.

### DIFF
--- a/examples/ner/run_ner.py
+++ b/examples/ner/run_ner.py
@@ -113,7 +113,7 @@ def main():
     logger.info(corpus)
 
     tag_type: str = "ner"
-    tag_dictionary = corpus.make_label_dictionary(tag_type)
+    tag_dictionary = corpus.make_label_dictionary(tag_type, add_unk=False)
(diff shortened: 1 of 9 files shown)


whoisjones (approved):

alanakbik:

PR #2840, opened 2022-06-27
Save and load "tag_format" for sequence tagger model

aragaer:
Since "tag_format" is not saved or loaded, when loading the saved model default BIOES format is used which breaks when actual format is BIO since there are no "E-" or "S-" tags in the dictionary.


flair/models/sequence_tagger_model.py
```diff
@@ -631,6 +632,7 @@ def _init_model_with_state_dict(cls, state, **kwargs):
             state,
             embeddings=state["embeddings"],
             tag_dictionary=state["tag_dictionary"],
+            tag_format=state.get("tag_format", "BIOES"),
```
whoisjones, line 635: I would remove this here since "BIOES" is the default anyway when initializing a SequenceTagger.

whoisjones: suggestion, to stay consistent with other state keys: add tag_format as in lines 618-624:
`tag_format = "BIOES" if "tag_format" not in state.keys() else state["tag_format"]`

and pass tag_format in super() method

alanakbik:
@aragaer thanks for fixing this!

PR #2846, opened 2022-06-29
Refactorings for initialization from state dict

### LINKED ISSUE #2843: Refactoring - _init_model_with_state_dict()
Remove if conditions with get() method which is a better solution for using default parameters.

whoisjones:
Replacing use of if conditions with .get() method using default parameters.


whoisjones:
Related to gh-2840 (addition of tag-format default parameter to SequenceTagger)

alanakbik:
Thanks @whoisjones!

PR #2858, opened 2022-07-11
add py.typed file for PEP-561 compatibility

helpmefindaname:
[PEP-561](https://peps.python.org/pep-0561) is the standard for adding typehints to packages. 
As flair itself has already type hints for most methods, simply adding a [py.typed](https://peps.python.org/pep-0561/#packaging-type-information) file to the package data.

To explain further of why PEP-561 is useful:

let's create an environment installing `flair==0.11.3` and `mypy`.
And create a file `example_flair.py` with the following content:
```


data = Sentence("hello world")

data[0] = "abc"
```
Note that this is clearly a violation as `Sentence` doesn't implement `__setitem__` and `data[0]` would be rather a `Token` instead of a `str`

running `mypy example_flair.py` lead to the following error:
```
example_flair.py:1: error: Skipping analyzing "flair.data": module is installed, but missing library stubs or py.typed marker
example_flair.py:1: note: See https://mypy.readthedocs.io/en/stable/running_mypy.html#missing-imports
Found 1 error in 1 file (checked 1 source file)
```
A user of flair can handle this by adding `type: ignore` to the import, or add the following to the mypy config:
```
[mypy]
ignore_missing_imports = True
```
either way running `mypy example_flair.py` would pass by ignoring all potential errors:
`Success: no issues found in 1 source file`

However, if we install flair locally, we get the errors we would expect from our incorrect script:
```
example_flair.py:6: error: Unsupported target for indexed assignment ("Sentence")
Found 1 error in 1 file (checked 1 source file)
```

Simply having this `py.typed` file within the package data is enough to provide the users the full mypy warnings for any missusages.

(Note that we have set `include_package_data=True`, therefore the `py.typed` file will be automatically added to the package data without further describing it.


alanakbik:
Thanks for adding this and providing the explanation!

Issue #2859, opened 2022-07-12
module 'conllu' has no attribute TokenList


yash-rathore:
![image](https://user-images.githubusercontent.com/63536277/178562229-5b6595c0-218b-4db9-978b-fafbfd1508ed.png)

This did not happen till today . I have been using this basic code for 6 months and this bug is new and did not appear anytime before today. 
Please resolve quickly

helpmefindaname:
Hey @yash-rathore, looking at the conllu [release notes](https://pypi.org/project/conllu/#history) it looks like they released a new version that seems to change the structure.
I suppose you can fix your issue by pinning the version `conllu<=4.4.2` to ensure that you install a version that is still compatible.

hbailla:
I have the same issue. The strange thing is I have this error when I run the code on Google Colab , while it works fine on my local machine

hbailla:
> Hey @yash-rathore, looking at the conllu [release notes](https://pypi.org/project/conllu/#history) it looks like they released a new version that seems to change the structure. I suppose you can fix your issue by pinning the version `conllu<=4.4.2` to ensure that you install a version that is still compatible.

This works just fine. Thank you SO MUCH

yash-rathore:
> conllu<=4.4.2

This works absolutely fine . Thankyou for resolving this issue

pallashadow:
another option: add these 2 lines before import flair
`
import conllu
conllu.TokenList = conllu.models.TokenList
`

EmilStenstrom:
Thank you for reporting. I just released 4.5.1 which makes TokenList available as an import directly from conllu, and declares the public API via `__all__`.

helpmefindaname:
Thank you @EmilStenstrom,
yes this fixes the issue!

alanakbik:
Awesome @EmilStenstrom thanks for fixing this so quickly! And thanks @helpmefindaname for testing!

EmilStenstrom:
This issue can be closed, as it has been fixed in conllu 6 months ago.

Issue #2860, opened 2022-07-13
Dependency package conllu version issue


lilsgit:
We were using Flair in our corn job pipeline for the sentiment. Yesterday the pipeline just broke, I checked and the error shows, 

`conllu.TokenList:
AttributeError: module 'conllu' has no attribute 'TokenList'
`

It turns out conllu package version just got updated yesterday to 4.4.3 which has compatibility issue with Flair...

I rolled conllu version back to 4.4.2 for now and it's all working fine.

PR #2861, opened 2022-07-13
use conllu.models.TokenList as intended

helpmefindaname:
fixes https://github.com/flairNLP/flair/issues/2859 and it's duplicate: https://github.com/flairNLP/flair/issues/2860


helpmefindaname:
Closing as the conllu creator was kind enough to fix the issue on his side

PR #2888, opened 2022-08-05
Convenience method for learning rate factor

alanakbik:
This PR adds a parameter to set a factor on the learning rate of the decoder, if fine-tuning a model. 

Usage: 

```python
trainer.fine_tune(f"path/to/output/folder",
                  mini_batch_size=4,
                  learning_rate=5e-5,
                  decoder_lr_factor=10,
                  )
```

### DIFF
--- a/flair/trainers/trainer.py
+++ b/flair/trainers/trainer.py
@@ -893,20 +893,28 @@ class ModelTrainer:
         mini_batch_size: int = 4,
         embeddings_storage_mode: str = "none",
         use_final_model_for_eval: bool = True,
-        decoder_lr_factor: float = 1.,
+        decoder_lr_factor: float = 1.0,
         **trainer_args,
     ):
 
         # If set, add a factor to the learning rate of all parameters with 'decoder' in name
-        if decoder_lr_factor != 1.:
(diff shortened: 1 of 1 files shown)


helpmefindaname:
Hi @alanakbik,
if I understand this PR correctly, this is to set a different LR for pretrained weights and randomly initialized weights on all `DefaultClassifier`'s?

I suppose this could be extended even further to for example `SequenceTagger`, by filtering by `embedding` e.g:

```
embedding_parameters = [param for name, param in self.model.named_parameters() if "embedding" in name]
model_parameters = [param for name, param in self.model.named_parameters() if "embedding" not in name]
```

Such that we could train a Transformer-Bert model with a higher LT for the CRF part

alanakbik:
Yes, it is for training the non-pretrained (i.e. randomly initialized) parts with a higher LR. Since the decoder is always randomly initialized, it is handled here. Extending this to the LSTM-CRF of the SequenceTagger would be great, but some embeddings (like CharacterEmbeddings) are randomly initialized, while others are not. So I think it's not easy to come up with a good heuristic to identify those parts.

Issue #2882, opened 2022-08-02
Distilbert embeddings error in sentence with trailing whitespace


alanakbik:
`TransformerWordEmbeddings` with `distilbert-base-uncased` cannot embed an empty token if at the end of a sentence: 

```python
# sentence with two empty tokens, one at end
sentence = Sentence(["hello", " ", "hm", " "])

# embed with distilbert
embeddings = TransformerWordEmbeddings(model="distilbert-base-uncased")
embeddings.embed(sentence)

# print embeddings
for token in sentence:
    print(token)
    print(token.embedding[:10])
```

This prints: 

```console
Token[0]: "hello"
tensor([-0.3750,  0.2257,  0.5354, -0.5856,  0.1343, -0.2033,  0.1340,  0.0988,
        -0.2931, -0.5633])
Token[1]: " "
tensor([0., 0., 0., 0., 0., 0., 0., 0., 0., 0.])
Token[2]: "hm"
tensor([ 0.0508, -0.0291,  0.6872, -0.0561,  0.3941,  0.2963,  0.4232,  0.4120,
         0.0139, -0.2129])
Token[3]: " "
tensor([])
```

I.e. the last token gets no embedding, which causes problems in training routines. Empty whitespaces as tokens unfortunately occur in some corpora, such as WNUT.

PR #2891, opened 2022-08-06
GH-2882: fix trainling comma at the end of the sentence

helpmefindaname:
add the example as test and fixes the issue


alanakbik:

PR #2892, opened 2022-08-06
pin flake8 to v4

helpmefindaname:
As long as https://github.com/tholo/pytest-flake8/pull/88 is not merged, flake8 need to be set to version 4.
i suggest this PR as a temporary fix until the plugins are up to date again.


alanakbik:

PR #2895, opened 2022-08-07
warn if resuming with too low max_epochs & ' additional_epochs'  parameter

helpmefindaname:


alanakbik:
@helpmefindaname thanks for improving this and adding the test!

PR #2905, opened 2022-08-15
use a different way for mypy to ignore the legacy embeddings when typechecking

helpmefindaname:
I have absolutely no idea why this is necessary, but the old way of ignoring types withing the `legacy.py` file doesn't work anymore.


alanakbik:
@helpmefindaname thanks for addressing this!

Issue #2960, opened 2022-10-13
Missing docstring parameters for ModelTrainer.train()


thlor:
While most of the parameters of `ModelTrainer` class' `train()` method are documented, the following (keyworded) parameters have no docstring documentations:

- `eval_batch_size`
- `initial_extra_patience`
- `anneal_with_prestarts`
- `anneal_against_dev_loss`
- `batch_growth_annealing`
- `write_weights`
- `use_amp`
- `amp_opt_level`
- `use_swa`
- `use_final_model_for_eval`
- `gold_label_dictionary_for_eval`
- `exclude_labels`
- `optimizer_state_dict`
- `scheduler_state_dict`
- `save_optimizer_state`

alanakbik:
PR merged, thanks again!

thlor:
I updated the list of missing parameters - half of them are still to be done.

PR #2961, opened 2022-10-13
GH-2960 add docstrings missing for ModelTrainer.train() parameters

thlor:


alanakbik:
@thlor thanks for improving this!

PR #2980, opened 2022-11-07
Add layoutlm layoutxlm support

helpmefindaname:


flair/data.py
```diff
@@ -830,39 +843,33 @@ def clear_embeddings(self, embedding_names: List[str] = None):
...
 
             if respect_document_boundaries and sentence.is_document_boundary:
                 break
 
-            left_context = [t.text for t in sentence.tokens] + left_context
-            if len(left_context) > context_length:
-                left_context = left_context[-context_length:]
-                break
-        return left_context
+            left_context = sentence.tokens + left_context
```
whoisjones, line 857: `left_context += sentence.tokens` to be consistent with right context

flair/datasets/ocr.py
```diff
@@ -0,0 +1,250 @@
...
+        in_memory: bool = True,
+        encoding: str = "utf-8",
+        load_images: bool = False,
+        normalize_coords_to_thousands: bool = True,
+        label_name_map: Dict[str, str] = None,
+    ):
+        """
+        Instantiates a Dataset from a OCR-Json format.
+        The folder is structured with a "images" folder and a "tagged" folder.
+        Those folders contain respectively .jpg an .json files with matching file name.
```
whoisjones, line 29: *and

flair/datasets/ocr.py
```diff
@@ -0,0 +1,250 @@
...
+        The json contains 3 fields "words", "bbox", "labels" which are lists of equal length
+        "words" is a list of strings, containing the ocr texts,
+        "bbox" is a list of int-Tuples, containing left, top, right, bottom
+        "labels" is a BIO-tagging of the sentences
+        :param path_to_split_directory: base folder with the task data
+        :param label_type: the label_type to add the ocr labels to
+        :param encoding: the encoding to load the .json files with
+        :param normalize_coords_to_thousands: if True, the coordinates will be ranged from 0 to 1000
```
whoisjones, line 37: is normalizing to thousands usual? If it is just selected by random, why not make the normalization factor as int and optional and normalize images if factor is provided.

flair/embeddings/transformer.py
```diff
@@ -372,32 +419,52 @@ def prepare_tensors(self, sentences: List[Sentence], device: torch.device = None
...
-            return_tensors="pt",
-            is_split_into_words=True,
-        )
-
-        tensor_args = self.__build_transformer_model_inputs(
-            batch_encoding, sentences, offsets, lengths, flair_tokens, device
-        )
-
-        return tensor_args
+        # random check some tokens to save performance.
```
whoisjones, line 422: Why not check entire dataset or discard inputs not having all required metadata?

flair/embeddings/transformer.py
```diff
@@ -372,32 +419,52 @@ def prepare_tensors(self, sentences: List[Sentence], device: torch.device = None
...
         sentence_lengths: List[int],
-        flair_tokens: List[List[str]],
+        flair_tokens: List[List[Token]],
         device: torch.device,
     ):
+        tokenizer_kwargs: Dict[str, Any] = {}
+        if self.tokenizer_needs_ocr_boxes:
+            tokenizer_kwargs["boxes"] = [[t.get_metadata("bbox") for t in tokens] for tokens in flair_tokens]
+        else:
+
```
whoisjones, line 455: run formatter again should remove empty line

flair/embeddings/transformer.py
```diff
@@ -440,13 +509,17 @@ def __build_transformer_model_inputs(
                     lang_id = lang2id.get(sentence.get_language_code(), 0)
                     model_kwargs["langs"][sentence_idx : sentence_idx + part_length] = lang_id
                     sentence_idx += part_length
-        if self.token_embedding:
+
+        if "bbox" in batch_encoding:
+            model_kwargs["bbox"] = batch_encoding["bbox"].to(device, non_blocking=True)
+
+        if self.token_embedding or self.needs_manual_ocr:
```
whoisjones, line 516: we check 2x for `self.token_embedding` (line 547) and `self.needs_manual_ocr` (line 549) and jointly in line 516. The part required by both are the word_ids. Can we move everything after line 526 also in the condition for `self.token_embeddings`? looks like `self.needs_manual_ocr` only needs `word_ids_list`.

whoisjones (commented):
Looks good, I couldn't really check the I/O operation of images in the dataset. I have some minor things you might want to look at before merging.

flair/embeddings/transformer.py
helpmefindaname: As noted in the comment, running the checks on a broader scale would impact the speed, this is rather to ensure that the user gets a good warning if there are no bounding boxes in general, without impacting the performance.

And Discarding inputs would silently hide errors, so I am against that.

flair/datasets/ocr.py
helpmefindaname: Normalizing to thousands is very usual, it is done by Layoutlm (& v2/v3), Docformer, Lambert, ...
I haven't seen an implementation so far that did it different

flair/data.py
helpmefindaname: Notice that addition on lists is not cummutative, hence your suggestion would lead to wrong results.

whoisjones:
👍

alanakbik:

PR #2981, opened 2022-11-07
Collective tests

helpmefindaname:
This PR is about creating unified test classes, to ensure that all models & embeddings have tested the basic functionality.

During development, I made the following observations/fixes:

* RelationClassifier did output the label type as score
* TextRegression prediction fails on empty sentences instead of ignoring them
* TextRegression evaluation provided tensors to the Sklearn functions, leading to errors when the tensors are on GPU.
* The TextRegression evaluation metric goes from -1 to 1, hence a "best score" of `-0.5` wouldn't trigger to save a best model. 
* The TextRegression label_type was not a property like any other model was.
* DocumentCNNEmbeddings failed for sequences lower than the kernel length
* using `trainer.resume()` emitted a warning when a "best model" was created before
* using `trainer.resume(additional_epochs=...)` started from epoch 0 instead of the start, leading to way more epochs than expected.


alanakbik:
@helpmefindaname thanks a lot for creating these unified tests! The standardization and higher coverage will be very useful to improve overall quality.

Issue #2964, opened 2022-10-18
Can't reproduce export_onnx example from TUTORIAL_13


klimentij:
**Describe the bug**
When I try to export `ner-large` transformer embeddings to ONNX, I see an error `TypeError: forward() missing 1 required positional argument: 'lengths'`

**To Reproduce**

1. Install Flair==0.11.3 from git repo: `pip install git+https://github.com/flairNLP/flair.git` (`TransformerOnnxEmbeddings` class is not available when installed from pip)
2. Run code from the tutorial 13 with my own sentences:  
```
from flair.models import SequenceTagger
from flair.embeddings import TransformerWordEmbeddings, TransformerDocumentEmbeddings

model = SequenceTagger.load("ner-large")
assert isinstance(model.embeddings, (TransformerWordEmbeddings, TransformerDocumentEmbeddings))


sentences = [Sentence(s) for s in ['short sentence', 'this is very long sentence']]
model = SequenceTagger.load("ner-large")
model.embeddings = model.embeddings.export_onnx("flert-embeddings.onnx", sentences, providers=["CUDAExecutionProvider", "CPUExecutionProvider"])
```
3. I get the error `TypeError: forward() missing 1 required positional argument: 'lengths'`

**Expected behavior**
This creates a file `flert-embeddings.onnx` which stores the exported Onnx Model.

**Screenshots**
<img width="689" alt="image" src="https://user-images.githubusercontent.com/3247669/196463874-61339c71-08f9-4524-921e-92dc59f7a3e2.png">

**Environment (please complete the following information):**
 - Deep Learning AMI GPU PyTorch 1.12.1 (Amazon Linux 2) 
 - NVIDIA driver version: 510.73.08
 - CUDA version: 11.6 
 - Flair 0.11.3
 - onnx 1.12.0
 - onnxruntime 1.12.1
 - Pytorch 1.12.1.post201

valeriosofi:
Hi @klimentij, please try to uninstall it and then install it again with `pip install git+https://github.com/valeriosofi/flair.git@nebullvm`, you were missing the branch info in your command ;)

klimentij:
@valeriosofi re-tried with the branch info, the same error

valeriosofi:
[...]

Issue #2930, opened 2022-09-01
RuntimeError while saving embedding model in ONNX format


kobiche:
**Describe the bug**
The model cannot be saved in ONNX format as shown in the flair/resources/docs/TUTORIAL_13_TRANSFORMERS_PRODUCTION.md tutorial. The export raises the following error: 
`RuntimeError: Dynamic shape axis should be no more than the shape dimension for sequ_length`

**To Reproduce**
```python
from flair.data import Sentence
from flair.models import SequenceTagger
from flair.embeddings import TransformerWordEmbeddings, TransformerDocumentEmbeddings
import flair.datasets

if __name__ == '__main__':
    model = SequenceTagger.load("ner-large")
    assert isinstance(model.embeddings, (TransformerWordEmbeddings, TransformerDocumentEmbeddings))

    # Serialize the model (use sample sentences)
    sentences = list(flair.datasets.CONLL_03_SPANISH().test)[:5]
    model.embeddings = model.embeddings.export_onnx("flert-embeddings.onnx", sentences, providers=["CUDAExecutionProvider", "CPUExecutionProvider"])
```

**Expected behavior**
Model saved in ONNX format

**Environment (please complete the following information):**
 - OS: Linux
 - Flair version: master (last commit: [ee7e619](https://github.com/flairNLP/flair/commit/ee7e6191ff4b9bf55d9ea3caf480c6a3fd975cba))
 - ONNX runtime version: 1.12.0

**Additional context**
I found out that this problem arises when the `torch.onnx.export` is [called](https://github.com/flairNLP/flair/blob/master/flair/embeddings/transformer.py#L674). While debugging, I found out that this comes from the order of `example_tensors`. The model expects the following order `['input_ids', 'lengths', 'attention_mask', 'overflow_to_sample_mapping', 'word_ids']` so I suggest that the `example_tensors` is reordered accordingly (quick fix):
```python
desired_order = ['input_ids', 'lengths', 'attention_mask', 'overflow_to_sample_mapping', 'word_ids']
example_tensors = {k: example_tensors[k] for k in desired_order}
```
And while we are at it, I would suggest to fix the expected type of the `args` argument of `torch.onnx.export` (should be a tuple)
```python
        torch.onnx.export(
            embedding,
            tuple(example_tensors.values()),
            ...
        )
```

PR #2990, opened 2022-11-14
separate between token-lengths and sub-token lengths

helpmefindaname:
This PR should speed up the transformer word embeddings by a tiny bit (and also saves a little gpu usage), as the output embeddings are now of the token-length instead of being expanded to the sub-token-length.

I haven't tested how big the speed impact actually is, this would be a further todo.

Besides this, this PR fixes the Onnx Export ( https://github.com/flairNLP/flair/issues/2964  https://github.com/flairNLP/flair/issues/2930 ) by enforcing the order of inputs to be right


alanakbik:
Thanks @helpmefindaname - I'll merge this now and fix the test in the next PR.

PR #3002, opened 2022-11-27
Restructure Tokenizer and Splitter modules

alanakbik:
When initializing a `Sentence` object, with `use_tokenizer=True` (set by default), it triggered an import of the `SegtokTokenizer` in the Sentence constructor. According to [this thread](https://stackoverflow.com/questions/29504313/python-what-is-the-cost-of-re-importing-modules), there is a cost of re-importing modules and since we often create large numbers of `Sentence` objects, this is unnecessary overhead. 

This PR makes the import global. To do this, it splits up the `flair.tokenization` module into tokenization (for all tokenizers) and splitter (for all sentence splitters). This way, there is no longer an import of `flair.data` in `flair.tokenization`.

### DIFF
--- a/flair/data.py
+++ b/flair/data.py
@@ -16,6 +16,7 @@ from torch.utils.data.dataset import ConcatDataset, Subset
 
 import flair
 from flair.file_utils import Tqdm
+from flair.tokenization import SegtokTokenizer, SpaceTokenizer, Tokenizer
 
 T_co = typing.TypeVar("T_co", covariant=True)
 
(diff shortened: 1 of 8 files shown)


alanakbik:
However, it doesn't really seem to make a difference in runtime. Tested with 

```python
import timeit
t = timeit.Timer(setup='from flair.data import Sentence',
                 stmt='Sentence("a a a a a a a a a a ")')
print(t.timeit())
```

PR #3006, opened 2022-12-03
Update MultiCoNer

helpmefindaname:
This PR adds the following features:

* MultiCoNer can now download the dataset from the S3 opendata registry. This also includes the test set.
* When loading a columncorpus, comments can be used to add metadata e.g. `# id 2c8f5b49-4df5-44b5-8c2a-bdc340010ea3	domain=de-lowner` will add a metadata `domain` with the value `de-lowner`. In general, this is about tab separated `<key>=<value>` patterns.
* MultiCoNerV2 is added, but needs to be added manually similar to how it used to be with MultiCoNer.
* Loading datasets via a columncorpus won't warn about creating empty sentences for every single example. 
* When loading a columncorpus, each last token of a sentence now correctly has a whitespace_after=0, instead of possible 1


flair/datasets/sequence_labeling.py
```diff
@@ -646,7 +646,9 @@ def _convert_lines_to_sentence(
 
             # otherwise, this line is a token. parse and add to sentence
             token = self._parse_token(line, word_level_tag_columns, token)
-            sentence.add_token(token)
+            tokens.append(token)
+
+        sentence: Sentence = Sentence(text=tokens)
```
alanakbik, line 651: This looks much cleaner than creating an empty sentence and adding tokens. I wonder if this means we can retire the `add_token` method entirely in a separate PR?

helpmefindaname: Looking at the usage after this PR, we still have the following usages:

* Tests, which can be rewritten
* `UniversalDependenciesDataset` which works the same way as the ColumnDataset worked before
* Sentence - constructor, which I would prefer to keep.

We can surely make the `add_token` method private

helpmefindaname: I added a proposal to this PR

alanakbik:
@helpmefindaname thanks for adding this and for the add_token modifications!

Issue #3009, opened 2022-12-05
Error(s) in loading state_dict for SequenceTagger: Unexpected key(s) in state_dict: "lstm_init_h", "lstm_init_c".


ambuje:
from flair.models import SequenceTagger
tagger : SequenceTagger = SequenceTagger(hidden_size=128,
                                       embeddings=embeddings,dropout=0.2,locked_dropout=0.2,
                                       tag_dictionary=tag_ditionary1,train_initial_hidden_state=True,
                                       tag_type=tag_type,rnn_layers=4,tag_format='BIO',
                                       use_crf=False)

If train_initial_hidden_state=True, then we receive the error:- 

Error(s) in loading state_dict for SequenceTagger: 
Unexpected key(s) in state_dict: "lstm_init_h", "lstm_init_c".

Task - NER

Can you please help me out?

Thank you

helpmefindaname:
Hi @ambuje 
seems like this is currently not possible, I created a PR to fix it, can you try out the following: https://github.com/flairNLP/flair/pull/3010 and see if that works for you?

PR #3010, opened 2022-12-05
save initial hidden states in sequence tagger

helpmefindaname:
Small fix for https://github.com/flairNLP/flair/issues/3009


alanakbik:
@helpmefindaname thanks for fixing this and adding the test!

Issue #3018, opened 2022-12-11
ValueError When Loading TACRED


dobbersc:
**Describe the bug**
When loading TACRED the following ValueError will be thrown:

```
Traceback (most recent call last):
  File "/glusterfs/dfs-gfs-dist/dobbersc/.local/bin/miniconda3/envs/flair/lib/python3.9/code.py", line 90, in runcode
    exec(code, self.locals)
  File "<input>", line 1, in <module>
  File "/glusterfs/dfs-gfs-dist/dobbersc/PyCharmProjects/flair/flair/datasets/relation_extraction.py", line 260, in __init__
    super(RE_ENGLISH_TACRED, self).__init__(
  File "/glusterfs/dfs-gfs-dist/dobbersc/PyCharmProjects/flair/flair/datasets/sequence_labeling.py", line 403, in __init__
    super(ColumnCorpus, self).__init__(
  File "/glusterfs/dfs-gfs-dist/dobbersc/PyCharmProjects/flair/flair/datasets/sequence_labeling.py", line 296, in __init__
    [
  File "/glusterfs/dfs-gfs-dist/dobbersc/PyCharmProjects/flair/flair/datasets/sequence_labeling.py", line 297, in <listcomp>
    ColumnDataset(
  File "/glusterfs/dfs-gfs-dist/dobbersc/PyCharmProjects/flair/flair/datasets/sequence_labeling.py", line 501, in __init__
    sentence = self._convert_lines_to_sentence(
  File "/glusterfs/dfs-gfs-dist/dobbersc/PyCharmProjects/flair/flair/datasets/sequence_labeling.py", line 696, in _convert_lines_to_sentence
    key, value = comment_row.split("=", 2)
ValueError: too many values to unpack (expected 2)
```

**To Reproduce**

```python
from flair.datasets import RE_ENGLISH_TACRED
RE_ENGLISH_TACRED()
```

**Environment (please complete the following information):**
 - OS: openSUSE Leap 15.3
 - Version: flair on master

**Additional context**
The cause of this error is from the changes in #3006 on how comments are handled in the `ColumnDataset`.
https://github.com/flairNLP/flair/blob/5a13598ef5c20fdb79e441720b82250c0881094e/flair/datasets/sequence_labeling.py#L694

The generated `.conllu` files from the original TACRED corpus contain comments not supported by the currently implemented comment parser. Example (since TACRED is a private dataset):

```
[...]

PR #3020, opened 2022-12-12
fix comment parsing for conllu datasets

helpmefindaname:
regarding https://github.com/flairNLP/flair/issues/3018

`str.split(text, s)` creates a lsit with up to s+1 entries,
hence to have 2 entries, s needs to be 1.

Besides that, I suppose that the entry doesn't contain valid metadata if it starts with "#"

### DIFF
--- a/flair/datasets/sequence_labeling.py
+++ b/flair/datasets/sequence_labeling.py
@@ -693,7 +693,7 @@ class ColumnDataset(FlairDataset):
             # to set the metadata "domain" to "de-orcas"
             for comment_row in comment.split("\t"):
                 if "=" in comment_row:
-                    key, value = comment_row.split("=", 2)
+                    key, value = comment_row.split("=", 1)
                     sentence.add_metadata(key, value)
 
         if len(sentence) > 0:
--- a/flair/embeddings/token.py
+++ b/flair/embeddings/token.py
@@ -661,7 +661,7 @@ class FlairEmbeddings(TokenEmbeddings):
             "am-forward": f"{am_path}/best-lm.pt",
             # Ukrainian
             "uk-forward": "https://huggingface.co/dchaplinsky/flair-uk-forward/resolve/main/best-lm.pt",
-            "uk-backward": "https://huggingface.co/dchaplinsky/flair-uk-backward/resolve/main/best-lm.pt"
+            "uk-backward": "https://huggingface.co/dchaplinsky/flair-uk-backward/resolve/main/best-lm.pt",
         }
 
         if isinstance(model, str):
--- a/flair/embeddings/token.py
+++ b/flair/embeddings/token.py
@@ -664,7 +664,7 @@ class FlairEmbeddings(TokenEmbeddings):
             "uk-backward": "https://huggingface.co/dchaplinsky/flair-uk-backward/resolve/main/best-lm.pt"
         }
 
-        if type(model) == str:
+        if isinstance(model, str):
 
             # load model if in pretrained model map
             if model.lower() in self.PRETRAINED_MODEL_ARCHIVE_MAP:
--- a/flair/embeddings/token.py
+++ b/flair/embeddings/token.py
@@ -665,7 +665,6 @@ class FlairEmbeddings(TokenEmbeddings):
         }
 
         if isinstance(model, str):
-
             # load model if in pretrained model map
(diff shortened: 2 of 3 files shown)


dobbersc:
Thanks for the quick fix!

When testing this PR with TACRED I got no errors. But the comment metadata (`# sentence_id = 1`) is just skipped because the comment starts with a `#`. I'm not sure if there is a general common format for comments and their metadata. But looking at the CONLL-U specification, they also have variants of comment metadata, e.g. the example in [this section](https://universaldependencies.org/format.html#paragraph-and-document-boundaries).

In general, why don't we use the metadata parser from the conllu package? They produce the [`.metadata`](https://github.com/EmilStenstrom/conllu/#parse-metadata-from-a-conll-u-file) attribute in the parser.

alanakbik:
@helpmefindaname thanks for fixing this and @dobbersc thanks for reporting this error. I'm a bit confused though: why are all rows with `#` skipped? Should not all rows that begin with `#` and contain a ` = ` be evaluated for their metadata?

helpmefindaname:
@alanakbik currently the the parsing expects a tab separated line, starting with `#`. 

In the case of the multiconer it is: `# id cd27886d-6895-4d02-a8df-e5fa763fa88f	domain=de-orcas` Where we only look at the `domain=de-orcas` as it contains a `=`.
However this only fits only that specific format, and fails with others like @dobbersc pointed out.

@dobbersc I think you are right, we can use conll-u's [parse_command_line](https://github.com/EmilStenstrom/conllu/blob/master/conllu/parser.py#L142) and provide specific parsers for specific datasets.
However I will take some time before I start working on that. 

I don't know if that dataset bug is blocking you, but I suggest that I revert my last commit so the loading is fixed and will add another MR later

dobbersc:
> I don't know if that dataset bug is blocking you, but I suggest that I revert my last commit so the loading is fixed and will add another MR later

For me personally, the comment parsing is not blocking me. When working with TACRED I don't really care about the additional comment parsing logic, since the newly parsed comments are not required for model training. Only the relation comment is.

This PR fixes the `ValueError` when loading TACRED but parses the comments incorrectly, since as you noted only the multiconer comment format is supported. It would be great to support other formats in another PR, but no rush from my side :). 

I don't know if there are other datasets affected by this as well. I think, checking the comment parsing correctness before the next release may help.

<details><summary>
Example for future reference</summary>
<p>

```
# global.columns = id form ner
# text = Tom Thabane resigned in October last year to form the All Basotho Convention -LRB- ABC -RRB- , crossing the floor with 17 members of parliament , causing constitutional monarch King Letsie III to dissolve parliament and call the snap election .
# sentence_id = 61b3a5c8c9a882dcfcd2
# relations = 11;13;1;2;org:founded_by
1       Tom     B-PERSON
2       Thabane I-PERSON
3       resigned        O
4       in      O
5       October B-DATE
6       last    I-DATE
7       year    I-DATE
8       to      O
9       form    O
10      the     O
11      All     B-ORGANIZATION
12      Basotho I-ORGANIZATION
13      Convention      I-ORGANIZATION
14      (       O
15      ABC     B-ORGANIZATION
16      )       O
17      ,       O
18      crossing        O
19      the     O
20      floor   O
21      with    O
22      17      B-NUMBER
23      members O
24      of      O
25      parliament      O
26      ,       O
27      causing O
28      constitutional  O
29      monarch O
30      King    O
31      Letsie  B-PERSON
32      III     I-PERSON
33      to      O
34      dissolve        O
35      parliament      O
36      and     O
37      call    O
38      the     O
39      snap    O
40      election        O
41      .       O
```

Current extracted sentence metadata: 
```python
{
  "# global.columns ": " id form ner\n",
  "# text ": " Tom Thabane resigned in October last year to form the All Basotho Convention -LRB- ABC -RRB- , crossing the floor with 17 members of parliament , causing constitutional monarch King Letsie III to dissolve parliament and call the snap election .\n",
  "# sentence_id ": " 61b3a5c8c9a882dcfcd2\n",
  "# relations ": " 11;13;1;2;org:founded_by\n",
 }
```

</p>
</details>

dobbersc (approved):

alanakbik:
Thanks for the clarification! I'll merge now and we can look into a solution for multiconer.

alanakbik:
@helpmefindaname thanks for fixing this, and also fixing the unit tests!

PR #3030, opened 2022-12-15
HunFlair: Fix links to datasets that are no longer available

mariosaenger:
This PR fixes links to datasets that are no longer available at their original location. Instead the links now target the newly created hunflair-corpora repository.


alanakbik:
@mariosaenger thanks for fixing this!

PR #3039, opened 2022-12-24
Biomedical Entity Linking

piamarlene:
Adds functionality for entity linking using models pretrained on biomedical entities
by WBI group at HU Berlin


flair/models/biomedical_entity_linking.py
```diff
@@ -0,0 +1,988 @@
...
+        self.encoder = TransformerDocumentEmbeddings(
+            model_name_or_path, is_token_embedding=False
+        )
+
+        return self.encoder
+
+    def load_sparse_encoder(
+        self, model_name_or_path: Union[str, Path]
+    ) -> SparseWordVectorizer:
+        sparse_encoder_path = os.path.join(model_name_or_path, "sparse_encoder.pk")
```
sg-wbi, line 336: "sparse_encoder.pkl"?
I vaguely remember that we talked about some "security concerns" when loading pickle objects...

flair/models/biomedical_entity_linking.py
```diff
@@ -0,0 +1,988 @@
...
+
+        if show_progress:
+            iterations = tqdm(
+                range(0, len(names), batch_size),
+                desc="Calculating sparse embeddings for dictionary",
+            )
+        else:
+            iterations = range(0, len(names), batch_size)
+
+        for start in iterations:
```
sg-wbi, line 394: I think that in this case we need to pass all `names` at once otherwise we get different `_vocabulary` in the vectorizer.

flair/models/biomedical_entity_linking.py
```diff
@@ -0,0 +1,988 @@
...
+
+                # embed batch
+                self.encoder.embed(batch)
+
+                dense_embeds += [
+                    name.embedding.cpu().detach().numpy() for name in batch
+                ]
+
+                if flair.device.type == "cuda":
+                    torch.cuda.empty_cache()
```
sg-wbi, line 444: Is this necessary? I vaguely remember `empty_cache` to be an expensive operation.

sg-wbi (changes requested):
I took the liberty to add some comments.
The premise is that I am not very familiar w/ the flair API so be patient if I misunderstood some parts.

@leonweber @mariosaenger 
The high level questions are:
- do we need cosine similarity?
- are we allowed to use `pickle` to save the `sklearn` vectorizer?
- preprocessing (rm punctuation, lowercase) should be left to the user and not built-in right?

flair/models/biomedical_entity_linking.py
```diff
@@ -0,0 +1,988 @@
...
+        :return res: d numpy array of ids [# of query , # of dict]
+        :return scores: numpy array of top scores
+        """
+
+        res = None
+        scores = None
+        for i in tqdm(
+            np.arange(0, score_matrix.shape[0], batch_size), disable=not show_progress
+        ):
+            score_matrix_tmp = torch.tensor(
```
sg-wbi, line 493: Why go to torch if then you go back to numpy?

sg-wbi: There's a vectorized implementation [here](https://github.com/dmis-lab/BioSyn/blob/master/src/biosyn/biosyn.py#L157)

flair/models/biomedical_entity_linking.py
```diff
@@ -0,0 +1,988 @@
...
+        :param mention: string of the mention to find candidates for
+        :param topk: number of candidates
+        :return res: d numpy array of ids [# of query , # of dict]
+        :return scores: numpy array of top predictions and their scores
+        """
+        # get dense embeds for mention
+        mention_dense_embeds = self.embed_dense(names=[mention])
+
+        # remove mean centering
+        if tgt_space_mean_vec is not None:
```
sg-wbi, line 528: I guess we can remove this

flair/models/biomedical_entity_linking.py
```diff
@@ -0,0 +1,988 @@
...
+            mention_dense_embeds -= tgt_space_mean_vec
+
+        # normalize mention vector
+        faiss.normalize_L2(mention_dense_embeds)
```
sg-wbi, line 532: We do not need L2 normalization unless we are using cosine similarity (see [here](https://github.com/facebookresearch/faiss/wiki/MetricType-and-distances#how-can-i-index-vectors-for-cosine-similarity))

flair/models/biomedical_entity_linking.py
```diff
@@ -0,0 +1,988 @@
...
+            else:
+                self.dict_dense_embeds = self.embed_dense(
+                    names=dictionary_names, show_progress=True
+                )
+                self.dict_sparse_embeds = None
+
+            # build dense index
+            dimension = self.dict_dense_embeds.shape[1]
+            # to use cosine similarity, we normalize the vectors and then use inner product
+            faiss.normalize_L2(self.dict_dense_embeds)
```
sg-wbi, line 656: Ok, we should the have a `use_cosine_similarity` arg for `BiEncoder` to decide which metric to use. However, from my experience IP works just fine

flair/models/biomedical_entity_linking.py
```diff
@@ -0,0 +1,988 @@
...
+            log.info(
+                "Loaded dictionary from cached file {}".format(cached_dictionary_path)
+            )
+
+            (self.dictionary, self.dict_sparse_embeds, self.dense_dictionary_index,) = (
+                cached_dictionary["dictionary"],
+                cached_dictionary["sparse_dictionary_embeds"],
+                cached_dictionary["dense_dictionary_index"],
+            )
+            self.dense_dictionary_index = faiss.index_cpu_to_gpu(
```
sg-wbi, line 696: this should be protected by `use_cuda`

flair/models/biomedical_entity_linking.py
```diff
@@ -0,0 +1,988 @@
...
+        elif dictionary_name_or_path == "ncbi-taxonomy":
+            self.dictionary = NEL_NCBI_TAXONOMY_DICT().data
+        # use custom dictionary file
+        else:
+            self.dictionary = DictionaryDataset(
+                dictionary_path=dictionary_name_or_path
+            ).data
+
+    def cache_dictionary(self, cache_index_dense, cache_folder, cached_dictionary_path):
+        cached_dictionary = {
```
sg-wbi, line 717: IMO there is no need to save "dictionary", this is just the output of reading a file w/ basically no preprocessing. For "dense_dictionary_emebds" we should use `torch.save` or whathever `flair` uses. For "sparse" there is still the question about using pickle...

flair/models/biomedical_entity_linking.py
```diff
@@ -0,0 +1,988 @@
...
+            cache_folder,
+            f"{file_name}.pk",
+        )
+
+        # If exists, load the cached dictionary indices
+        if os.path.exists(cached_dictionary_path):
+            self.load_cached_dictionary(cached_dictionary_path)
+
+        # else, load and embed
+        else:
```
sg-wbi, line 631: Here is where we can us `TextProcess` to rm punctuation and what not, w/ a parameter in the fn

PR #3043, opened 2023-01-04
Release optimizer memory and fix legacy tokenization

helpmefindaname:
This PR does 2 little things:

* it makes it easier to clean up the gpu memory, by cleaning gradients and deleting the reference of the optimzier. Hence `torch.cuda.empty_cache()` will be able to clean up more parameters. 
* some (slow) tokenizers like RobertaTokenizer seem to add a [SEP] token at the end, breaking the legacy subword token mapping, which is fixed by allowing that specific token


helpmefindaname:
With those changes, I managed to train XLM-Roberta-Large (>2GB model) with adam and full precision (4x the memory requirement) on my 6GB laptop graphic card, using [transformer-smaller-training-vocab](https://github.com/helpmefindaname/transformer-smaller-training-vocab)

PR #3051, opened 2023-01-16
Replace issue templates with issue forms

helpmefindaname:
Suggestion to use the newer [Issue Forms](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/syntax-for-issue-forms) from github.
That way, certain information will be mandatory to add and formatting (for example code for reproduction) will be added automatically.

The issue forms can be tested at [my fork](https://github.com/helpmefindaname/flair/issues) where you can try it out by creating new issues


alanakbik:

PR #3052, opened 2023-01-16
fix label alignment if the sentence contains invalid tokens

helpmefindaname:


flair/datasets/sequence_labeling.py
```diff
@@ -661,6 +660,7 @@ def _convert_lines_to_sentence(
                     bioes_tags = [
                         re.split(self.column_delimiter, line.rstrip())[span_column] for line in filtered_lines
                     ]
+                    bioes_tags = [tag for tag, token in zip(bioes_tags, tokens) if token._internal_index is not None]
```
alanakbik, line 663: Could you add some comments explaining what happens here? I fear that at some point in the future someone might look at this and wonder why this filter exists.

alanakbik (changes requested):
Thanks for fixing this! See comment for request for comment.

flair/datasets/sequence_labeling.py
helpmefindaname: The comment is added

alanakbik:
Thanks @helpmefindaname!

PR #3054, opened 2023-01-17
Removed a name that is undefined

openrefactory:
In file: __init__.py, the list named __all__ contains an undefined name which can result in errors when this module is imported. We removed the undefined name from the list. 

This is following the best practices of importing fields from a package (https://docs.python.org/3/tutorial/modules.html#importing-from-a-package).


alanakbik:
@openrefactory thanks for fixing this!

PR #3055, opened 2023-01-18
added weighted sum layer

whoisjones:
see above


alanakbik:

PR #3057, opened 2023-01-18
Call `super().__setstate__()` in Embeddings

SimonZehetner:
When loading a model that was trained with an older torch version, it is possible that an `AttributeError` occurs. This can happen if new attributes were added to `torch.nn.Module`.

For backwards compatibility, they implemented their `__setstate__` to add these attributes if they are not present (https://github.com/pytorch/pytorch/blob/ae4ec7de1e6f60aa7dcf4d5187c720f2f40543c3/torch/nn/modules/module.py#L1564).

In oder to be independent from these changes and increase backwards compatibility, it would be good, if the super-method would be called when unpickling Embeddings instead of setting `__dict__`.

### DIFF
--- a/flair/embeddings/document.py
+++ b/flair/embeddings/document.py
@@ -426,7 +426,7 @@ class DocumentRNNEmbeddings(DocumentEmbeddings):
             self.eval()
 
         else:
-            self.__dict__ = d
+            super().__setstate__(d)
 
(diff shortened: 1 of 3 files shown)


alanakbik:
@SimonZehetner thanks for fixing this!

PR #3060, opened 2023-01-20
remove konoha from requirements.txt

marctorsoc:


alanakbik:
Hello @marctorsoc, thanks for the PR! The tests are failing since we test for Japanese tokenization. That test would need to be removed. You should also run black formatting (see [here](https://github.com/flairNLP/flair/blob/master/CONTRIBUTING.md#code-formatting)) to pass the checks.

marctorsoc:
@alanakbik this should be fixed now, I think

alanakbik:
@marctorsoc thanks for fixing this!

marctorsoc:
> @marctorsoc thanks for fixing this!

For what I see in https://github.com/flairNLP/flair/blob/master/.github/workflows/ci.yml, there's no automatic release. Could you release a new version after this? Do you plan to add auto-release when a PR is merged to master, or you prefer to keep manual like now?

alanakbik:
We'll try to release a new version soon, and after that switch to more frequent updates!

Issue #3061, opened 2023-01-20
[Bug]: train() fails if corpus without dev split, even with train_with_dev=False


whoisjones:
### Describe the bug

Line 717 in trainer.py: I do not want anneal against some dev loss since i do not have a dev split

### To Reproduce

```python
from flair.datasets import ColumnCorpus

main():
support_set = ColumnCorpus(data_folder="data/fewshot/conll03/", train_file=f"{split}.txt", sample_missing_splits=False, column_format={0: "text", 1: "ner"})

...

trainer.fine_tune(train_with_dev=False)


if __name__ == "__main__":
main()
```

### Expected behaivor

training should run through and ignore dev split / score

### Logs and Stack traces

```stacktrace
Traceback (most recent call last):
  File "/vol/fob-vol7/mi18/goldejon/flair/flair/trainers/trainer.py", line 717, in train
    if dev_score > best_validation_score:
UnboundLocalError: local variable 'dev_score' referenced before assignment
```

##### Flair
0.11.3
##### Pytorch
1.12.1
##### Transformers
4.23.1
#### GPU
True

PR #3062, opened 2023-01-20
Fix train error with missing dev split (gh-3061)

whoisjones:
closes gh-3061.


alanakbik (changes requested):
Thanks @whoisjones! Can you add a unit test to the PR that tests for this error?

Issue #3024, opened 2022-12-13
Wrong context preperation in SequenceTagger.predict()


alzoubi36:
Hi everyone,

I finetuned a TransformerWordEmbeddings based on 'microsoft/mdeberta-v3-base' for a NER task and set use_context=True. I noticed some discrepancy between the test that happens automatically after the training (implemented by flair) and the tests that I do with SequenceTagger manually afterwards. After debugging I noticed that the sentences in a batch are being reordered in SequenceTagger.predict():

https://github.com/flairNLP/flair/blob/5a13598ef5c20fdb79e441720b82250c0881094e/flair/models/sequence_tagger_model.py#L465

So the context that is being expanded from this batch is wrong. While the context that is being prepared in the automatic test after training is correct (no sentence reordering happens and SequenceTagger._prepare_tensors(batch) is directly called).

Can you let me know if my observation is correct or if I missed something?

alanakbik:
Hello @alzoubi36 thanks for noticing this - it is indeed a big problem that needs to be fixed. The sorting is done to speed-up inference, but when document context needs to be inferred on the fly, the new ordering messes up the contexts. The simplest solution would be to remove the sorting, but this would lead to slow-downs for models that don't require context. We have to think how to best address this.

helpmefindaname:
I think a hotfix could be to first call the embedding and then run the predictions, e.g.
```
model.embeddings.embedd(sentences)
model.predict(sentences)
```
or to just set the context beforehands:
```
for first, second in zip(sentences, sentences[1:]):
    first._next_sentence = second
    second._first_sentence = first
```

but I agree that this is something to be fixed in the library.
I also noticed in the code for adding the context:
There is currently no way to specify that a sentence is the first one and has no `left_context`, as the check if context is set is `self._previous_sentence is not None`.

@alanakbik  Maybe we can move that logic to the .predict method, such that the embeddings always expect the context to be set already?

alanakbik:
> @alanakbik Maybe we can move that logic to the .predict method, such that the embeddings always expect the context to be set already?

That would be a solution, however it would mean adding this logic to every predict method in Flair (the `SequenceTagger` uses a different one from `DefaultClassifier`) and also this introduces some overhead for models that don't require context (though probably setting the context is not so expensive).

helpmefindaname:
This is fixed by the PR above

PR #3063, opened 2023-01-23
Improved Tars Context

helpmefindaname:
This PR is fixing https://github.com/flairNLP/flair/issues/3024 and adds the following changes while doing so:

* tars models (classifcation & ner) won't fail on labels with only 1 character.
* models that predict on modified sentences (`TARSClassifier`, `TARSTagger` and `RelationClassifier`) now also make use of the context and therefore can be trained and tested in a flert manner.
* Empty sentences will not be discarded in the context chain, however they won't change the context as they are empty.
* it is possible to disable context, by setting `sentence._has_context = True` without setting `sentence._next_sentence` or `sentence._previous_sentence`. That way, batch prediction on context independent sentences will work.
* context is considered to be set if only "_next_sentence" is set. So every first sentence won't be reset again

### DIFF
--- a/flair/data.py
+++ b/flair/data.py
@@ -713,6 +713,7 @@ class Sentence(DataPoint):
 
         # internal variables to denote position inside dataset
         self._previous_sentence: Optional[Sentence] = None
+        self._has_context: bool = False
         self._next_sentence: Optional[Sentence] = None
         self._position_in_dataset: Optional[typing.Tuple[Dataset, int]] = None
 
@@ -1064,7 +1065,25 @@ class Sentence(DataPoint):
         Return True or False depending on whether context is set (for instance in dataloader or elsewhere)
         :return: True if context is set, else False
         """
-        return self._previous_sentence is not None or self._position_in_dataset is not None
+        return self._has_context or self._previous_sentence is not None or self._position_in_dataset is not None
+
+    def copy_context_from_sentence(self, sentence: "Sentence") -> None:
+        self._previous_sentence = sentence._previous_sentence
+        self._next_sentence = sentence._next_sentence
+        self._position_in_dataset = sentence._position_in_dataset
+
+    @classmethod
+    def set_context_for_sentences(cls, sentences: List["Sentence"]) -> None:
+        previous_sentence = None
+        for sentence in sentences:
+            if sentence.is_context_set():
+                continue
+            sentence._previous_sentence = previous_sentence
+            sentence._next_sentence = None
+            sentence._has_context = True
+            if previous_sentence is not None:
+                previous_sentence._next_sentence = sentence
+            previous_sentence = sentence
 
     def get_labels(self, label_type: str = None):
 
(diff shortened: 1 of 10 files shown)


alanakbik:

The TARS context expansion only partly works, however. The problem is that TransformerBaseEmbeddings' `__expand_sentence_with_context` is called after the TARS sentence is created. This puts the context around the sentence + TARS label. 

To illustrate, run this script: 

```python
# init the TARS sequence tagger
from flair.data import Sentence, Dictionary
from flair.datasets import CONLL_03
from flair.models import FewshotClassifier, TARSTagger

# quick init of corpus and label dictionary
corpus = CONLL_03(in_memory=False)
label_dictionary = Dictionary()
label_dictionary.add_item('LOC')

# init transformer embeddings with context
embeddings: TransformerWordEmbeddings = TransformerWordEmbeddings('distilbert-base-uncased', use_context=5)

# init few-shot tagger
tart_tagger: FewshotClassifier = TARSTagger(
    embeddings=embeddings,
    num_negative_labels_to_sample=1,
    prefix=True,
    task_name='ner',
    label_type='ner',
)

# get tars formatted sentence
sentence = corpus.train[1]
print("\n - This is the original sentence:")
print(sentence.text)

tars_sentence: Sentence = tart_tagger._get_tars_formatted_sentence('LOC', sentence)

print("\n - This is the TARS sentence:")
print(tars_sentence.text)

# expand sentence as done in def __expand_sentence_with_context(self, sentence)
left_context = tars_sentence.left_context(5)
right_context = tars_sentence.right_context(5)
expanded_sentence = left_context + tars_sentence.tokens + right_context

print("\n - This is the expanded TARS sentence:")
print(" ".join(token.text for token in expanded_sentence)) 
```

This will print: 

```console
 - This is the original sentence:
Peter Blackburn

 - This is the TARS sentence:
LOC [SEP] Peter Blackburn

 - This is the expanded TARS sentence:
to boycott British lamb . LOC [SEP] Peter Blackburn BRUSSELS 1996-08-22 The European Commission
```

But the expanded TARS sentence is suboptimal. It should be
- `LOC [SEP] to boycott British lamb . Peter Blackburn BRUSSELS 1996-08-22 The European Commission`, instead of 
- `to boycott British lamb . LOC [SEP] Peter Blackburn BRUSSELS 1996-08-22 The European Commission`

I am not sure what the best solution here could be. One idea that comes to mind is to separate context from main text through separators. I.s. something like: 
-  `to boycott British lamb . [CONTEXT] LOC [SEP] Peter Blackburn [CONTEXT] BRUSSELS 1996-08-22 The European`, or
-  `to boycott British lamb . [SEP] LOC [SEP] Peter Blackburn [SEP] BRUSSELS 1996-08-22 The European` ...

What do you think?

alanakbik:
@helpmefindaname thanks for adding this! I'll merge this now and may fix the context issue later in a separate PR.

Issue #3072, opened 2023-01-27
[Bug]: RuntimeError when do onnx export


OleksandrKorovii:
### Describe the bug

I try reproduce Tutorial 13 and get next error RuntimeError

### To Reproduce

```python
from flair.models import SequenceTagger
from flair.embeddings import TransformerWordEmbeddings, TransformerDocumentEmbeddings

model = SequenceTagger.load("ner_model.pt")
assert isinstance(model.embeddings, (TransformerWordEmbeddings, TransformerDocumentEmbeddings))

from flair.datasets import CONLL_03
sentences = list(CONLL_03().test)[:5]

model.embeddings = model.embeddings.export_onnx("Ner-Model-Embeddings.onnx", sentences, providers=["CUDAExecutionProvider", "CPUExecutionProvider"])
```

### Expected behaivor

The embeddings convert into onnx format

### Logs and Stack traces

```stacktrace
File ~/ELL/flair-venv/lib/python3.10/site-packages/flair/embeddings/transformer.py:1279, in TransformerEmbeddings.forward(self, input_ids, sub_token_lengths, token_lengths, attention_mask, overflow_to_sample_mapping, word_ids, langs, bbox, pixel_values)
   1271     all_token_embeddings[:, :, sentence_hidden_states.shape[2] :] = fill_masked_elements(
   1272         all_token_embeddings[:, :, sentence_hidden_states.shape[2] :],
   1273         sentence_hidden_states,
   (...)
   1276         token_lengths,
   1277     )
   1278 elif self.subtoken_pooling == "mean":
-> 1279     all_token_embeddings = fill_mean_token_embeddings(
   1280         all_token_embeddings, sentence_hidden_states, word_ids, token_lengths
   1281     )
   1282 else:
   1283     raise ValueError(f"subtoken pooling method: `{self.subtoken_pooling}` is not implemented")

File ~/ELL/flair-venv/lib/python3.10/site-packages/torch/jit/_trace.py:1139, in _script_if_tracing.<locals>.wrapper(*args, **kwargs)
   1136     return fn(*args, **kwargs)
   1138 compiled_fn = script(wrapper.__original_fn)  # type: ignore[attr-defined]
-> 1139 return compiled_fn(*args, **kwargs)

[...]

PR #3075, opened 2023-01-30
Gh 3072/transformer mean pooling

helpmefindaname:
fixes https://github.com/flairNLP/flair/issues/3072
also fixes the `--force-cpu` flag for tests


alanakbik:

PR #3086, opened 2023-02-07
Add CoNLL 2012 shared task

plonerma:
This PR extends ColumnDataset in order to read data from the CoNLL 2012 shared task and adds this dataset to `flair.datasets.sequence_labeling` (see https://cemantix.org/data/ontonotes.html for details).


flair/datasets/sequence_labeling.py
```diff
@@ -628,6 +661,9 @@ def _read_next_sentence(self, file):
             if len(lines) > 0 and self.__line_completes_sentence(line):
                 break
 
+            if line.strip() == self.document_separator_line:
+                break
```
alanakbik, line 665: What does this do? Can you add a comment?

flair/datasets/sequence_labeling.py
```diff
@@ -638,23 +674,31 @@ def _convert_lines_to_sentence(
...
-            # otherwise, this line is a token. parse and add to sentence
-            token = self._parse_token(line, word_level_tag_columns, token)
-            tokens.append(token)
-        sentence: Sentence = Sentence(text=tokens)
+        sentence: Sentence
 
-        # check if this sentence is a document boundary
-        if sentence.to_original_text() == self.document_separator_token:
+        if len(lines) == 1 and lines[0].strip() == self.document_separator_line:
+            sentence = Sentence(text=[Token(lines[0])])
```
alanakbik, line 681: `lines[0]` may not necessarily be the token text (some datasets have text in the second column). `self.text_column` stores in which column the text is.

flair/datasets/sequence_labeling.py
```diff
@@ -435,9 +444,12 @@ def __init__(
         banned_sentences: List[str] = None,
         in_memory: bool = True,
         document_separator_token: str = None,
+        document_separator_line: str = None,
         encoding: str = "utf-8",
         skip_first_line: bool = False,
         label_name_map: Dict[str, str] = None,
+        preprocess_bioes_tags: Callable[[List[str]], List[str]] = None,
+        span_level_tag_columns: List[int] = None,
```
alanakbik, line 452: This argument needs to be passed on to the three `ColumnDataset` calls below, otherwise it is not used.

flair/datasets/sequence_labeling.py
alanakbik: Plus, could you add a comment to explain this part of the refactor? Is this for speedup purposes?

alanakbik (changes requested):
Thanks a lot for adding this! There are some small issues, so the NER tags are not parsed, reproducible by running this: 

```python
# get Zelda corpus and print statistics
corpus = CONLL_2012(in_memory=False)
print(corpus)

# get an example sentence and print
sentence = corpus.test[1]
print(sentence)

# iterate over linked entities in sentence and print each
for entity in sentence.get_labels('ner'):
    print(entity) 
```

Additionally, the corpus should be added to the `__init__.py` of the datasets module so that it can be loaded with
```python
from flair.datasets import CONLL_2012
```

flair/datasets/sequence_labeling.py
plonerma: I am assuming document boundaries are separate sentences. Since there is no newline after the document separator line in CoNLL2012, the current sentence (i.e. document boundary) is completed directly.
This code should be moved to ColumnDataset.__line_completes_sentence (I fixed this in one of the commits).

flair/datasets/sequence_labeling.py
plonerma: In CoNNL 2012, document boundaries are not marked by a specific token, but a line ("#end document" without any of the other columns -> treating this line as a token will result in a parsing error). I added this as a comment.

Regarding the index: lines[0] should be the only line in lines (with len 1), its not the first column in this line.

flair/datasets/sequence_labeling.py
plonerma: Yes, thanks for pointing that out. I missed that when transferring the code and did not spot the error: This is also the source of the issue with the parsing of the NER tags (fixed in the commits).

plonerma:
Thanks for your feedback. The NER parsing should now be fixed.

plonerma:
I now also extended the dataset to only load specific domains (newswire, broadcast news, etc), and/or sources (cnn, msnbc, etc.).

E.g.:

```python
# load complete dataset
corpus = CONLL_2012(in_memory=False)

# only load data from the broadcast conversation domain
corpus = CONLL_2012(in_memory=False, domain="bc")

# only load data from the broadcast conversation and news domains
corpus = CONLL_2012(in_memory=False, domain=["bc", "nw"])

# only load data from specific sources in one domain
corpus = CONLL_2012(in_memory=False, domain="bc", source=["cnn", "msnbc"])

# only load data from specific sources in multiple domains
corpus = CONLL_2012(in_memory=False, source={"bc": ["cnn", "msnbc"], "nw": "wsj"})
```

plonerma:
Same goal as: https://github.com/flairNLP/flair/pull/3094 (which may be more complete)

PR #3087, opened 2023-02-07
Add support for NERMuD 2023 Dataset

stefan-it:
Hi,

this PR adds support for the NERMuD 2023 Dataset. This dataset is a task presented at [EVALITA 2023](https://www.evalita.it/campaigns/evalita-2023/) consisting in the extraction and classification of named-entities in a document, such as persons, organizations, and locations.

From the Shared Task page:

NERMuD 2023 will include two different sub-tasks:

* Domain-agnostic classification (DAC). Participants will be asked to select and classify entities among three categories (person, organization, location) in different types of texts (news, fiction, political speeches) using one single general model.

* Domain-specific classification (DSC). Participants will be asked to deploy a different model for each of the above types, trying to increase the accuracy for each considered type.

For this purpose, the added `NER_NERMUD` dataset is implemented as a Multi Corpus. That means, different corpora (domains in this case) can be used and combined.

## Usage

To use all domains, the following example can be used:

```python
from flair.datasets import NER_NERMUD

all_nermud_domains = NER_NERMUD(domains="all")
```

It is also possible to combine domains, such as:

```python
from flair.datasets import NER_NERMUD

combined_domains = NER_NERMUD(domains=["WN", "ADG"])
```

Possible domains are:

* `WN` - Wikinews
* `FIC` - Fiction
* `ADG`  - De Gasperi

More information can be found [here](https://github.com/dhfbk/KIND/tree/main/evalita-2023).


alanakbik:

PR #3100, opened 2023-02-10
Add label noise simulation

aynetdia:
This PR enables uniform label noise simulation in any dataset available in Flair via an additional function in the Corpus class.


alanakbik:
The black checks are failing, can you do `black --config pyproject.toml flair/ && isort flair/` and `black --config pyproject.toml tests/ && isort tests/` in the project root?

Issue #3097, opened 2023-02-10
[Bug]: Multitask evaluation (and therefore training) fails on current master.


helpmefindaname:
### Describe the bug

The Multitask training fails when evaluating, as it now gets a `str` for the label_type, while still expecting a dictionary.

### To Reproduce

```python
from flair.embeddings import TransformerWordEmbeddings
from flair.trainers import ModelTrainer
from flair.models import SequenceTagger, RelationExtractor, MultitaskModel

corpus = create_corpus()  # any corpus with multiple tasks annotated should work.

embeddings = TransformerWordEmbeddings(
        model="bert-base-cased",
    )

ner_model = SequenceTagger(
    embeddings=embeddings,
    tag_dictionary=corpus.make_label_dictionary("ner", add_unk=False),
    tag_type="ner",
)
rel_model = RelationExtractor(
    embeddings=embeddings,
    label_dictionary=corpus.make_label_dictionary("rel", add_unk=True),
    entity_label_type="ner",
    label_type="rel",
)

model = MultitaskModel([ner_model, rel_model])

trainer = ModelTrainer(model, corpus)

trainer.train("training-path")
```

### Expected behaivor

A training should run trough without any issues and produce a trained model

### Logs and Stack traces

```stacktrace
File "/mnt/raid/bfuchs/.cache/pypoetry/virtualenvs/xDHSzyCG-py3.8/lib/python3.8/site-packages/clearml/binding/hydra_bind.py", line 173, in _patched_task_function
    return task_function(a_config, *a_args, **a_kwargs)
  File "/mnt/raid/bfuchs/train.py", line 203, in main
    results = trainer.fine_tune(
  File "/mnt/raid/bfuchs/.cache/pypoetry/virtualenvs/xDHSzyCG-py3.8/lib/python3.8/site-packages/flair/trainers/trainer.py", line 946, in fine_tune
    return self.train(
  File "/mnt/raid/bfuchs/.cache/pypoetry/virtualenvs/xDHSzyCG-py3.8/lib/python3.8/site-packages/flair/trainers/trainer.py", line 643, in train
    dev_eval_result = self.model.evaluate(
  File "/mnt/raid/bfuchs/.cache/pypoetry/virtualenvs/xDHSzyCG-py3.8/lib/python3.8/site-packages/flair/models/multitask_model.py", line 160, in evaluate
    + self.label_type.get(task_id)
```

### Additional Context

[...]

PR #3101, opened 2023-02-13
gh-3097: fix multitask model training

helpmefindaname:
This PR fixes https://github.com/flairNLP/flair/issues/3097 by introducing an `evaluate_all` parameter. That way, users can use `multitask_model.evaluate(corpus.test, gold_label_type="Task_0", evaluate_all=False)` to evaluate the first task. If `evaluate_all` is set to true, the value of `gold_label_type` will be ignored and every task will be evaluated on their respective label type.

While doing this implementation I also noticed, that the evaluation currently evaluates sentences that have multiple tasks annotated to be evaluated for only one random task. -> now the evaluation uses all sentences for all tasks they are assigned to as it should be expected.

I also added the possibilities to compute the loss of a sentence on all assigned tasks during training. Hence Training Knowledge Graph Construction (NER + Relation Extraction + NEL) with a shared transformer embedding, would lead to only slightly increased training time compared to NER only, while leaveraging information of all 3 tasks at any time.

I found a bug where the embeddigns where not embedded right, as the `identify_dynamic_embeddings` only looks at the first sentence of the batch which could be unenbedded (e.g. for relation extraction or NEL but has no NER labels). Now the logic searches the whole batch and continues searching until it finds a batch that contains ANY embedding.

### DIFF
--- a/flair/embeddings/base.py
+++ b/flair/embeddings/base.py
@@ -45,7 +45,7 @@ class Embeddings(torch.nn.Module, Generic[DT]):
         if not isinstance(data_points, list):
             data_points = [data_points]
 
-        if not self._everything_embedded(data_points) or not self.static_embeddings:
+        if not self._everything_embedded(data_points):
             self._add_embeddings_internal(data_points)
 
         return data_points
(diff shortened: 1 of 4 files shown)


alanakbik:
@helpmefindaname thanks a lot for fixing and improving this!

Issue #3092, opened 2023-02-08
[Feature]: Add OntoNotes corpus


whoisjones:
### Problem statement

a flair dataset for ontonotes.

### Solution

-

PR #3102, opened 2023-02-13
GH-3092/add ontonotes corpus

whoisjones:
closes gh-3092, ready for review. @plonerma @alanakbik

can be called like this:
```
corpus = ONTONOTES() # all files
corpus = ONTONOTES(domain=["bc"]) # only for domain BC
corpus = ONTONOTES(domain=["bc", "bn"]) # multiple domains
corpus = ONTONOTES(language="arabic") # different languages

corpus.make_label_dictionary("ner")
```

Currently supports NER and POS but all lines should be parsed and can be included in the column_map in the init. However, someone should take a look regarding correctness before integrating, I've made some quality checks for NER nad POS.

Last thing, do we have idea for properly testing downloading and extracting, folder structure setup without loading all data?


plonerma:
Regarding the testing of download and extraction:

I'm not sure what's the best way to test the download function. I guess we could check whether the URL is reachable -- do you think that makes sense? Otherwise, the download function relies only on two flair-function calls (`unpack_file` and `cached_path`, the second is tested in UD_ENGLISH, I'm not sure if `unpack_file` is tested anywhere).

To test the processing, it would probably make sense to include a few sample files in the `tests/resources` directory. By using `base_path` (and a temporary dir) it should then be possible to test the parsing on these sample files.

plonerma:
The [Best Configuration per Dataset](https://github.com/flairNLP/flair/blob/master/resources/docs/EXPERIMENTS.md#ontonotes-named-entity-recognition-english) already contains a train script for a (flattened) ontonotes NER task. This should probably also be updated to use the newly implemented dataset class.

alanakbik:
It may be necessary to merge master into this branch as this fixes the failing multitask tests. Also remember to do `black --config pyproject.toml flair/ && isort flair/` and `black --config pyproject.toml tests/ && isort tests/` in the project root to make sure the formatting tests are passed.

plonerma (approved):

alanakbik (changes requested):
Thanks for adding this! English and Chinese works, but there seems to be a problem with Arabic. 

To reproduce: 

```python
corpus = ONTONOTES(language="arabic", in_memory=False)
print(corpus)

# get a sentence
sentence = corpus.train[1]

# print the text
print(sentence.text)

# print all annotations
# print(sentence) 
```

This prints not only the text of a sentence, but also some additional tags.

alanakbik:
@whoisjones @plonerma can you check?

whoisjones:
depends on gh-3107, otherwise pre-commit hooks would run through.
 
 output for Arabic is now like this:
 ```
corpus = ONTONOTES(language="arabic", in_memory=False)
2023-02-15 12:04:03,125 OntoNotes splits for v4/arabic have not been generated yet, generating it now.
2023-02-15 12:04:03,125 Generating train split for v4/arabic
2023-02-15 12:04:04,968 Generating development split for v4/arabic
2023-02-15 12:04:05,233 Generating test split for v4/arabic
sentence = corpus.train[0]
print(sentence.text)
" سَيّاف أَبُو " عَلَى -القَضاءِ لِ- تَدْرِيبِيَّةٍ بِعْثَةٍ فِي اليَوْمِ مِن الفِيلِيبِّين إِلَى أَمِيرْكِيّاً جُنْدِيّاً 650

```

alanakbik:
@whoisjones @plonerma thanks for adding this!

PR #3125, opened 2023-02-24
GH-3124: Add class-dependent label noise generation

### LINKED ISSUE #3124: [Feature]: Class-dependent label noise simulation
### Problem statement

Currently Flair allows to generate only uniform label noise, which does not account for any similarities between the data points in the dataset. Generating class-dependent label noise would allow for a more realistic scenario, in which models have to deal with labelling mistakes/inaccuracies.

### Solution

Extend the `_corrupt_labels()` method of the `Corpus` by allowing to pass a pre-defined noise transition matrix.

### Additional Context

_No response_

aynetdia:
This PR closes #3124 . Noise transition matrix hast to be in a format of a dict of lists, in which the keys reflect the target label values.

### DIFF
--- a/flair/data.py
+++ b/flair/data.py
@@ -1476,7 +1476,7 @@ class Corpus(typing.Generic[T_co]):
 
         return label_dictionary
 
-    def _corrupt_labels(
+    def add_label_noise( 
         self,
         label_type: str,
         labels: List[str],
--- a/flair/data.py
+++ b/flair/data.py
@@ -1476,18 +1476,24 @@ class Corpus(typing.Generic[T_co]):
 
         return label_dictionary
 
-    def _corrupt_labels(self, noise_share: float, label_type: str, labels: List[str], split: str = "train"):
+    def _corrupt_labels(
+        self,
+        label_type: str,
+        labels: List[str],
+        noise_share: float = 0.2,
+        split: str = "train",
+        noise_transition_matrix: Optional[Dict[str, List[float]]] = None,
+    ):
         """
         Generates uniform label noise distribution in the chosen dataset split.
-        :noise_share: the desired share of noise in the train split.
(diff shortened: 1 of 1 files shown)


flair/data.py
```diff
@@ -1476,19 +1476,25 @@ def make_label_dictionary(self, label_type: str, min_count: int = -1, add_unk: b
 
         return label_dictionary
 
-    def _corrupt_labels(self, noise_share: float, label_type: str, labels: List[str], split: str = "train"):
+    def _corrupt_labels(
```
alanakbik, line 1479:
```suggestion
    def add_label_noise( 
```

alanakbik: this function should be made public

alanakbik (changes requested):
Thanks! Everything works, though I think the function could be made public.

alanakbik:
Info: tested with

```python
# make a corpus of 20 sentences
sentences = []
for i in range(10):
    sentence = Sentence("Berlin")
    sentence[0].add_label('ner', 'CITY')
    sentences.append(sentence)
for i in range(10):
    sentence = Sentence("Jens")
    sentence[0].add_label('ner', 'PERSON')
    sentences.append(sentence)

corpus = Corpus(train=FlairDatapointDataset(sentences))
print(corpus)

label_dictionary = corpus.make_label_dictionary('ner', add_unk=False)

print(label_dictionary.get_items())

corpus._corrupt_labels(noise_share=0.5,
                       label_type='ner',
                       labels=label_dictionary.get_items(),
                       noise_transition_matrix={'PERSON': [0.0, 1.0], 'CITY': [1.0, 0.0]},
                       )

for sentence in corpus.train:
    print(sentence)
```

alanakbik (approved):

alanakbik:
@aynetdia thanks for adding this!

PR #3127, opened 2023-02-27
reduce testing time

helpmefindaname:
This PR is here to evaluate the testing times and see which tests can be speed up or removed without loss of testing quality


flair/data.py
```diff
@@ -1434,25 +1434,25 @@ def make_label_dictionary(self, label_type: str, min_count: int = -1, add_unk: b
...
+                        label_dictionary.span_labels = True
+                        break
 
             if not label_dictionary.multi_label:
                 if len(labels) > 1:
                     label_dictionary.multi_label = True
 
         # if an unk threshold is set, UNK all label values below this threshold
-        erfasst_count = 0
+        total_count = 0
```
alanakbik, line 1450: oops, thanks for correcting this

flair/data.py
```diff
@@ -1434,25 +1434,25 @@ def make_label_dictionary(self, label_type: str, min_count: int = -1, add_unk: b
...
-                # check if there are any span labels
-                if type(label.data_point) == Span and len(label.data_point) > 1:
-                    label_dictionary.span_labels = True
+            label_value_counter.update(label.value for label in labels if label.value not in all_sentence_labels)
+            if not label_dictionary.span_labels:
+                for label in labels:
+                    # check if there are any span labels
+                    if type(label.data_point) == Span and len(label.data_point) > 1:
+                        label_dictionary.span_labels = True
+                        break
```
alanakbik, line 1443: is this to speed up the code or another reason? if so, could you add a quick description of non-unit test changes to the issue description?

flair/datasets/sequence_labeling.py
```diff
@@ -469,7 +469,7 @@ def __init__(
         path_to_column_file = Path(path_to_column_file)
         assert path_to_column_file.exists()
         self.path_to_column_file = path_to_column_file
-        self.column_delimiter = column_delimiter
+        self.column_delimiter = re.compile(column_delimiter)
```
alanakbik, line 472: Cool!

alanakbik (approved):
Ready to merge, but a few more details on the changes would be nice in the issue text.

flair/data.py
helpmefindaname: This is for speeding up the code. 
here the change is basically to use `Counter.update()` instead of treating the counter as a dictionary and count manually.

PR #3132, opened 2023-03-06
make transformer training vocab optional

helpmefindaname:


alanakbik:
@helpmefindaname thanks for fixing this! A final fix could be done after we have integrated the pluggable trainer.

PR #3138, opened 2023-03-10
Update required version of transformers library

tsproisl:
`flair/embeddings/transformer.py` has the import statement `from transformers.utils import PaddingStrategy`. That is only available from version 4.18.0 of the transformers library. Up until version 4.17.0 it is called `transformers.file_utils.PaddingStrategy`.


alanakbik:
@tsproisl thanks for fixing this!

PR #3142, opened 2023-03-13
Use to_plain_string instead of to_original_text for the property text of a Sentence

mauryaland:
When splitting a huge document into several flair Sentence objects and thus keeping trace of the token indexes from the document, the method `Sentence.to_original_text()`, used to defined the `Sentence.text` property, will store string sentences with a lot of useless whitespaces corresponding to the `start_position` of the first token of each Sentence. A simple fix is to replace `.to_original_text()` by `.to_plain_string()` for the `text` property.

For example for the following text which is split in two Sentence objects in my application:
`text = "Amaury et Valentin mangent au 77 boulevard  Perreire.\n\n\n\n Paul  a modifié le tokenizer."`

It gives before the fix:
```python
[Sentence[9]: "Amaury et Valentin mangent au 77 boulevard  Perreire.",
 Sentence[6]: "                                                          Paul  a modifié le tokenizer."]
```

And after the fix:
```python
[Sentence[9]: "Amaury et Valentin mangent au 77 boulevard  Perreire.",
 Sentence[6]: "Paul  a modifié le tokenizer."]
```

### DIFF
--- a/flair/data.py
+++ b/flair/data.py
@@ -865,7 +865,7 @@ class Sentence(DataPoint):
 
     @property
     def text(self):
-        return self.to_original_text()
+        return self.to_plain_string()
 
     def to_tokenized_string(self) -> str:
         if self.tokenized is None:


alanakbik:
Hello @mauryaland, I tested the following code and it seems to work as expected: 

```python
from flair.splitter import SegtokSentenceSplitter

# example text with many sentences
text = "Amaury et Valentin mangent au 77 boulevard  Perreire.\n\n\n\n Paul  a modifié le tokenizer."

# initialize sentence splitter
splitter = SegtokSentenceSplitter()

# use splitter to split text into list of sentences
sentences = splitter.split(text)

# predict tags for sentences
tagger = SequenceTagger.load('ner')
tagger.predict(sentences)

# iterate through sentences and print predicted labels
for sentence in sentences:
    print(sentence)
    print(sentence.start_position)
```

This prints: 

```console
Sentence[9]: "Amaury et Valentin mangent au 77 boulevard  Perreire." → ["Amaury et Valentin"/MISC, "Perreire"/LOC]
0
Sentence[6]: "Paul  a modifié le tokenizer." → ["Paul"/PER]
58
```

Could you elaborate how you get these printouts? 

Also, one potential problem of replacing .to_original_text() with .to_plain_string() would be snippets like 

```python
sentence = Sentence(" ... and then?")

print(sentence)
```

where a missing initial whitespace in the printout might confuse users.

mauryaland:
@alanakbik my bad, I did not use the flair.splitter module ! Thanks for the code snippet, I close the PR.

mauryaland:
Hello @alanakbik, I have find the difference with my homemade splitter. 

For our application, we put the whole text in a `Sentence` object and then split into multiple `Sentence` objects using the tokens. It is very useful to keep inside each token the index from the original text. So the printouts are coming from the fact that the method `to_original_text()` is doing this `self[0].start_position * " "`.

The `flair.splitter` module is keeping trace of the global index only in the `Sentence` object which makes it a bit tedious when dealing with a lot of spans and tokens for whose we need a fast access to the index across different class and methods.

Do you think we could think about a solution or the current implementation is enough? 

Thanks in advance for your return.

mauryaland:
Hello @alanakbik,

Unfortunately, it is really difficult to incorporate my homemade sentence splitter into the flair logic. Indeed, we used the tokenizer and thus the flair.Token objects to find the good split (in French we have several cases with words directly followed by a dot that is not the end of a sentence, for example "M. Dupont"). Moreover, my documents have several line breaks in a row that makes sentence splitter from spacy inefficient.

Maybe a solution could be to adapt the method Sentence.to_original_text() as following:

```python
    def to_original_text(self) -> str:
        # if sentence has no tokens, return empty string
        if len(self) == 0:
            return ""
        # otherwise, return concatenation of tokens with the correct offsets    
        elif isinstance(self.text, str) or all([isinstance(sent, str) for sent in self.text]):
            return self[0].start_position * " " + "".join([t.text + t.whitespace_after * " " for t in self.tokens]).strip()
        # no whitespace at the beginning of the sentence
        return "".join([t.text + t.whitespace_after * " " for t in self.tokens]).strip()
```

PR #3146, opened 2023-03-15
Integrate BigBio NER data sets into HunFlair

mariosaenger:
This PR implements an adapter to integrate biomedical named entity recognition data sets provided by the BigScience biomedical initiative (also known as BigBio):

https://github.com/bigscience-workshop/biomedical

BigBIO is an open library of biomedical dataloaders built using Huggingface's datasets library. It provides programmatic and harmonized access to over 120 biomedical datasets. This PR implements an adapter to the named entity recognition data sets of the library enabling users to easily work with these corpora in HunFlair (e.g. for model training or evaluation).


alanakbik:
@mariosaenger thanks for adding this! Does this mean that the "old" HUNER dataset classes (like `HUNER_CELL_LINE`) can be removed?

mariosaenger:
Hi @alanakbik! No, the "old" data sets are sill needed and used. This is due to the fact that in HUNER, in addition to the more technical harmonisation (e.g. common data format), we also standardize data sets on a content-related / semantic level (e.g. standardise different entity type labels).

alanakbik:
Ah I see - but I guess the non-standardized corpora like JNLPBA can be replaced with the BIGBIO version, or are they also still needed?

mariosaenger:
Good point. We have to check these corpora.

mariosaenger:
Hej @alanakbik! We discussed the deletion of data sets in our developer group. We would be rather reluctant to do this as it would break existing implementations referencing to these data sets. Furthermore, the BigBio datasets are (unfortunately) sometimes still of mixed quality and our data set implementations are more assured, in contrast.

However, if you insist on deleting these data sets: could we implement this in an separate PR and mark the data sets as deprecated first, since deleting the data sets would result in massive code changes?

alanakbik:
No worries, we can keep the old classes in this case.

marctorsoc:
@alanakbik after merging this, or some other PR, could we release a new version?

a) README says we're in 0.12.2, but the last one I can see is 0.12.1 :)
b) I'm eager to get hugging-face-hub unpinned (https://github.com/flairNLP/flair/pull/3149) to unpin a bunch of deps in my repos

If that's not possible, when do you expect to release a new version? is there a calendar?

marctorsoc:
amazing! thanks a ton

alanakbik:
Hello @mariosaenger, everything mostly looks good. The handling of different sentence splitters however is suboptimal with the new classes: In the old Huner classes, we appended the sentence splitter name to the generated files. This made it possible to switch sentence splitters. 

Here is an illustration: 

```python
corpus = HUNER_GENE_CELL_FINDER()
print(corpus)

corpus = HUNER_GENE_CELL_FINDER(sentence_splitter=SegtokSentenceSplitter())
print(corpus)
```

This prints two different corpus sizes, as the first corpus is loaded using the default SciSpaCy sentence splitter, and the second with a different splitter.

However, when doing this with the new classes:

```python
corpus = HUNER_GENE_TMVAR_V3()
print(corpus)

corpus = HUNER_GENE_TMVAR_V3(sentence_splitter=SegtokSentenceSplitter())
print(corpus)
```

The same corpus gets loaded twice. For the second corpus, the segtok sentence splitter is not applied. 

Can you fix this? Easiest would probably be to use the same solution as for the old classes.

mariosaenger:
@alanakbik I will have a look at it

mariosaenger:
@alanakbik Fixed this issue. Now the new data sets work as expected :wink: 

```
2023-04-12 16:28:29,421 Reading data from /home/mario/.flair/datasets/huner_gene_tmvar_v3/SciSpacySentenceSplitter_core_sci_sm_0.2.5_SciSpacyTokenizer_core_sci_sm_0.2.5
2023-04-12 16:28:29,422 Train: /home/mario/.flair/datasets/huner_gene_tmvar_v3/SciSpacySentenceSplitter_core_sci_sm_0.2.5_SciSpacyTokenizer_core_sci_sm_0.2.5/train.conll
2023-04-12 16:28:29,422 Dev: None
2023-04-12 16:28:29,422 Test: None
Corpus: 4454 train + 495 dev + 550 test sentences
2023-04-12 16:28:31,072 Reading data from /home/mario/.flair/datasets/huner_gene_tmvar_v3/SegtokSentenceSplitter
2023-04-12 16:28:31,072 Train: /home/mario/.flair/datasets/huner_gene_tmvar_v3/SegtokSentenceSplitter/train.conll
2023-04-12 16:28:31,072 Dev: None
2023-04-12 16:28:31,072 Test: None
Corpus: 4364 train + 485 dev + 539 test sentences
```

alanakbik:
@mariosaenger thanks for adding this!

PR #3149, opened 2023-03-15
unpin huggingface-hub

marctorsoc:
Simply moving from == to >= for `huggingface-hub`, since this is creating problems for me to install newer versions of `transformers` that require a minimum `huggingface-hub` version.

Some warnings appear for newer versions, so had to add some config to ignore warnings for `LayoutLMv3FeatureExtractor`, similarly as what was prev done for v2


marctorsoc:
@alanakbik very similar to my prev [PR](https://github.com/flairNLP/flair/pull/3060). Please LMK WYT

marctorsoc:
Hi @alanakbik, I'd like to have a release of flair without this problem, because otherwise (long dependency cycle) I cannot install ipykernel in my env, which is very annoying. Is it possible to release a flair version with this PR applied (but maybe not up to current master branch)?

marctorsoc:
Hi @alanakbik @helpmefindaname , can we get a release for flair? the last one was many months ago and I removed some pins that are really useful to me...

alanakbik:
Yes, we will release soon (the semester is over in 2 weeks so we can turn our attention again to preparing the next release). Sorry for the long wait! There will be more frequent updates from now on.

PR #3155, opened 2023-03-20
Fix transformer smaller training vocab

helpmefindaname:
this PR fixes the usage of the transformer smaller training vocab and improves documentation:

* fix that transformer smaller training vocab are actually used for training and not just temporarily reduce the vocab size before the training starts.
* enforce using a newer version of the library. There are 2 changes:
  * embedding-parameters that are not trainable, but present in the optimizer, will now be rightfully kept not trainable. Hence, running `embeddings.model.embeddings.word_embeddings.requires_grad_(False)` before training will work with `reduce=True`.
  * The config rightfully sets the vocab size and therefore reduced models can be saved and later loaded as such.
* the transformer smaller training vocab is now documented in the tutorials
* the onnx tutorial is now slightly improved to fit the newer structure

### DIFF
--- a/flair/trainers/trainer.py
+++ b/flair/trainers/trainer.py
@@ -384,513 +384,515 @@ class ModelTrainer:
 
             train_data = self.corpus.train
 
-        # if training also uses dev/train data, include in training set
-        if train_with_dev or train_with_test:
-            parts = [self.corpus.train]
-            if train_with_dev and self.corpus.dev:
-                parts.append(self.corpus.dev)
-            if train_with_test and self.corpus.test:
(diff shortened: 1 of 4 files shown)


alanakbik:
Thanks for this @helpmefindaname! I tested on our cluster - it works with 'distilbert-base-uncased', but running a FLERT training script with 'xlm-roberta-large' throws a weird CUDA error (`RuntimeError: CUDA error: CUBLAS_STATUS_NOT_INITIALIZED when calling `cublasCreate(handle)``). Can you check?

alanakbik:
@helpmefindaname thanks, it works now!

PR #3157, opened 2023-03-22
GH: 3156 caching bugs

alanakbik:
Draft PR that removes custom equality on DataPoint and increases cache size for context computation. More testing needed if the first is possible without breaking other parts of the code.

### DIFF
--- a/flair/data.py
+++ b/flair/data.py
@@ -420,14 +420,6 @@ class DataPoint:
     def __lt__(self, other):
         return self.start_position < other.start_position
 
-    def __eq__(self, other):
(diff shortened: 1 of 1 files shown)


alanakbik:
Update: to fix bug #3157, the following things were changed:
- the custom equality definition from `DataPoint` was removed. Two data points are now only the same if they are the same object (previously two data points were the same if they had the same text)
- the internal handling of how  `_PartOfSentence` objects (like `Span`) are created is changed to avoid duplicate object creation
- the caching was removed altogether. I tried different caching sizes and there seems to be no impact on overall speeds.

PR #3159, opened 2023-03-22
Flair with Fabric - DDP + Accelerators

aniketmaurya:
Fabric is a better and improved version of LightningLite. With this PR we will be able to train the Flair language model using DDP, easily switch hardware, and enable mixed precision training.

--- 
## Example:

You can configure all the [accelerator](https://lightning.ai/docs/fabric/stable/fundamentals/accelerators.html), [precision](https://lightning.ai/docs/fabric/stable/fundamentals/precision.html), and [distributed training strategy](https://lightning.ai/docs/fabric/stable/fundamentals/launch.html) configs in the `LanguageModelTrainer` class.

```python
"""This example is taken from https://github.com/flairNLP/flair/blob/master/resources/docs/TUTORIAL_9_TRAINING_LM_EMBEDDINGS.md"""

from flair.models import LanguageModel
from flair.trainers.language_model_trainer import LanguageModelTrainer, TextCorpus

# are you training a forward or backward LM?
is_forward_lm = True

# load the default character dictionary
dictionary: Dictionary = Dictionary.load("chars")

# get your corpus, process forward and at the character level
# data-> https://github.com/flairNLP/flair/files/9899426/penn_lm.zip
corpus = TextCorpus("resources/data/penn_lm", dictionary, is_forward_lm, character_level=True)

if __name__== "__main__":
    # instantiate your language model, set hidden size and number of layers
    language_model = LanguageModel(dictionary, is_forward_lm, hidden_size=128, nlayers=1)

    # train your language model
    trainer = LanguageModelTrainer(model=language_model, corpus=corpus, accelerator="auto", devices="auto")

    trainer.train(
        "resources/taggers/language_model",
        sequence_length=10,
        mini_batch_size=64,
        max_epochs=10,
        num_workers=2,
    )
```


Borda (approved):

aniketmaurya:
hi @dchaplinsky, Fabric already supports torch AMP - you can enable this by setting the `precision` value in the [LanguageModelTrainer](https://github.com/flairNLP/flair/pull/3159/files#diff-3182f6723676a693f90cbf4051dd11492b865c7d43283f80b1a0fb36a76bb259R169)

aniketmaurya:
> hi @dchaplinsky, Fabric already supports torch AMP - you can enable this by setting the `precision` value in the [LanguageModelTrainer](https://github.com/flairNLP/flair/pull/3159/files#diff-3182f6723676a693f90cbf4051dd11492b865c7d43283f80b1a0fb36a76bb259R169)

I would also like to add here that Fabric is designed to scale raw PyTorch code with minimal code changes. So, it handles boilerplate logic like DDP and mixed precision autocasting.

alanakbik:
Thanks @aniketmaurya this does look much cleaner!

I ran the above code on a machine with 3 RTX 3090 cards once with `devices="auto"` and once with `devices=[0]`. It was actually much faster with only one GPU in this case, and the final perplexity was roughly equal. Is this because the model is so small? Or should I decrease the batch size when training with more GPUs? 

Also, the model training at some point stopped indefinitely without an error message when training with 3 GPUs. Not sure if this is a problem with my setup. Any ideas what could be the cause of this?

aniketmaurya:
hi @alanakbik, thank you so much for testing the PR!

>  the model training at some point stopped indefinitely without an error message when training with 3 GPUs. Not sure if this is a problem with my setup. Any ideas what could be the cause of this?

-> few operations like file writing might be creating a race condition so I have guarded those to run on rank zero only ([here](https://github.com/flairNLP/flair/pull/3159/files#diff-3182f6723676a693f90cbf4051dd11492b865c7d43283f80b1a0fb36a76bb259R370)).


I am checking on the performance issue, may I know what is the training time with and without multi GPUs?

heukirne:
Excelent job, @aniketmaurya ! It will be a great improvement for Flair LLMs! 🙏 
Will be also possible to run inference in multiple GPUs like [Lightning-Transformers](https://github.com/Lightning-Universe/lightning-transformers/blob/master/docs/source/features/large_model.rst)?

alanakbik:
> hi @alanakbik, thank you so much for testing the PR!
> 
> > the model training at some point stopped indefinitely without an error message when training with 3 GPUs. Not sure if this is a problem with my setup. Any ideas what could be the cause of this?
> 
> -> few operations like file writing might be creating a race condition so I have guarded those to run on rank zero only ([here](https://github.com/flairNLP/flair/pull/3159/files#diff-3182f6723676a693f90cbf4051dd11492b865c7d43283f80b1a0fb36a76bb259R370)).
> 

Hello @aniketmaurya sorry I only now got around to testing it. Unfortunately, I am still getting the same error on 3 GPUs. It runs for 2 epochs, then runs indefinitely without any progress or GPU utilization without error message. 

Is the above script working for you in a multi-GPU setup? 

I'm attaching the script and the training data I use to test.
[script_and_data.zip](https://github.com/flairNLP/flair/files/11183152/script_and_data.zip)

Borda:
@aniketmaurya how is it going here? 🐿️

PR #3175, opened 2023-03-30
Add stacklevel to log_line in order to display correct file and line number (backwards compatible)

plonerma:
Same as GH-3170, but checks if the python version is >= 3.8 first.


helpmefindaname (approved):

alanakbik:
@plonerma thanks for improving this!

PR #3176, opened 2023-03-30
unpin gdown dependency

helpmefindaname:
I don't thing there is a reason why gdown is pinned to a specific version, however the version 4.4.0 imports `pkg_resources` which leads to a deprecation warning.


alanakbik:
Makes sense, thanks for improving this!

PR #3178, opened 2023-03-31
fix seralization of config in transformers

helpmefindaname:
I noticed that transformer embeddings download the config from the hub whenever they are loaded.
This PR fixes that.

Example to run on master:
```

tagger = SequenceTagger.load("ner-large")
tagger.save("ner-large.pt")

# now delete the huggingface cache and disable internet, before running the code below

reloaded_tagger = SequenceTagger.load("ner-large.pt")
# this will fail on master but not on this branch
```

### DIFF
--- a/flair/embeddings/transformer.py
+++ b/flair/embeddings/transformer.py
@@ -1171,11 +1171,24 @@ class TransformerEmbeddings(TransformerBaseEmbeddings):
     @classmethod
     def from_params(cls, params):
         params["use_context"] = params.pop("context_length", 0)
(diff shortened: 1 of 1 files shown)


alanakbik:
@helpmefindaname thanks for spotting and fixing this!

alanakbik:
@helpmefindaname I believe this introduces a new error. If I run: 

```python
from flair.models import MultitaskModel

linker = MultitaskModel.load('zelda')
```

I now get the following error: 
```console
TypeError: __init__() got an unexpected keyword argument 'truncate'
```

PR #3180, opened 2023-04-03
Add biomedical entity normalization

mariosaenger:
This PR implements a named entity recognition model focussing on the biomedical domain.  

The main contribution is a entity linking model which uses dense (transformer-based) embeddings and (optionally) sparse character-based representations, for normalizing an entity mention to specific identifiers in a knowledge base / dictionary. To this end, the model embeds the entity mention text and all concept names from the knowledge base and outputs the k best-matching concepts based on embedding similarity.


flair/models/biomedical_entity_linking.py
```diff
@@ -702,7 +777,9 @@ def _load_emebddings(self, model_name_or_path: str, dictionary_name_or_path: str
 
         cache_folder = flair.cache_root / "datasets"
 
-        embeddings_cache_file = cache_folder / f"{file_name}.pk"
+        pp_name = self.preprocessor.name if self.preprocessor is not None else "null"
```
mariosaenger, line 788: Typo in function name: _load_emeddings

flair/models/biomedical_entity_linking.py
```diff
@@ -946,9 +960,16 @@ class BiomedicalEntityLinker:
...
+    def __init__(
+        self,
+        candidate_generator: AbstractCandidateGenerator,
+        preprocessor: AbstractEntityPreprocessor,
+        entity_type: str,
+    ):
         self.preprocessor = preprocessor
         self.candidate_generator = candidate_generator
+        self.entity_type = entity_type
+        self.annotation_layer = ENTITY_TYPE_TO_ANNOTATION_LAYER[self.entity_type]
```
mariosaenger, line 966: Good idea. However, it should be also possible to specify the (or even better a list of) annotation_layer directly by the user (because e.g. the user has trained his / her own NER model with it's own typing). 

Moreover I recognized that the annotation layer is different when the user loads an single entity type model (e.g. Classifier.load("hunflair-disease")).

flair/models/biomedical_entity_linking.py
```diff
@@ -0,0 +1,1187 @@
...
+    ) -> Tuple[List[int], List[Span], List[str]]:
+        """Unpack all mentions in sentences for batch search."""
+
+        source = []
+        data_points = []
+        mentions = []
+        mention_annotationa_layers = []
+
+        # get all valid annotation layers: pre-determined and user input
+        annotation_layers = (
```
sg-wbi, line 976: Hey @mariosaenger I have addressed all your comments. 

Honestly I am still a bit confused by  the `annotation_layer` business. Can you take a look at this line and tell me if it's ok?

flair/models/biomedical_entity_linking.py
sg-wbi: See comment below

flair/data.py
```diff
@@ -433,61 +433,71 @@ def __len__(self):
...
+
+        def is_sorted(lst, key=lambda x: x, comparison=lambda x, y: x > y):
+            for i, el in enumerate(lst[1:]):
+                if comparison(key(el), key(lst[i])):
+                    return False
+            return True
+
+        # candidates must be sorted, regardless if higher is better or not
+        assert is_sorted(candidates, key=lambda x: x.score) or is_sorted(
+            candidates, key=lambda x: x.score, comparison=lambda x, y: x < y
```
mariosaenger, line 483: In theory two concepts can get the same score. Therefore we should lessen the constraint to less-/greater-equal here.

flair/models/biomedical_entity_linking.py
```diff
@@ -21,17 +21,29 @@
...
 from flair.embeddings import TransformerDocumentEmbeddings
 from flair.file_utils import cached_path
 
+FAISS_VERSION = "1.7.4"
+
+try:
+    import faiss
+except ImportError as error:
+    raise ImportError(
+        f"You need to install to run the biomedical entity linking: `pip faiss faiss-cpu=={FAISS_VERSION}`"
```
mariosaenger, line 44: Typo: Install command should be "pip install faiss-cpu.." (instead of "pip faiss faiss-cpu").
 
Moreover, I would recommend to adjust the warning and refer to the GPU version of faiss too, i.e. add "pip install faiss-gpu..."

flair/models/biomedical_entity_linking.py
mariosaenger: Mhh I would rather opt to apply the default annotation layers only iff the user doesn't provide a value for the annotation layers parameter. Otherwise the user has no full control over the behaviour (i.e. he/she can't get rid of the default annotation layers).

flair/models/biomedical_entity_linking.py
```diff
@@ -963,79 +935,77 @@ def __init__(
...
-        """
-        Unpack all mentions in sentences for batch search.
-        Output is list of (sentence index, mention text).
-        """
+        """Unpack all mentions in sentences for batch search."""
 
         source = []
         data_points = []
         mentions = []
+        mention_annotationa_layers = []
```
mariosaenger, line 976: Typo

flair/models/biomedical_entity_linking.py
```diff
@@ -482,6 +482,29 @@ def search(self, entity_mentions: List[str], top_k: int) -> List[List[Tuple[str,
         :result: list of tuples in the form: (entity / concept name, concept ids, similarity score).
         """
 
+    def build_candidate(self, candidate: Tuple[str, str, float]) -> EntityLinkingCandidate:
+        """Get nice container with all info about entity linking candidate"""
+
+        concept_name = candidate[0]
+        concept_id = candidate[1]
+        score = candidate[2]
+        database_name = self.dictionary.database_name
```
mariosaenger, line 491: Access "self.dictionary" is kind of a hidden assumption about the attributes of the implementation. Maybe provide the dictionary via an parameter instead?

flair/models/biomedical_entity_linking.py
sg-wbi: I have removed the option to place the index on the GPU.
Large dictionaries require a lot of GPU RAM and unless we offer some compression it does not make too much sense.
We can leave it to as a next feature in a separate PR.

PR #3181, opened 2023-04-03
github: bug report description should be textarea

stefan-it:
Hi,

for the current description of a bug report, only one line of text is currently possible:

image

This is very short, so this PR changes the input text to a text area.


alanakbik:
Thanks for fixing this!

Issue #3070, opened 2023-01-27
 [Question]:  Confidence scores for sequence tagger when crf layer is used


MarouaBoudabous:
### Question

Hi,
In the sequence tagger implementation, the confidence scores are associated to the max of scores matrix . Isn't it more accurate to use the decoding process over the scores matrix to get the confidence score for each token?

Actually, I noticed in the results that the tags associated to the confidence scores are different from the ones in the all_tags matrix. Is it the expected way to compute the confidence scores ?

Any help is appreciated.

alanakbik:
@MarouaBoudabous thanks for raising this question - @whoisjones can you take a look?

whoisjones:
hi @MarouaBoudabous,

regarding your first question: since naming is sometimes different across functions, I try to be as precise as I can. If use use a CRF in flair, the `forward` method of the `SequenceTagger` will use the CRF in both, the `forward_loss` called during training, and `predict` during evaluation. So `scores` (in `forward_loss`) and `features` (in `predict`) are equal since we do not know at that time whether we are actually dealing with scores or features. If you use a CRF in `predict`, we use Viterbi Decoding for inference to assign the scores using the decoding process. In line 209 of the `viterbi.py`:

```
        scores = softmax(scores_upto_t, dim=2)
        confidences = torch.max(scores, dim=2)
```
should do exactly what you are asking for, right?

Regarding your second question, can you please provide a minimal example of how this occurs?

MarouaBoudabous:
hi @whoisjones ,

Thanks for your response.

Exactly I am interested in the lines you mentioned above. I wonder why you associate the highest confidence score with the predicted tags. I thought that CRF finds the best tag sequence, which does not assume that the selected tag at each step is the most confident (actually, it depends also on the transitions score).

The mismatch can also be seen in the outputs of the `viterbi_decoder.decode` function. Hereafter an example of the results I obtained:

````
tags = [('o', 0.9924472570419312),

 ('o', 0.9755145907402039),

 ('o', 0.9918153882026672),

 ('o', 0.9741249680519104),

 ('o', 0.5291169285774231),

 ('o', 0.8400788307189941),

 ('o', 0.9923754334449768),

 ('o', 0.9985438585281372),

 ('o', 0.8875763416290283),

 ('o', 0.5794795751571655),

 ('o', 0.6552749276161194),

 ('o', 0.8611258268356323),...]
````

```
[...]

PR #949, opened 2019-08-02
Fix inconsistency between best_path, best score and scores

mauryaland:
Fixing inconsistency between the couple (best_path, best_scores) and scores in the SequenceTagger model using the CRF for predictions. 

Indeed, what is happening sometimes is that the tag with the highest score is not the same between *scores* and the pair (*best_path*, *best_scores*). This PR should fix this behavior.

PR #3189, opened 2023-04-06
GH-3070: Fix inconsistency between best path and scores in ViterbiDecoder

mauryaland:
Related to https://github.com/flairNLP/flair/issues/3070. It is the same fix used back then https://github.com/flairNLP/flair/pull/949 that has been removed with the refacto. It is definitely not a good interpretation of probabilities from the Viterbi decoding but at least avoid confusion. 

Some improvements could be achieved, see for example this solution from Manning https://github.com/stanfordnlp/stanza/issues/744#issuecomment-886864522 and this CRF implementation https://github.com/RasaHQ/rasa/blob/main/rasa/utils/tensorflow/crf.py. 

I would be happy to discuss more about what can be done.


mauryaland:
Hi @alanakbik @whoisjones, any thought on this topic? I think it should be discuss regarding the consequences of the CRF scoring implementation.

mauryaland:
@alanakbik @whoisjones @helpmefindaname it could be great to discuss this topic before the release of flair 0.13. Thanks in advance!

helpmefindaname:
Hi @mauryaland ,
I agree this should be part of the realse, I am currently fixing the merge conflicts etc. and then talk to @alanakbik about it.

alanakbik:
Thanks @mauryaland for improving this! 

Code to test: 

```python

# example sentence
sentence = Sentence("I work for the Humboldt Universität of Berlin")

# print NER tags
tagger = SequenceTagger.load("ner-fast")
tagger.predict(sentence, return_probabilities_for_all_classes=True)
print(sentence)

# one token has a higher probability for another class pre-viterbi
difficult_token = sentence[6]

# print the problem token "of"
print(difficult_token)

# print token prediction probabilities
sorted_predictions = sorted(difficult_token.get_tags_proba_dist("ner"), key=lambda x: x.score, reverse=True)
print(sorted_predictions[:2])
```

Before the PR, the token "of" will have the following individual probabilities: 

```console
['Token[6]: "of"'/'O' (0.8465999960899353), 'Token[6]: "of"'/'I-ORG' (0.08910000324249268)]
```
After the PR, the token "of" will have the following individual probabilities: 

```console
['Token[6]: "of"'/'I-ORG' (0.8465999960899353), 'Token[6]: "of"'/'O' (0.08910000324249268)]
```

Since it is labeled as I-ORG after Viterbi, the score post-PR is more appropriate.

PR #3195, opened 2023-04-16
get_spans_from_bio: Start new span for previous S- if class also changed

Lingepumpe:
In previous flair versions (e.g. 0.10) get_span_from_bio would start a new span if the previous tag was a "S-" token of a different class than the current token, see:
https://github.com/flairNLP/flair/blob/v0.10/flair/data.py#L698

 It happens that our production model semi-frequently produces these kinds of (invalid BIOES) prediction, and that the new span extraction performs worse on our data.
This adds back this special check for previous tag "S-", making span calculation more similar to what it was in 0.10, and remaining the same for all 100% valid BIOES tagging.

Also included are some minor code tweaks to the function to make it prettier.

### DIFF
--- a/flair/data.py
+++ b/flair/data.py
@@ -1769,7 +1769,7 @@ def get_spans_from_bio(bioes_tags: List[str], bioes_scores=None) -> List[typing.
     # return complex list
     found_spans = []
     # internal variables
-    current_tag_weights: Dict[str, float] = defaultdict(lambda: 0.0)
+    current_tag_weights: Dict[str, float] = {}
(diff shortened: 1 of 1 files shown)


alanakbik:
Hello @Lingepumpe thanks for fixing this. Reviewing the code I think the line

Lingepumpe:
Removed the unnecessary if, also wrote the whole "starts_new_span = True" conditions a bit more compactly

alanakbik:
Thanks @Lingepumpe, looks great!

PR #3201, opened 2023-04-19
Add datasets STS-B and SST-2 to flair

plonerma:
I added the datasets STS-B and SST-2. Now, all datasets for the GLUE benchmark are available.


alanakbik:

PR #3202, opened 2023-04-19
Add TextPair regression model

plonerma:
The STS-B task (added in #3201) of the glue benchmark requires a regression model which operates on sentence pairs. The added model can be used to solve this task, e.g. using following training script (this should result in a spearman correlation of ~0.8730):

```python
import sys
from pathlib import Path

import flair
from flair.datasets import GLUE_STSB
from flair.models import TextPairRegressor


def train(device, seed=0):
    flair.device = device
    flair.set_seed(seed)

    corpus = GLUE_STSB()

    label_type: str = "similarity"

    embeddings = TransformerDocumentEmbeddings("bert-base-uncased", layers="-1", fine_tune=True)

    model = TextPairRegressor(embeddings=embeddings, label_type=label_type)

    trainer.fine_tune(
        Path("resources/regression/stsb_model"),
        learning_rate=5e-5,
        mini_batch_size=32,
        max_epochs=5,
        warmup_fraction=0.06,
        main_evaluation_metric=("correlation", "pearson"),
        weight_decay=0.1,
    )


if __name__ == "__main__":
    assert len(sys.argv) == 2, "Pass a device to train on."
    train(sys.argv[1])

```

### DIFF
--- a/flair/models/__init__.py
+++ b/flair/models/__init__.py
@@ -4,6 +4,7 @@ from .language_model import LanguageModel
 from .lemmatizer_model import Lemmatizer
 from .multitask_model import MultitaskModel
 from .pairwise_classification_model import TextPairClassifier
+from .pairwise_regression_model import TextPairRegressor
 from .regexp_tagger import RegexpTagger
 from .relation_classifier_model import RelationClassifier
 from .relation_extractor_model import RelationExtractor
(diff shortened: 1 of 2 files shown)


alanakbik:
Merged with #3211

PR #3203, opened 2023-04-19
TokenClassifier model

alanakbik:
This PR introduces the `TokenClassifier` class, a renamed and extended version of `WordTagger`. It directly inherits from  `DefaultClassifier` and should be used for all token-level prediction tasks that do not require an LSTM-CRF decoder (for such tasks, the `SequenceTagger` should be used).

The main idea is to offer a model that inherits from `DefaultClassifier` for each label type we predict, i.e.:
- `TokenClassifier` for predicting `Token` labels
- `TextPairClassifier` for predicting `TextPair` labels
- `RelationClassifier` for predicting `Relation` labels
- `SpanClassifier` for predicting `Span` labels (this class is currently called `EntityLinker` and should be renamed)
- `TextClassifier` for predicting `Sentence` labels (might need to be renamed to SentenceClassifier)

An advantage of such a structure is that most functionality (such as new decoders) needs to only be implemented once in `DefaultClassifier` and then is immediately usable for all model classes.

Edit: This class also changes the default behavior of the make_label_dictionary method. The UNK token is no longer automatically added to a dictionary. We now skip unknown labels to handle loss computation in such cases.

### DIFF
--- a/flair/models/word_tagger_model.py
+++ b/flair/models/word_tagger_model.py
@@ -11,6 +11,15 @@ from flair.embeddings import TokenEmbeddings
 log = logging.getLogger("flair")
 
 
+def WordTagger(embeddings, tag_dictionary, tag_type, **classifierargs):
+    from warnings import warn
+
+    warn("The WordTagger class is deprecated after Flair version 0.12.2. Use TokenClassifier instead!")
+    return TokenClassifier(
+        embeddings=embeddings, label_dictionary=tag_dictionary, label_type=tag_type, **classifierargs
+    )
+
+
 class TokenClassifier(flair.nn.DefaultClassifier[Sentence, Token]):
     """This is a simple class of models that tags individual words in text."""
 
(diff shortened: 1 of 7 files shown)


flair/models/word_tagger_model.py
```diff
@@ -5,48 +5,82 @@
...
-from flair.data import Dictionary, Sentence, Token
+from flair.data import Dictionary, Sentence, Span, Token
 from flair.embeddings import TokenEmbeddings
 
 log = logging.getLogger("flair")
 
 
-class WordTagger(flair.nn.DefaultClassifier[Sentence, Token]):
+class TokenClassifier(flair.nn.DefaultClassifier[Sentence, Token]):
     """This is a simple class of models that tags individual words in text."""
```
helpmefindaname, line 15: I think we shouldn't just remove the WordTagger as this breaks models using the WordTagger (Especially if they are part of a MultiTaskModel).

I would suggest, that we use a `DeprecationHelper` like specified [here](https://stackoverflow.com/a/9008509), but specify the version when we delete (let's say 0.14.0; incrementing by 2.)

That way we could give users a chance to upgrade between multiple versions.

We can also discuss this further in private

alanakbik: Thanks, good idea!

PR #3208, opened 2023-04-20
#3095: add dual encoder

### LINKED ISSUE #3095: [Feature]: Add Dual Encoder
### Problem statement

This paper describes a dual encoder approach for few-shot domain adaption ([link](https://arxiv.org/abs/2203.08985)).

### Solution

Implement and test approach.

### Additional Context

_No response_

whoisjones:
closes #3095.

- Adds Dual Encoder class
- two minor typing reformatting in decoder.py at Prototypical Decoder
- added no_header attribute to WordEmbeddings. Changes nothing but the possibility to load glove vectors as well that are not in word2vec format
- made create_internal_label_dictionary static from TokenClassifer so that it can be used outside of class (required for Dual Encoder on token level)
- removed self.training in get_data_points_from_sentences. otherwise spans will be wrongly created in eval mode. + possibility for BIO encoding
- removed TODO in model.py since flake8 complained


flair/models/word_tagger_model.py
```diff
@@ -108,18 +109,22 @@ def _get_embedding_for_data_point(self, prediction_data_point: Token) -> torch.T
 
     def _get_data_points_from_sentence(self, sentence: Sentence) -> List[Token]:
         # special handling during training if this is a span prediction problem
-        if self.training and self.span_prediction_problem:
+        if self.span_prediction_problem:  # do we need self.training here?
```
alanakbik, line 112: The conversion is only necessary during training: we take Span labels and encode them as Token-level labels. During prediction, this is not necessary.

flair/nn/decoder.py
```diff
@@ -121,3 +123,112 @@ def forward(self, embedded):
...
+
+        Raises:
+            RuntimeError: If an unknown decoding type is specified.
+        """
+        if self.training or not self.label_encoder._everything_embedded(self.verbalized_labels):
+            self.label_encoder.embed(self.verbalized_labels)
+
+        label_tensor = torch.stack([label.get_embedding() for label in self.verbalized_labels])
+
+        if self.training or not self.label_encoder._everything_embedded(self.verbalized_labels):
```
alanakbik, line 224: The second condition is not needed: During, training always store embeddings. Otherwise, do not.

alanakbik (changes requested):
Thanks for adding this! Changes requested since there are some important type declarations missing, and one unnecessary if-statement. 

Additionally: have you tested "cosine-similarity"? Is that working?

flair/nn/decoder.py
```diff
@@ -121,3 +123,49 @@ def forward(self, embedded):
         scores = -distance
 
         return scores
+
+
+class LabelVerbalizerDecoder(torch.nn.Module):
+    def __init__(self, label_encoder, label_dictionary: Dictionary, decoding: str = "dot-product"):
```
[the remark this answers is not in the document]
whoisjones: DocumentEmbeddings raise ImportError, looks like circular dependency.
Embeddings as used in nn/model.py instead of DocumentEmbeddings works. If we do not want to take Embeddings, I can open up new issue to inspect this.

```ImportError while loading conftest '/Users/jgolde/PycharmProjects/flair/tests/conftest.py'.
tests/conftest.py:6: in <module>
    import flair
flair/__init__.py:28: in <module>
    from . import (  # noqa: E402 import after setting device
flair/models/__init__.py:1: in <module>
    from .clustering import ClusteringModel
flair/models/clustering.py:14: in <module>
    from flair.embeddings import DocumentEmbeddings
flair/embeddings/__init__.py:13: in <module>
    from .document import (
flair/embeddings/document.py:21: in <module>
    from flair.nn import LockedDropout, WordDropout
flair/nn/__init__.py:1: in <module>
    from .decoder import LabelVerbalizerDecoder, PrototypicalDecoder
flair/nn/decoder.py:15: in <module>
    from flair.embeddings import DocumentEmbeddings
E   ImportError: cannot import name 'DocumentEmbeddings' from 'flair.embeddings' (/Users/jgolde/PycharmProjects/flair/flair/embeddings/__init__.py)```

flair/nn/decoder.py
whoisjones: how should the decoder know about the embeddings_storage_mode? Trainer stores embeddings depending on the mode, thus we would need to include this param into the forward loss called in trainer.train().

whoisjones:
I have removed the cosine logic, it worked but just if we adjust some functions of the default model. I am currently experimenting with it and will open a new branch with it.

PR #3210, opened 2023-04-21
Update mypy

Lingepumpe:
/!\ PR contains autogenerated code, please review the commits individually, the automatic adjustments are confined to the last commit /!\

Currently we have a old version of mypy pinned in requirements-dev. This is because after mypy 0.982, the "implicit Optional" was removed, so you can no longer do "x: int = None", you must do "x: Optional[int] = None". This produces lots of errors to fix, but there is a tool that mypy recommends to automatically fix your codebase, which I applied in this PR:
https://github.com/hauntsaninja/no_implicit_optional
- mypy==0.982
+ mypy>=1.2.0


alanakbik:
@Lingepumpe thanks for improving this!

Issue #3207, opened 2023-04-20
[Bug]: TarsModels do redownload embeddings


helpmefindaname:
### Describe the bug

related but not the same as: https://github.com/flairNLP/flair/issues/3167

Locally saving a TarsModel won't save the huggingface config and therefore requires internet connection & a longer loading time when loading it on a new machine without hf-cache.
This happens due to tars pickeling the internal SequenceTagger/TextClassifier instead of just seralizing their embeddings.

### To Reproduce

```python
from flair.models import TARSClassifier

model = TARSClassifier.load("tars-base")
model.tars_embeddings.model.config._name_or_path = "bert-base-uncased"
model.tars_embeddings.base_model_name = "bert-base-uncased"
model.tars_embeddings.name = "transformer-bert-base-uncased"
model.save("local-tars-base.pt")

# clear huggingface cache or copy `local-tars-base.pt` to another machine or docker container.

model.load("local-tars-base.pt")
```

### Expected behavior

The model should load without the need of internet and it shouldn't require me to wait until the embedding config and weights are downloaded.

### Logs and Stack traces

```stacktrace
Downloading (…)okenizer_config.json: 100%|██████████| 28.0/28.0 [00:00<00:00, 17.2kB/s]
Downloading (…)lve/main/config.json: 100%|██████████| 570/570 [00:00<00:00, 517kB/s]
Downloading (…)solve/main/vocab.txt: 100%|██████████| 232k/232k [00:00<00:00, 2.12MB/s]
Downloading (…)/main/tokenizer.json: 100%|██████████| 466k/466k [00:00<00:00, 3.61MB/s]
Downloading pytorch_model.bin: 100%|██████████| 440M/440M [00:35<00:00, 12.5MB/s]
```

##### Flair
0.12.2 (master branch)
##### Pytorch
2.0.0+cu117
##### Transformers
4.28.1
#### GPU
True

Issue #3167, opened 2023-03-28
[Bug]: Training a Model Results in an OSError Related to Model Loading


ynusinovich:
### Describe the bug

When I am creating a few shot learning model by finetuning tars-base, the model crashes after training without saving to my local drive like it's supposed to.

### To Reproduce

```python
# 1. what label do you want to predict?
label_type = 'label'

# 2. make a label dictionary
label_dict = corpus.make_label_dictionary(label_type=label_type)

# 3. start from our existing TARS base model for English
tars = TARSClassifier.load("tars-base")

# 4. switch to a new task (TARS can do multiple tasks so you must define one)
tars.add_and_switch_to_new_task(task_name="classification",
                                label_dictionary=label_dict,
                                label_type=label_type,
                                )

# 5. initialize the text classifier trainer
trainer = ModelTrainer(tars, corpus)

# 6. start the training
trainer.train(base_path='../example_data/models/few_shot_model_flair',  # path to store the model artifacts
              learning_rate=0.02,  # use very small learning rate
              mini_batch_size=1,
              max_epochs=20,  # terminate after 20 epochs
              patience=1
              )
```

### Expected behaivor

I would expect the model to save to the folder.

### Logs and Stack traces

```stacktrace
HTTPError                                 Traceback (most recent call last)
File ~/Documents/env/lib/python3.9/site-packages/huggingface_hub/utils/_errors.py:213, in hf_raise_for_status(response, endpoint_name)
    212 try:
--> 213     response.raise_for_status()
    214 except HTTPError as e:

File ~/Documents/env/lib/python3.9/site-packages/requests/models.py:1021, in Response.raise_for_status(self)
   1020 if http_error_msg:
-> 1021     raise HTTPError(http_error_msg, response=self)

HTTPError: 401 Client Error: Unauthorized for url: https://huggingface.co/None/resolve/main/tokenizer_config.json

The above exception was the direct cause of the following exception:

RepositoryNotFoundError                   Traceback (most recent call last)
File ~/Documents/env/lib/python3.9/site-packages/transformers/utils/hub.py:409, in cached_file(path_or_repo_id, filename, cache_dir, force_download, resume_download, proxies, use_auth_token, revision, local_files_only, subfolder, user_agent, _raise_exceptions_for_missing_entries, _raise_exceptions_for_connection_errors, _commit_hash)
    407 try:
    408     # Load from URL or cache if already cached
--> 409     resolved_file = hf_hub_download(
[...]

PR #3212, opened 2023-04-21
Fix tars loading

helpmefindaname:
closes https://github.com/flairNLP/flair/issues/3207 and https://github.com/flairNLP/flair/issues/3167


alanakbik:

PR #3214, opened 2023-04-22
Use poetry instead of pipenv for developer/testing

Lingepumpe:
This PR removes the (defunct) pipenv setup, which was not maintained and did not pass the tests. Instead of using requirements.txt and requirements-dev.txt, it switches to use poetry https://python-poetry.org 

Installation of the flair library continues to not require poetry, it can still be installed directly via pip as before, so for flair users this change should be transparent. For the development/testing dependencies to be installed, poetry is required.

~~In this PR it was decided to *not* include the poetry.lock file, allowing the CI/CD to find the most recent legal package versions each time. This took 3 min 18s just now on the github runner. This slows down the pipeline and the first poetry install of a new developer. Alternatively we could commit the poetry.lock file, but would have to remember to do "poetry update" from time to time manually, to allow the CI/CD pipeline and the developer setups to update as new versions are released.~~

Update 1: ~~After some thinking on it, I have added poetry.lock to be commited. The main downside is that there is some work involved for developers/contributors to occasionally run "poetry update", but the advantage is that the pipeline and new developer setup is faster, and that the pipeline will be more predictable (not failing because a new version introduces a issue) and that it also solves the inconsistency/version-drift issue between pre-commit hooks package versions and versions installed in the main environment:
The exact same versions can be pinned in poetry.lock and .pre-commit-config.yaml - this could (should!) then also be enforced by CI.~~

Update 2: Since it was decided to remove pre-commit entirely the case for using the poetry.lock in CI/CD is weaker again, so I have added "poetry lock" for each pipeline run back again. Note that I did leave the poetry.lock checked in, allowing new developers to start quicker.


.github/workflows/ci.yml
```diff
@@ -13,19 +13,63 @@ jobs:
...
-      - name: Install Flair dependencies
-        run: pip install -e .
-      - name: Install unittest dependencies
-        run: pip install -r requirements-dev.txt
-      - name: Show installed dependencies
-        run: pip freeze
+      #----------------------------------------------
+      #       Try to load cached poetry installation
+      #----------------------------------------------
+      - name: Load cached Poetry installation
```
helpmefindaname, line 23: This caching mechanism looks rather complex,

`actions/setup-python` provides that functionality out of the box, see [their caching section](https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#caching-packages)

CONTRIBUTING.md
```diff
@@ -29,41 +27,54 @@ the code should hopefully be easy.
 
 ### Setup
 
-You can either use [Pipenv](https://pipenv.readthedocs.io/) for this:
+We use [poetry](https://python-poetry.org) for dependency and virtual environment management. Additionally
+[pre-commit](https://pre-commit.com/) is recommended to do code formatting and checking on your code when you commit.
+Install poetry and pre-commit on your system if you have not already. Using the [pipx](https://pypa.github.io/pipx/)
```
helpmefindaname, line 32: What is the benefit from pipx?
Why not use the [official installation guide](https://python-poetry.org/docs/#installing-with-the-official-installer) from poetry?

(asking as someone who hasn't used pipx before)

Lingepumpe: pipx allows easy and consistent upgrading/listing/uninstalling of installed packages. It does not pollute your python environment (like pip install would). And it allows installing pre-commit and poetry the same way :)

.github/workflows/ci.yml
Lingepumpe: It follows https://github.com/marketplace/actions/install-poetry-action - it looks like setup-python has something that is similar to the "cached venv" part, but does not seem to cover the "cached poetry install" part.
If we wish to prioritize a simpler ci.yaml we can use this, but it will be a bit slower, as it will always install poetry.

CONTRIBUTING.md
Lingepumpe: Reformulated the README to not put as much emphasis on pipx, also removed pre-commit as requested. Pipeline runs "poetry lock" again every time, allowing quicker noticing of issues with updated versions.

Lingepumpe:
Rebased on current master, also removed pre-commit now.

helpmefindaname (approved):
LGTM, thank you for implementing this!

alanakbik:
@Lingepumpe thanks for adding this!

PR #3220, opened 2023-04-25
Fix best epoch score update

lephong:
We need to update `best_epoch_score` correctly after each epoch.


alanakbik:
@lephong thanks for spotting and fixing this!

PR #3222, opened 2023-04-26
Better tars inference

helpmefindaname:
This PR implements batch inference for both `TARSClassifier` and `TARSTagger` and for `TARSClassifier` it allows a new parameter `force_label` which - when active - returns at least 1 label on multi-label predictions.

Current status:
I managed to speed up the inference drastically, predicting 1 sentence at a time on 6 labels now takes roughly half the time (computing 6 tars-sentences in a batch), however I seem to have a regression, yielding worse results, which I need to look into more.


alanakbik:

Issue #2801, opened 2022-06-03
RuntimeError: The expanded size of the tensor (514) must match the existing size (512) at non-singleton dimension 1.  Target sizes: [8, 514].  Tensor sizes: [1, 512]


elazzouzi1080:
Hello, 
Can someone help me to solve this problem I am trying to train a ner model with BERT and BiLSTM+CRF 
![image](https://user-images.githubusercontent.com/81225479/171873123-718cfa66-28f0-484c-a9a6-232a9bca282e.png)

kaliaanup:
I am having similar issues with Flair.

stefan-it:
Hi @elazzouzi1080 and @kaliaanup ,

could you provide an example code snippet for the fine-tuning? I guess RoBERTa is used as model :thinking:

kaliaanup:
Hey @stefan-it I was using a custom pretrain model based on google/rembert. For some odd reasons in self.tokenizer.model_max_length was coming larger than LARGE_INTEGER. This resulted in making self.truncate = False. Hence, the issue. So I kind of hacked it to fix self.tokenizer.model_max_length=256 [or any number less than LARGE_INTEGER. This solved the issue.

stefan-it:
Hi @kaliaanup,

thanks for the explanation! I'm currently trying to reproduce it and:

```
In [1]: from transformers import AutoTokenizer

In [2]: tokenizer = AutoTokenizer.from_pretrained("google/rembert")
Downloading: 100%|██████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 263/263 [00:00<00:00, 136kB/s]
Downloading: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 4.48M/4.48M [00:00<00:00, 41.7MB/s]

Downloading: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 8.31M/8.31M [00:01<00:00, 6.80MB/s]

Downloading: 100%|█████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████████| 156/156 [00:00<00:00, 95.0kB/s]

In [3]: 

In [3]: 

In [3]: tokenizer.model_max_length
Out[3]: 256
```

correctly outputs 256, and the current Flair code also uses the `model_max_length`:

[...]

PR #3223, opened 2023-04-27
Rename EntityLinker to SpanClassifier | test poetry

alanakbik:
This PR renames the `EntityLinker` class to `SpanClassifier`, to make it consistent with the other classes that directly inherit from `DefaultClassifier`.

This renaming also serves to test the new poetry setup.

It also: 
- adapts the deprecation notice in WordTagger to name the version when the class will be removed
- deletes all transformer legacy embeddings (deprecated since Flair 0.4.5) since otherwise the `pytest` command was throwing errors in poetry
- updates the poetry.lock because otherwise the tests wouldn't run (@Lingepumpe @helpmefindaname is that correct?)
- Removes **kwards from AutoModel.load call in transformers.py to address the issue noted in https://github.com/flairNLP/flair/pull/3178#issuecomment-1525598816 (@helpmefindaname does this make sense?)


alanakbik:
@Lingepumpe I am not sure how to interpret the error. It simply states "`/home/runner/work/_temp/c986e87b-e923-4c46-9455-88d50c8dd8f8.sh: line 1: .venv/bin/activate: No such file or directory`". I guess this has something to do with the poetry change?

Lingepumpe:
> @Lingepumpe I am not sure how to interpret the error. It simply states "`/home/runner/work/_temp/c986e87b-e923-4c46-9455-88d50c8dd8f8.sh: line 1: .venv/bin/activate: No such file or directory`". I guess this has something to do with the poetry change?

Yes it does, I have also received this error in the ruff PR. This is due to the installation of poetry being cached, so the `poetry config virtualenvs.in-project true` is missing when we run poetry lock, making it create the virtualenv in `/home/runner/.cache/pypoetry/virtualenvs`, instead of `.venv`, which then makes the activation not work.

helpmefindaname:
> * Removes **kwards from AutoModel.load call in transformers.py to address the issue noted in https://github.com/flairNLP/flair/pull/3178#issuecomment-1525598816

Removing the `**kwargs` would solve the issue described there, but limit the usage of transformers, e.g. https://github.com/flairNLP/flair/issues/2801 wouldn't work anymore, so I wouldn't solve the problem that way.

I assume the parameters that create that error:
`truncate, stride, embedding_length, use_lang_emb`
I would prefer to add to this method: https://github.com/flairNLP/flair/blob/master/flair/embeddings/transformer.py#L1180 that we pop those parameters, e.g.:
```
params.pop("truncate", None)
params.pop("stride", None)
params.pop("embedding_length", None)
params.pop("use_lang_emb", None)
```
That way, we should be able to still load models that were created before that patch but won't limit the ability to change the transformer config.

alanakbik:
> Yes it does, I have also received this error in the ruff PR. This is due to the installation of poetry being cached, so the `poetry config virtualenvs.in-project true` is missing when we run poetry lock, making it create the virtualenv in `/home/runner/.cache/pypoetry/virtualenvs`, instead of `.venv`, which then makes the activation not work.
> 
> I believe the fix is commit [1b90fd1](https://github.com/flairNLP/flair/commit/1b90fd1baa1da834b3e59d0d9592c44155ed4625) - ~but the pipeline is still running so I have not verified my theory yet :D~. You can cherry-pick it if the ruff pipeline passes, or merge the ruff PR and rebase.

Thanks, I think I would merge the ruff PR first and then do a rebase!

PR #3225, opened 2023-04-28
Rename EntityLinker to SpanClassifier

alanakbik:
This PR renames the `EntityLinker` class to `SpanClassifier`, to make it consistent with the other classes that directly inherit from `DefaultClassifier`.

This renaming also serves to test the new poetry setup.

It also: 
- adapts the deprecation notice in WordTagger to name the version when the class will be removed
- deletes all transformer legacy embeddings (deprecated since Flair 0.4.5) since otherwise the `pytest` command was throwing errors in poetry
- updates the poetry.lock because otherwise the tests wouldn't run (@Lingepumpe @helpmefindaname is that correct?)


alanakbik:
@helpmefindaname I removed the not-so-great fix for the issue discussed in https://github.com/flairNLP/flair/pull/3223#issuecomment-1526514432 - perhaps you could do a PR for the better fix that allows kwargs to be passed?

alanakbik:
Tests are failing, but this seems to be an issue in transformers library: https://github.com/huggingface/transformers/issues/22986

Lingepumpe:
> Tests are failing, but this seems to be an issue in transformers library: [huggingface/transformers#22986](https://github.com/huggingface/transformers/issues/22986)

Fix via cherry-pick a6512dc433615b4df491af1b1016377549391076 - in the ci.yml I previously had a "python -c 'import flair'" to work around the issue with the transformers invalid sequences (the import is run without "-W error", and fills the import cache, so when pytest runs with -W error, it doesn't break because it uses the warm import cache and doesn't re-evaluate the escape sequences.
Since you cleaned up the imports in the legacy,py, this fix was no longer working (we were not importing those offending files from transformers via 'import flair'), the fix is to directly import the offending file before running pytest - until we update the tranformer library dependency to a version that is fixed.

Issue #3216, opened 2023-04-24
[Bug]: Cannot save SequenceTagger ner-english-fast


marcglobality:
### Describe the bug

Loading a model from HF hub, then saving, gives an error

### To Reproduce

```python
import flair
import torch
import transformers
from pathlib import Path
from flair.models import SequenceTagger
import os
from os.path import expanduser
import shutil

try:
    shutil.rmtree(expanduser("~/.flair"))
    print("Removed!")
except FileNotFoundError:
    print("Not found")

[lib.__version__ for lib in [flair, torch, transformers]]
### ['0.12.2', '1.9.0+cu102', '4.26.0']

tagger = SequenceTagger.load("flair/ner-english-fast")

path = Path("test-flair")
path.mkdir(exist_ok=True)
tagger.save(path / "flair-ner-model.pt")
```

### Expected behavior

Saving the model without error

### Logs and Stack traces

```stacktrace
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
Cell In[38], line 3
      1 path = Path("test-flair")
      2 path.mkdir(exist_ok=True)
----> 3 tagger.save(path / "flair-ner-model.pt")

File ~/miniconda3/envs/wae/lib/python3.9/site-packages/flair/nn/model.py:104, in Model.save(self, model_file, checkpoint)
     99 def save(self, model_file: Union[str, Path], checkpoint: bool = False):
    100     """
    101     Saves the current model to the provided file.
    102     :param model_file: the model file
    103     """
--> 104     model_state = self._get_state_dict()
    106     # in Flair <0.9.1, optimizer and scheduler used to train model are not saved
    107     optimizer = scheduler = None

File ~/miniconda3/envs/wae/lib/python3.9/site-packages/flair/models/sequence_tagger_model.py:591, in SequenceTagger._get_state_dict(self)
    587 def _get_state_dict(self):
    588     """Returns the state dictionary for this model."""
    589     model_state = {
    590         **super()._get_state_dict(),
--> 591         "embeddings": self.embeddings.save_embeddings(use_state_dict=False),
    592         "hidden_size": self.hidden_size,
    593         "tag_dictionary": self.label_dictionary,
    594         "tag_format": self.tag_format,
    595         "tag_type": self.tag_type,
    596         "use_crf": self.use_crf,
    597         "use_rnn": self.use_rnn,
    598         "rnn_layers": self.rnn_layers,
    599         "use_dropout": self.use_dropout,
[...]

PR #3228, opened 2023-04-28
Fix loading of old models

helpmefindaname:
closes https://github.com/flairNLP/flair/issues/3216 

moves the "add default values for older embeddings" logic to the __setstate__ method and ensures, that the old WordEmbeddings will store the torch weights correctly on resave.


alanakbik:

marcglobality:
amazing! thanks. When will this be released?

PR #3229, opened 2023-04-28
Fix loading of (not so) old models

helpmefindaname:
This MR fixes the loading of models like Zelda,
in general, this affects all models with transformer-embeddings that were created before https://github.com/flairNLP/flair/pull/3178 was merged, but after https://github.com/flairNLP/flair/pull/3011


alanakbik:

Issue #3218, opened 2023-04-24
[Question]: RuntimeError: The expanded size of the tensor (6) must match the existing size (0) at non-singleton dimension 0.  


AylaRT:
### Question

I'm trying to train a Sequence Classifier with transformer embeddings (bert-base-cased) for an IOB-labelling task. The process starts just fine, but after epoch 1 - iter 324/548, I get the following error message:
"""
RuntimeError: The expanded size of the tensor (6) must match the existing size (0) at non-singleton dimension 0.  Target sizes: [6, 768].  Tensor sizes: [0, 768]
"""

Which is preceded by:
"""
File "projectfolder/.venv/lib/python3.10/site-packages/torch/jit/_trace.py", line 1220, in wrapper
    return fn(*args, **kwargs)
  File "projectfolder/.venv/lib/python3.10/site-packages/flair/embeddings/transformer.py", line 117, in fill_masked_elements
    all_token_embeddings[i, : lengths[i], :] = insert_missing_embeddings(  # type: ignore
"""

Looking at similar issues online, they are mostly related to overly long sequences, but this doesn't seem to be the case here, as the size of the tensor is 6 and 0. 
I don't immediately see any irregularities in the dataset, so I was wondering whether anyone has encountered similar issues, or if anyone knows how to go about solving the issue.

helpmefindaname:
Hi @AylaRT 
I suppose your dataset contains a magic sentence that the embeddings cannot handle.
Could you try to figure out what sentence lead to the error (and verify if it is always the same or not).

you can use the following script to do so:

```
corpus = ...
embeddings = ...

for sentence in corpus.get_all_sentences():
    print(sentence)
    embeddings.embed(sentence)
    sentence.clear_embeddings()
```
The last sentence printed will be the sentence leading to the error.

AylaRT:
Thank you so much! There are indeed a few special characters in the corpus (like Unicode character 0xF8F9) which apparently cause issues. I will remove them from the corpus.

PR #3230, opened 2023-04-28
fix bug with sentences that do not contain a single valid transformer token

helpmefindaname:
closes https://github.com/flairNLP/flair/issues/3218


alanakbik:

PR #3240, opened 2023-05-16
Making gradient clipping optional & max gradient norm variable

plonerma:
This PR allows users to disable gradient clipping (by setting `max_grad_norm=None`) and makes the maximum norm a gradient is clipped to variable.


alanakbik:

Issue #3243, opened 2023-05-20
[Feature]: Support `pickle` and `copy.deepcopy` for Sentences


dobbersc:
### Problem statement

Serializing and deserializing sentences with `pickle` is a common practice when using Flair in multiprocessing systems, e.g. with the [Ray](https://docs.ray.io/en/latest/ray-core/walkthrough.html) library. Similarly, creating deep copies of objects is a usually natural feature. Unfortunately, loading a pickled sentence with `pickle` or creating a deep copy with `copy.deepcopy` fails for sentences with span or relation annotations.

Example:

```python
import copy
import pickle

from flair.data import Sentence

sentence: Sentence = Sentence("Berlin is the capital of Germany.")

# Works: pickle / copy on unannotated sentence
pickled: Sentence = pickle.loads(pickle.dumps(sentence))
copied: Sentence = copy.deepcopy(sentence)
assert pickled.to_original_text() == copied.to_original_text() == "Berlin is the capital of Germany."

# Add span annotations
sentence[:1].add_label("ner", value="LOC")  # Berlin
sentence[5:6].add_label("ner", value="LOC")  # Germany

# Does not work: pickle / copy on span-annotated sentence
# TypeError: __new__() missing 1 required positional argument: 'tokens'
pickled: Sentence = pickle.loads(pickle.dumps(sentence))
copied: Sentence = copy.deepcopy(sentence)
assert (
    {span.text for span in pickled.get_spans("ner")}
    == {span.text for span in copied.get_spans("ner")}
    == {"Berlin", "Germany"}
)

```

Related to #3191.

### Solution

It would be great if sentences with arbitrary annotations were pickable or at least serializable in any manner and copyable.

### Additional Context

The reason for the `TypeError` is sentences annotated with spans or relations contain references to these, where their implementations use the `__new__` method. 

https://github.com/flairNLP/flair/blob/5c53e35805e09ddd62e8443180d2e65dde928d38/flair/data.py#L550
https://github.com/flairNLP/flair/blob/5c53e35805e09ddd62e8443180d2e65dde928d38/flair/data.py#L609

To fix this the `pickle` (and `copy`) module require implementing a [`__getnewargs__`](https://docs.python.org/3/library/pickle.html#object.__getnewargs__) or [`__getnewargs_ex__`](https://docs.python.org/3/library/pickle.html#object.__getnewargs_ex__) method that provides the arguments for `__new__` when unpickling an object.

PR #3245, opened 2023-05-22
gh-3243: support pickle & deepcopy

helpmefindaname:
Closes gh-3243

A suggestion how we could support pickle and deepcopy without having to create multiple spans etc.

The idea is to use `__getitem__` to create spans and relations by indexing: `sentence[0:5]` to create a span, `sentence[span_a, span_b]` to create a relation.

Downside would be, that those who still instantiate `Span` or `Relation` manually, would receive unexpected results.

@dobbersc @alanakbik what do you think?

### DIFF
--- a/flair/data.py
+++ b/flair/data.py
@@ -630,25 +630,9 @@ class Token(_PartOfSentence):
 class Span(_PartOfSentence):
     """This class represents one textual span consisting of Tokens."""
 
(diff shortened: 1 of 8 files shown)


jeffpicard:
This would simplify the interface for [multi gpu support](https://github.com/flairNLP/flair/pull/3548) a lot.

PR #3247, opened 2023-05-23
Add support for AfricaPOS Dataset

stefan-it:
Hi,

this PR adds support for the recently proposed [AfricaPOS](https://www.uni-saarland.de/en/department/lst/news/nine-papers-accepted-at-acl-2023-26586.html) Dataset.

Details can be found in [this](https://twitter.com/davlanade/status/1653725938012520448) tweet.

The dataset is available in this repo: https://github.com/masakhane-io/masakhane-pos

I received preprint of the paper and wrote unit tests to check number of parsed sentences for dataset splits for each language.

Example usage of AfricaPOS:

```python
from flair.datasets import AFRICA_POS

corpus = AFRICA_POS(languages="bam")
```


stefan-it:
@alanakbik Please let me know, if dataset name is ok: it does not quite match into the `UD_` naming scheme, because dataset is not in Universal Dependencies format, as it only has token and upos.

dadelani:
@stefan-it , the dataset name is MasakhaPOS, arXiv paper will be out tomorrow

stefan-it:
Thanks @dadelani for feedback, I corrected the dataset name now :)

stefan-it:
Preprint is now available [here](https://arxiv.org/abs/2305.13989) :hugs:

stefan-it:
Hi @helpmefindaname do you accidentally have an idea, why `poetry` stage will get stuck in timeout:

https://github.com/flairNLP/flair/actions/runs/5062453919/jobs/9099744358?pr=3247

This was the already the case yesterday, I've just re-ran the build, but same error.

helpmefindaname:
> Hi @helpmefindaname do you accidentally have an idea, why `poetry` stage will get stuck in timeout:
> 
> https://github.com/flairNLP/flair/actions/runs/5062453919/jobs/9099744358?pr=3247
> 
> This was the already the case yesterday, I've just re-ran the build, but same error.

The dependency resolution took needlessly long, as it tried out all 300+ `boto3` versions with an incompatible dependency before taking the right one.
https://github.com/flairNLP/flair/pull/3249 should fix the problems.

stefan-it:
After #3258 I will do a rebase now :)

alanakbik:
@stefan-it thanks for adding this! And thanks @dadelani for creating this dataset!

PR #3249, opened 2023-05-25
fix dependency resolution error

helpmefindaname:
Trying to figure out why `poetry lock` get's stuck


helpmefindaname:
The initial selection of `urllib3==2.0.2` leads to incompatibility with `botocore` leading to poetry trying out several 100 `botocore` versions before selecting a `urllib3` version that is compatible. 

Pinning urllib3 to >=1.0.0<2.0.0 should work faster.

stefan-it:
Thanks for this @helpmefindaname !

I think we should also update GitHub-related dependencies for Actions:

```
- uses: actions/checkout@v3 (current v2)
- uses: actions/setup-python@v4 (current v2)
```

in a follow-up PR :)

Issue #3250, opened 2023-05-25
[Bug]: final-model.pt on KeyboardInterrupt even if `save_final_model=False`


plonerma:
### Describe the bug

Flair's trainer saves a "final-model.pt" on a keyboard-interrupt even if `save_final_model` is false.

I haven't opened a PR yet, since I'm not sure, if this behavior is intended (e.g. to enable the continuation of training). Personally, I would not expect flair to produce a "final-model.pt" in this case.

The following line (as well as in the following except-block), would need to be encased by an if-statement:

https://github.com/flairNLP/flair/blob/5c53e35805e09ddd62e8443180d2e65dde928d38/flair/trainers/trainer.py#L707

### To Reproduce

```python
Execute any training with `save_final_model=False` and interrupt the training using Ctrl+C.
```

### Expected behavior

I would expect flair to terminate the training (after doing the final evaluation) and to not write a final model file.

##### Flair
0.12.2
##### Pytorch
2.0.1+cu117
##### Transformers
4.29.2
#### GPU
True

alanakbik:
Good catch! Yes, that is unintended behavior I would say.

PR #3251, opened 2023-05-25
GH-3250: Save final model only if `save_final_model` is True (even if the training is interrupted)

plonerma:
`ModelTrainer` should only save `final-model.pt` if the respective flag is True (regardless of the reason the training ended (see #3250).


alanakbik:
Makes sense! Thanks for adding this @plonerma!

PR #3258, opened 2023-06-05
Remove poetry

helpmefindaname:
@alanakbik and I agreed to remove poetry again as an environment handler.

The reasons are that:
* There are several people who had issues setting up poetry, leading to a larger hurde for potential contributors.
* poetry happens to uninstall torch-cuda versions and replaces them with torch cpu versions on windows


flair/embeddings/token.py
```diff
@@ -77,8 +77,8 @@ def __init__(self, embeddings: List[TokenEmbeddings], overwrite_names: bool = Tr
         # IMPORTANT: add embeddings as torch modules
         for i, embedding in enumerate(embeddings):
             if overwrite_names:
-                embedding.name = f"{str(i)}-{embedding.name}"
-            self.add_module(f"list_embedding_{str(i)}", embedding)
+                embedding.name = f"{i!s}-{embedding.name}"
```
alanakbik, line 80: interesting, never came across this notation.

alanakbik:
Thanks @helpmefindaname!

Issue #3266, opened 2023-06-14
TypeError: state_dict() got multiple values for argument 'destination'


elazzouzi1080:
### Describe the bug

When I want to save my best model after one epoch, I have this error: TypeError: state_dict() got multiple values for argument 'destination'

2023-06-14 10:23:21,212 ----------------------------------------------------------------------------------------------------
2023-06-14 10:23:21,212 Corpus: 5483 train + 5483 dev + 5483 test sentences
2023-06-14 10:23:21,213 ----------------------------------------------------------------------------------------------------
2023-06-14 10:23:21,213 Train:  5483 sentences
2023-06-14 10:23:21,214         (train_with_dev=False, train_with_test=False)
2023-06-14 10:23:21,214 ----------------------------------------------------------------------------------------------------
2023-06-14 10:23:21,214 Training Params:
2023-06-14 10:23:21,215  - learning_rate: "0.1" 
2023-06-14 10:23:21,215  - mini_batch_size: "64"
2023-06-14 10:23:21,215  - max_epochs: "3"
2023-06-14 10:23:21,216  - shuffle: "True"
2023-06-14 10:23:21,216 ----------------------------------------------------------------------------------------------------
2023-06-14 10:23:21,216 Plugins:
2023-06-14 10:23:21,217  - AnnealOnPlateau | patience: '3', anneal_factor: '0.5', min_learning_rate: '0.0001'
2023-06-14 10:23:21,217 ----------------------------------------------------------------------------------------------------
2023-06-14 10:23:21,217 Final evaluation on model from best epoch (best-model.pt)
2023-06-14 10:23:21,217  - metric: "('micro avg', 'f1-score')"
2023-06-14 10:23:21,218 ----------------------------------------------------------------------------------------------------
2023-06-14 10:23:21,218 Computation:
2023-06-14 10:23:21,218  - compute on device: cuda:0
2023-06-14 10:23:21,219  - embedding storage: none
2023-06-14 10:23:21,219 ----------------------------------------------------------------------------------------------------
2023-06-14 10:23:21,219 Model training base path: "/donnees/home/elazzouzi/NER_BERT_BILSTM_CRF_FINAL/LSTM_CRF_FastText_mBERT"
2023-06-14 10:23:21,219 ----------------------------------------------------------------------------------------------------
2023-06-14 10:23:21,220 ----------------------------------------------------------------------------------------------------
[...]

PR #3272, opened 2023-06-19
fix multiple arguments destination

helpmefindaname:
Closes https://github.com/flairNLP/flair/issues/3266


alanakbik:
Thanks for fixing this @helpmefindaname!

Issue #3275, opened 2023-06-28
[Feature]: Enable users to disable `sample_missing_splits` in relevant subclasses


plonerma:
### Problem statement

When running experiments on the GLUE datasets, users might not expect/want the test datasets to be sampled from the training data as currently is the case.

### Solution

When sampling missing splits from training data in the `Corpus` class, a warning should be logged to notify the user of this sampling.

Additionally, it should be possible to disable this feature (which is possible in the `Corpus` class, but not all relevant derived classes).

I will open a corresponding PR soon.

plonerma:
In most GLUE classes, this does actually seem to be the case. I missed this in #3201, when adding the SST-2 corpus.

PR #3276, opened 2023-06-28
Gh3275: sample_missing_splits in SST-2

plonerma:
See #3275


alanakbik:
@plonerma this PR should make it so that GLUE datasets have the option of being loaded without sampling the test split?

I am getting an error when running this: 

```python
from flair.datasets import GLUE_SST2

# default setting: SST-2 with sampling missing splits
corpus = GLUE_SST2()
print(corpus)

# without splits
corpus = GLUE_SST2(sample_missing_splits=False)
print(corpus)
```

alanakbik:
Thanks for fixing this @plonerma!

PR #3288, opened 2023-07-28
Extend German LER Dataset

stefan-it:
Hi,

recently, proper dataset splits for the German LER Dataset were introduced:

https://github.com/elenanereiss/Legal-Entity-Recognition/commit/eda16d35f2c174845c83005e35d23bab10cd7dad

I updated the `NER_GERMAN_LEGAL` dataset, so these newly introduced splits are used.

A basic unit test was also added that checks if number of instances for all dataset splits are identical with reported numbers (see e.g. [here](https://huggingface.co/datasets/elenanereiss/german-ler#data-splits)).


alanakbik:
@stefan-it thanks for adding this! And thanks @elenamer for releasing the data!

Issue #3291, opened 2023-08-06
[Bug]: fasttext embeddings don't work


sinaahmadi:
### Describe the bug

Thanks for making it possible to use [custom embeddings](https://github.com/flairNLP/flair/blob/master/resources/docs/embeddings/FASTTEXT_EMBEDDINGS.md). Using FastText is particularly useful for less-resourced languages that are not supported in your own embeddings yet.

I have an issue working with FastText embeddings (binary files). When using FastText, it is not possible to save the model even though it gets trained without any problem. I get an error saying `FileNotFoundError: [Errno 2] No such file or directory: '/var/folders/gj/dvhshvn52d92wrk8r7cqgx9r0000gp/T/tmpglky2zrb/fasttext.model.vectors_vocab.npy`.

To make sure that there are not other issues regarding my code, I converted the FastText embeddings to Gensim and could train and save the model successfully.

I am not sure what the problem with FastText is but there seems to be a bug either with saving the embeddings or pointing to the correct directory.

### To Reproduce

```python
from flair.data import Corpus
from flair.datasets import ColumnCorpus
from flair.embeddings import WordEmbeddings, StackedEmbeddings, FlairEmbeddings
from flair.models import SequenceTagger
from flair.trainers import ModelTrainer
from flair.embeddings import FastTextEmbeddings

columns = {0: 'text', 1: 'pos'}

# this is the folder in which train, test and dev files reside
data_folder = 'datasets'

# init a corpus using column format, data folder and the names of the train, dev and test files
corpus: Corpus = ColumnCorpus(data_folder, columns,
                              train_file='train.txt',
                              test_file='test.txt',
                              dev_file='dev.txt')

len(corpus.train)
print(corpus.train[1].to_tagged_string('pos'))

label_type = 'pos'
label_dict = corpus.make_label_dictionary(label_type=label_type)
print(label_dict)

embeddings_fasttext = FastTextEmbeddings('/Users/sina/Bucket/Embeddings/cc.ckb.300.bin')

model = SequenceTagger(hidden_size=256,
                        embeddings=embeddings_fasttext,
                        tag_dictionary=label_dict,
                        tag_type=label_type)

trainer = ModelTrainer(model, corpus)

trainer.train('models',
              learning_rate=0.1,
              mini_batch_size=32,
              max_epochs=20)
```

### Expected behavior

I expect that to see the model saved after training. 

### Logs and Stack traces

```stacktrace
[...]

PR #3293, opened 2023-08-07
prevent fasttext embeddings to be stored separately

helpmefindaname:
Closes https://github.com/flairNLP/flair/issues/3291


alanakbik:
Can confirm that this works! Thanks @helpmefindaname for fixing this!

PR #3300, opened 2023-08-21
add action to remove Awaiting Response label when an response was made

helpmefindaname:
The basic Concept of the "Awaiting Response" label is:

* Whenever an issue requires more input or an additional check of the author, mark it with the "Awaiting Response"
* When looking for unhandled issues, add `-label:"Awaiting Response"`  to the filter

Whenever the Author responds again to the issue, the "Awaiting Response" label will be removed again

### DIFF
--- a/.github/workflows/issues.yml
+++ b/.github/workflows/issues.yml
@@ -1,0 +1,11 @@
+on: issue_comment
+
+jobs:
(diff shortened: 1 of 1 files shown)


alanakbik:
Thanks @helpmefindaname!

Issue #3298, opened 2023-08-17
[Bug]: When Finetuning, TypeError is thrown due to missing decoder


raykyn:
### Describe the bug

When finetuning a pretrained model by following Tutorial 9, a TypeError is thrown due to the language model not having a decoder-object.

### To Reproduce

```python
from flair.data import Dictionary
from flair.models import LanguageModel
from flair.trainers.language_model_trainer import LanguageModelTrainer, TextCorpus
from flair.embeddings import FlairEmbeddings


if __name__ == "__main__":
    
    root = "."
    
    #task_id = sys.argv[1]
    task_id = "1"

    # are you training a forward or backward LM?
    if task_id == "1":
        language_model = FlairEmbeddings('news-forward').lm
        short = "fw"
    elif task_id == "2":
        language_model = FlairEmbeddings('news-backward').lm
        short = "bw"

    print(language_model.decoder)
    
    # are you training a forward or backward LM?
    is_forward_lm = language_model.is_forward_lm

    # load the default character dictionary
    dictionary: Dictionary = language_model.dictionary

    # get your corpus, process forward and at the character level
    corpus = TextCorpus(f'{root}/corpus',
                        dictionary,
                        is_forward_lm,
                        character_level=True)

    # instantiate your language model, set hidden size and number of layers
    trainer = LanguageModelTrainer(language_model, corpus)

    # finetune your language model
    trainer.train(f'{root}/lm_finetune_{short}',
              sequence_length=250,
              mini_batch_size=100,
              learning_rate=20,
              patience=10,
              checkpoint=True)
```

### Expected behavior

To start the fine-tuning process as usual.

### Logs and Stack traces

```stacktrace
Traceback (most recent call last):
  File "finetune_lm.py", line 44, in <module>
    trainer.train(f'{root}/lm_finetune_{short}',
  File "C:\Users\iprada\anaconda3\envs\MachineLearning\lib\site-packages\flair\trainers\language_model_trainer.py", line 305, in train
    output, rnn_output, hidden = self.model.forward(data, hidden)
  File "C:\Users\iprada\anaconda3\envs\MachineLearning\lib\site-packages\flair\models\language_model.py", line 97, in forward
    decoded = self.decoder(output)
TypeError: 'NoneType' object is not callable
```

##### Flair
0.12.2
##### Pytorch
2.0.1+cu118
##### Transformers
4.31.0
#### GPU
True

raykyn:
[...]

PR #3301, opened 2023-08-21
fix embedding finetuning tutorial

helpmefindaname:
Closes https://github.com/flairNLP/flair/issues/3298


alanakbik:

PR #3319, opened 2023-09-18
fix build doc page action trigger

helpmefindaname:


marcglobality:
@helpmefindaname @alanakbik could you release a new version please? last one is from 6 months ago

PR #3321, opened 2023-09-25
orange secondary color

helpmefindaname:
this PR fixes the buttons on the new docpage being purple instead of orange


alanakbik:
Thanks @helpmefindaname!

PR #3325, opened 2023-10-02
don't pickle classes & plugins in modelcard

helpmefindaname:


alanakbik:
Looks good! Thanks for fixing this @helpmefindaname!

stefan-it:
Thanks for that!

I really like this new plugins interface :hugs: I think a nice plugin code also be for carbon emission tracking with [`codecarbon`](https://github.com/mlco2/codecarbon) as it is currently done in Transformers (see [here](https://huggingface.co/docs/hub/model-cards-co2)) :)

PR #3351, opened 2023-10-23
Add German MobIE NER Dataset

stefan-it:
Hey,

this PR adds support for the German MobIE NER Dataset.

> This is a German-language dataset that has been human-annotated with 20 coarse- and fine-grained entity types, and it includes entity linking information for geographically linkable entities. The dataset comprises 3,232 social media texts and traffic reports, totaling 91K tokens, with 20.5K annotated entities, of which 13.1K are linked to a knowledge base. In total, 20 different named entities are annotated.

### Usage

In Flair the dataset can be used like this:

```python
from flair.datasets import NER_GERMAN_MOBIE

corpus = NER_GERMAN_MOBIE()

print(corpus)
```

### Example

I've already fine-tuned models with Flair (see [this](https://huggingface.co/blog/stefan-it/autotrain-flair-mobie) blog post) and uploaded them to the Hugging Face Model Hub. Here's an example image:

image


alanakbik:
Thanks a lot for adding this @stefan-it! And cool blog post!

PR #3363, opened 2023-10-27
fix trainer final evaluation with tstv & best-model

helpmefindaname:


alanakbik:
Fixes the main issue, but this script throws errors: 

```python
# 1. get the corpus
from flair.data import Corpus
from flair.datasets import TREC_6

corpus: Corpus = TREC_6()

# 2. what label do we want to predict?
label_type = "question_class"

# 3. create the label dictionary
label_dict = corpus.make_label_dictionary(label_type=label_type)

# 4. initialize transformer document embeddings (many models are available)
document_embeddings = TransformerDocumentEmbeddings("xlm-roberta-large", fine_tune=True)

# 5. create the text classifier
classifier = TextClassifier(document_embeddings, label_dictionary=label_dict, label_type=label_type)

# 6. initialize trainer

# 7. fine-tune the model, but **reduce the vocabulary** for faster training
trainer.train(
    "resources/taggers/question-classification-with-transformer_final",
    reduce_transformer_vocab=True,  # set this to False for slow version
    max_epochs=2,
    learning_rate=0.0001,
    use_final_model_for_eval=False,
)
```

It works for `distilbert-base-uncased` and `bert-base-uncased`, but throws an error for `xlm-roberta-large`: 

```console
RuntimeError: Error(s) in loading state_dict for TextClassifier:
        size mismatch for embeddings.model.embeddings.word_embeddings.weight: copying a param with shape torch.Size([8426, 1024]) from checkpoint, the shape in current model is torch.Size([250003, 1024]).
```

@helpmefindaname any idea why `xlm-roberta-large` is causing a problem here?

PR #3364, opened 2023-10-27
fix embedding size for xlm roberta models

helpmefindaname:


stefan-it:
Hey @helpmefindaname ,

many thanks for fixing it! I can confirm that the patch is working with XLM-R :hugs:

alanakbik:
Can also confirm! Thanks @helpmefindaname!

Issue #3367, opened 2023-11-01
[Bug]: UnboundLocalError: local variable 'has_unknown_label' referenced before assignment


anna-shopova:
### Describe the bug

After completing the first training epoch and attempting to evaluate on the dev data, I encounter an error. This issue arises when I set both **add_unk=True** and **add_dev_test=True** for the **corpus.make_label_dictionary()** method.

### To Reproduce

```python
import flair
from flair.data import Corpus
from flair.datasets import ColumnCorpus
from flair.embeddings import TransformerWordEmbeddings
from flair.models import SpanClassifier
from flair.trainers import ModelTrainer


corpus: Corpus = ColumnCorpus(data_folder, columns,
                              train_file='train.txt',
                              test_file='test.txt',
                              dev_file='dev.txt',
                              document_separator_token="-DOCSTART-")

label_dictionary = corpus.make_label_dictionary('nel', add_unk=True)
# label_dictionary = corpus.make_label_dictionary('nel', add_dev_test=True)

corpus.filter_empty_sentences()
corpus.filter_long_sentences(800)

embeddings = TransformerWordEmbeddings(
    model='xlm-roberta-base',
    layers="-1",
    subtoken_pooling="first",
    fine_tune=True,
    use_context=True
)

tagger = SpanClassifier(
    embeddings=embeddings,
    label_dictionary=label_dictionary
)

trainer = ModelTrainer(tagger, corpus)

trainer.train('./results/',
              learning_rate=5.0e-5,
              optimizer=torch.optim.AdamW,
              mini_batch_size=1,
              train_with_dev=False,
              shuffle=False
              )
```

### Expected behavior

The model should evaluate without any errors after the training epoch.

### Logs and Stack traces

```stacktrace
[...]

PR #3372, opened 2023-11-06
fix has unknown label is not always initialized

helpmefindaname:
fixes https://github.com/flairNLP/flair/issues/3367

### DIFF
--- a/flair/nn/model.py
+++ b/flair/nn/model.py
@@ -828,6 +828,7 @@ class DefaultClassifier(Classifier[DT], typing.Generic[DT, DT2], ABC):
 
             overall_loss = torch.zeros(1, device=flair.device)
             label_count = 0
+            has_any_unknown_label = False
             for batch in batches:
                 # filter data points in batch
                 batch = [dp for dp in batch if self._filter_data_point(dp)]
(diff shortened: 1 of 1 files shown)


alanakbik:
@helpmefindaname thanks for fixing this! The error seems to occur because there is no valid data point to be predicted on in the test set, and therefore the `has_unknown_label` is never initialized.

When testing with the example script below, this PR does indeed fix this error. However, it throws another error later when the evaluate code tries to access the `classification_report_dict["accuracy"]` field - since there is no valid data point to make a prediction, no accuracy can be computed.

This error can be reproduced with: 

```python
import flair
from flair.data import Corpus
from flair.datasets import ColumnCorpus, WNUT_17
from flair.models import SpanClassifier

flair.set_seed(123)

corpus: Corpus = WNUT_17().downsample(0.01)

label_dictionary = corpus.make_label_dictionary("ner", add_unk=True)

    model="distilbert-base-uncased", layers="-1", subtoken_pooling="first", fine_tune=True, use_context=True
)

tagger = SpanClassifier(embeddings=embeddings, label_dictionary=label_dictionary)

trainer.fine_tune(
    "./results/",
    max_epochs=1,
)
```

alanakbik:
The second error is another issue, merging this now and opening a separate PR for the above error.

PR #3374, opened 2023-11-06
use batch count instead of total training samples for logging metrics

### LINKED ISSUE #3373: [Feature]: use iteration count instead of total_train_samples when logging metrics
### Problem statement

With the new Plugin system, we could add services like wandb or clearml to create nice plots about training, for example this one:
image

However, if we use models with variational amounts of samples (e.g. relation extraction, where the sample count is the amount of entities squared), the plots won't look that nice anymore:
image
This is due to the logging using the total_train_samples as iteration number (x-axis), hence variation is clearly displayed.

### Solution

Instead of using the `total_train_samples` as variable for logging, count the number of batches and use that one for the iteration.

### Additional Context

_No response_

helpmefindaname:
closes #3373 

this makes the logs look more smoothly:
image

### DIFF
--- a/flair/trainers/trainer.py
+++ b/flair/trainers/trainer.py
@@ -538,6 +538,7 @@ class ModelTrainer(Pluggable):
             # At any point you can hit Ctrl + C to break out of training early.
             try:
                 total_train_samples = 0
+                batch_count = 0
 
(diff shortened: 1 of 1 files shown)


alanakbik:

PR #3375, opened 2023-11-12
Fix classification report if dataset has no labels

alanakbik:
If the test dataset has no labels, and the model makes no predictions, no evaluation numbers (precision/recall/F1) can be computed. This causes errors in cases such as this: 

```python
import flair
from flair.data import Corpus
from flair.datasets import ColumnCorpus, WNUT_17
from flair.models import SpanClassifier

flair.set_seed(123)

corpus: Corpus = WNUT_17().downsample(0.01)

label_dictionary = corpus.make_label_dictionary("ner", add_unk=True)

    model="distilbert-base-uncased", layers="-1", subtoken_pooling="first", fine_tune=True, use_context=True
)

tagger = SpanClassifier(embeddings=embeddings, label_dictionary=label_dictionary)

trainer.fine_tune(
    "./results/",
    max_epochs=1,
)
```

This PR slightly refactors the `evaluate()` code in the `DefaultClassifier` such that the error handling for this case becomes more explicit. This also fixes the above error.

### DIFF
--- a/flair/nn/model.py
+++ b/flair/nn/model.py
@@ -434,6 +434,9 @@ class Classifier(Model[DT], typing.Generic[DT], ReduceTransformerVocabMixin, ABC
                 labels=labels,
             )
 
+            # compute accuracy separately as it is not always in classification_report (e.. when micro avg exists)
+            accuracy_score = round(sklearn.metrics.accuracy_score(y_true, y_pred), 4)
+
             # if there is only one label, then "micro avg" = "macro avg"
             if len(target_names) == 1:
                 classification_report_dict["micro avg"] = classification_report_dict["macro avg"]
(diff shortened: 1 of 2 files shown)


flair/nn/model.py
```diff
@@ -434,71 +434,66 @@ def evaluate(
...
-        result = Result(
-            main_score=main_score,
-            detailed_results=detailed_result,
-            classification_report=classification_report_dict,
-            scores=scores,
-        )
+        else:
+            # issue error and default all evaluation numbers to 0.
+            error_text = (
+                f"ACHTUNG! It was not possible to compute evaluation values because: "
```
helpmefindaname, line 484: I think this PR is a good opportunity to remove the German word and phrase it completely in english.
Besides that:
The current output looks like this:
```
2023-11-13 11:07:43,133 ACHTUNG! It was not possible to compute evaluation values because: - The evaluation data has no gold labels for label_type='nel'!- And no predictions were made!Double check your corpus (if the test split has labels), and how you initialize the ModelTrainer!
2023-11-13 11:07:43,133 ACHTUNG! It was not possible to compute evaluation values because: - The evaluation data has no gold labels for label_type='nel'!- And no predictions were made!Double check your corpus (if the test split has labels), and how you initialize the ModelTrainer!
```
I would
* Fix the spacing
* Not log it twice: either log the error or add it in the detailed_results, but not both

helpmefindaname (changes requested):

flair/nn/model.py
alanakbik: Fixed in latest push.

PR #3387, opened 2023-12-21
Added support for biomedical datasets with multiple entity types

WangXII:
In this pull request, we've added support for datasets annotating multiple entities like BC5CDR for chemical and disease entities.

This enables training of models which can recognize multiple entites at once without the need for specialized model for every entity type. Exemplary for this implementation is the class [HUNER_ALL_BIORED](https://github.com/flairNLP/flair/blob/280701b776ff0bb679e5022caebb97e58c12845a/flair/datasets/biomedical.py#L5551), which tags genes, chemicals, cell lines, diseases and species in a single corpora.


mariosaenger (commented):
@WangXII thank you very much for implementing all the changes and going through all datasets.

Concerning the integration of the task description, however, I would opt for another solution (and not materialzing the prompts into the conll files), since at the end the user should be able to use the following api:

```python
ner_tagger = Classifier.load("hunflair2")

sentence = Sentence("TP53 is an onco-gene.")
ner_tagger.predict(sentence)
```

This means, the end-user should not responsible / bothered with prepending the task prompts etc. I will commit  a suggestion for an implementation right away.

alanakbik:
Cool new feature! Thanks @WangXII and @mariosaenger for adding this!

PR #3388, opened 2023-12-24
Entity Mention Linker

helpmefindaname:
via @mariosaenger in https://github.com/flairNLP/flair/pull/3180

"The main contribution is a entity linking model which uses dense (transformer-based) embeddings and (optionally) sparse character-based representations, for normalizing an entity mention to specific identifiers in a knowledge base / dictionary. To this end, the model embeds the entity mention text and all concept names from the knowledge base and outputs the k best-matching concepts based on embedding similarity."

for each "gene", "disease", "chemical" & "species" I created and uploaded a model to hf, 
Those models can be loaded via `EntityMentionLinker.load("bio-{label_type}"` or `EntityMentionLinker.load("bio-{label_type}-exact-match"` The first represents the recommended default configuration (note for species it is currently exact match due to lack of alternative, while the latter represents the most simple model to use.

I suppose the recommendation of models will change soon, but would recommend to not make this part of this PR, but rather change it afterwards.


alanakbik (commented):
Thanks for this PR @helpmefindaname and @WangXII! 

It was a bit hard to put together the code to actually tag a sentence, for several reasons: 
- It is not obvious that one first needs to run a biomedical NER tagger
- The entity linker models have a cryptic `entity_tag_type` which needs to be overwritten
- The predictions are just integer codes, and it was not obvious to me to which knowledge base they refer to 
- If I interpret correctly, the tested model seems to get even simple things wrong 

Here is the code used for testing: 

```python
from flair.datasets import NCBI_GENE_HUMAN_DICTIONARY
from flair.models import EntityMentionLinker
from flair.nn import Classifier

# Example sentence
sentence = Sentence("We observed A2M, CDKN1A and alpha-2-macroglobulin in the specimen.")

# instantiate NER tagger for genes
ner_tagger = Classifier.load("hunflair-gene")

# instantate Gene linker (the entity_label_type needs to be set to "ner")
gene_linker = EntityMentionLinker.load("bio-gene")
gene_linker.entity_label_type = "ner"

# use both taggers to predict
ner_tagger.predict(sentence)
gene_linker.predict(sentence)

# interpret results using dictionary
dictionary = NCBI_GENE_HUMAN_DICTIONARY()

print(sentence)

for entity in sentence.get_labels("gene"):
    print(entity)
    link = dictionary[entity.value]
    print(f" -> linked to: '{link.concept_name}'")
```

In this snippet, though mentioned explicitly, the genes A2M and alpha-2-macroglobulin are linked to some random entries. The "exact match" model performs only marginally better.

@WangXII - am I doing something wrong or is the accuracy of the linking in such cases so low?

Some suggestions / questions: 
1. You could prepare `MultitaskModel`s for each type (gene, disease, etc.) that combine the necessary NER tagger and linker, and correctly set the entity_tag_type. This would allow users to easily instantiate a single model that directly works. 
2.  The label_type of all linking models could be "link" - not sure what is gained from having different label_type names for different biomedical NER classes.
3. Is there some way of including the dictionary into the linker and preparing convenience functions to interpret the links?

WangXII:
Thanks for creating out the updated pull request @helpmefindaname and for pointing out the low accuracy values @alanakbik! We indeed had a bug with linking to the correct knowledge base identifiers. Now this should be fixed and A2M and alpha-2-macroglobulin point to the correct NCBI Gene identifier number 2.

alanakbik:
@WangXII thanks for the update! Is there a way I can compute evaluation numbers in Flair? i.e. load a gold dataset, load the model, make predictions and evaluate? Could you post a snippet for this?

WangXII:
@sg-wbi can answer this the best. I think we have yet to update the evaluation script to the revised Flair API

alanakbik:
@sg-wbi can you share an evaluation script? I would like to use it to test the models for accuracy before merging.

sg-wbi:
> @sg-wbi can you share an evaluation script? I would like to use it to test the models for accuracy before merging.

Unfortunately this is not  straightforward. When we developed this our efforts in the integration into flair stopped at the model level. This is becasue we had  a very specific use case. At the current state our evaluation scripts require multiple preprocessing steps. 

We (@mariosaenger, @WangXII) will get back to you asap with a workable solution (to your suggestions as well).

sg-wbi:
@alanakbik 

## Tests for accuracy

Here's the script that will give you accuracy results for 3 commonly used datasets.

```python
from collections import defaultdict

from datasets import load_dataset

from flair.models import EntityMentionLinker
from flair.models.entity_mention_linking import BioSynEntityPreprocessor

ENTITY_TYPE_TO_MODEL = {
    "diseases": "dmis-lab/biosyn-sapbert-ncbi-disease",
    "chemical": "dmis-lab/biosyn-sapbert-bc5cdr-chemical",
    "genes": "dmis-lab/biosyn-sapbert-bc2gn",
}

ENTITY_TYPE_TO_DATASET = {"diseases": "ncbi_disease", "chemical": "bc5cdr", "genes": "gnormplus"}

ENITY_TYPE_TO_DICTIONARY = {"diseases": "ctd-diseases", "chemical": "ctd-chemicals", "genes": "ncbi-gene"}


def main():
    for entity_type, model in ENTITY_TYPE_TO_MODEL.items():
        ds_name = ENTITY_TYPE_TO_DATASET[entity_type]
        dictionary = ENITY_TYPE_TO_DICTIONARY[entity_type]

        ds = load_dataset(f"bigbio/{ds_name}", f"{ds_name}_bigbio_kb", trust_remote_code=True)
        print(f"Loaded corpus: `{ds_name}`")

        annotations = [a for d in ds["test"] for a in d["entities"]]
        if ds_name == "bc5cdr":
            annotations = [a for a in annotations if a["type"].lower() == entity_type]

        mention_to_uids = defaultdict(list)
        uid_to_link = {}
        for a in annotations:
            # skip mentions without normalization
            if len(a["normalized"]) == 0:
                continue

            if ds_name == "gnormplus":
                # no prefix for NCBI Gene
                uid_to_link[a["id"]] = [n["db_id"] for n in a["normalized"]]
            else:
                uid_to_link[a["id"]] = [":".join((n["db_name"], n["db_id"])) for n in a["normalized"]]

            for t in a["text"]:
                mention_to_uids[t].append(a["id"])

        linker = EntityMentionLinker.build(
            model,
            entity_type,
            dictionary_name_or_path=dictionary,
            hybrid_search=True,
            preprocessor=BioSynEntityPreprocessor(),
            batch_size=1024,
        )

        mentions = sorted(linker.preprocessor.process_mention(m) for m in set(mention_to_uids))
        results = linker.candidate_generator.search(entity_mentions=mentions, top_k=1)

        hits = 0
        total = 0
        for m, r in zip(mentions, results):
            for uid in mention_to_uids[m]:
                y_true = uid_to_link[uid]
                y_pred, _ = r[0]
                total += 1
                if y_pred in y_true:
                    hits += 1

        accuracy = round(hits / total * 100, 2)

        print(f"EVALUATION |  MODEL: `{model}`, CORPUS: `{ds_name}`,  ACCURACY@1: {accuracy}")


if __name__ == "__main__":
    main()
```

You should get the following results

```bash
EVALUATION |  MODEL: `dmis-lab/biosyn-sapbert-ncbi-disease`, CORPUS: `ncbi_disease`,  ACCURACY@1: 84.3
EVALUATION |  MODEL: `dmis-lab/biosyn-sapbert-bc5cdr-chemical`, CORPUS: `bc5cdr`,  ACCURACY@1: 93.85
EVALUATION |  MODEL: `dmis-lab/biosyn-sapbert-bc2gn`, CORPUS: `gnormplus`,  ACCURACY@1: 74.26
```

**NOTE**: the script reports accuracy on the _gold mentions_  of the dataset (i.e. no NER) since this is how these models are commonly evaluated. Testing that `EntityMentionLinker` can correctly handle the output of a `Classifier` are in "test_biomedical_entity_linking.py"

**NOTE**: links to fetch models from the huggingface hub need to be updated. Let us know where we should put pre-trained models.

## Suggestions

> Is there some way of including the dictionary into the linker and preparing convenience functions to interpret the links?

Here we still have to find a nice solution. As you and @mariosaenger discussed last week - we may opt to build a distinct `NENLabel` class which inherits from Label and extends it by introducing two additional attributes: dictionary and canonical_name.

> The label_type of all linking models could be "link" - not sure what is gained from having different label_type names for different biomedical NER classes.

For this, we adapted the implementation according to the approach done in the `RelationClassifier` model. Users are now able to specifically define on which label and entity types the linker model should be applied, e.g. by passing `"ner"` for all labels or `{"ner": ["genes", "diseases"]}` for restricting the linker to certain entity types for a given label type.

> You could prepare `MultitaskModels` for each type (gene, disease, etc.) that combine the necessary NER tagger and linker, and correctly set the entity_tag_type.

Since this PR is already quite substantial, if that's ok with you, we would provide bundled models which correctly link the new NER and NEN models together when all code changes are done.

helpmefindaname:
@sg-wbi

I am working on integrating you evaluation script into this PR.
While debugging, I think I found a bug:
```diff
        ....

-       mentions = sorted(linker.preprocessor.process_mention(m) for m in set(mention_to_uids))
-       results = linker.candidate_generator.search(entity_mentions=mentions, top_k=1)
+       mentions = sorted(mention_to_uids.keys())
+       preproc_mentions = [linker.preprocessor.process_mention(m) for m in mentions]
+       results = linker.candidate_generator.search(entity_mentions=preproc_mentions, top_k=1)

        hits = 0
        total = 0
        for m, r in zip(mentions, results):
        ....
```

Imo the before ignored all labels for mentiones that were preprocessed (e.g. the hard ones).

With the initial script I get the following results:
```
EVALUATION |  MODEL: `dmis-lab/biosyn-sapbert-ncbi-disease`, CORPUS: `ncbi_disease`,  ACCURACY@1: 83.91
EVALUATION |  MODEL: `dmis-lab/biosyn-sapbert-bc5cdr-chemical`, CORPUS: `bc5cdr`,  ACCURACY@1: 94.0
EVALUATION |  MODEL: `dmis-lab/biosyn-sapbert-bc2gn`, CORPUS: `gnormplus`,  ACCURACY@1: 69.31
```
after the change I get some worse results, with the exception of the genes corpus, that improves a lot:
```
EVALUATION |  MODEL: `dmis-lab/biosyn-sapbert-ncbi-disease`, CORPUS: `ncbi_disease`,  ACCURACY@1: 76.35
EVALUATION |  MODEL: `dmis-lab/biosyn-sapbert-bc5cdr-chemical`, CORPUS: `bc5cdr`,  ACCURACY@1: 84.62
EVALUATION |  MODEL: `dmis-lab/biosyn-sapbert-bc2gn`, CORPUS: `gnormplus`,  ACCURACY@1: 79.58
```

I also noticed, that some labels in the dataset do not exist in the dictionary and therefore cannot be predicted right.
I found the following upper bounds on the detection of the respective dictionary-dataset pairs:
```
EVALUATION |  NOT_IN_DATA: 81, MAX_ACCURACY@1: 91.56
EVALUATION |  NOT_IN_DATA: 271, MAX_ACCURACY@1: 95.06
EVALUATION |  NOT_IN_DATA: 227, MAX_ACCURACY@1: 92.93
```

Can you comfirm this?

sg-wbi:
> I am working on integrating you evaluation script into this PR.

Thanks for taking care of this.

> While debugging, I think I found a bug:
> Imo the before ignored all labels for mentiones that were preprocessed (e.g. the hard ones).

Yes you are right. In my version only mentions which do not change after preprocessing were evaluated. However, in your version only those which do change after preprocessing are evaluated. This is assuming you do not modify anything else.
So I believe the correct one would be:

```diff
....

-       mentions = sorted(linker.preprocessor.process_mention(m) for m in set(mention_to_uids))
+       mention_to_uids = {linker.preprocessor.process_mention(m): uids for m, uids in mention_to_uids.items()}
+       mentions = set(mention_to_uids)
       results = linker.candidate_generator.search(entity_mentions=mentions, top_k=1)

        hits = 0
        total = 0
        for m, r in zip(mentions, results):
....
```

Which gives:
```
EVALUATION |  MODEL: `dmis-lab/biosyn-sapbert-ncbi-disease`, CORPUS: `ncbi_disease`,  ACCURACY@1: 75.72
EVALUATION |  MODEL: `dmis-lab/biosyn-sapbert-bc5cdr-chemical`, CORPUS: `bc5cdr`,  ACCURACY@1: 79.03
EVALUATION |  MODEL: `dmis-lab/biosyn-sapbert-bc2gn`, CORPUS: `gnormplus`,  ACCURACY@1: 79.24
```

> I also noticed, that some labels in the dataset do not exist in the dictionary and therefore cannot be predicted right.
> Can you comfirm this?

Yes this is to be expected. Dictionary labels change over time (become obsolete/are merged). Some of these corpora were created 10 years ago

flair/datasets/entity_linking.py
```diff
@@ -1,20 +1,470 @@
...
+class EntityLinkingDictionary:
+    """Base class for downloading and reading of dictionaries for entity entity linking.
+
+    A dictionary represents all entities of a knowledge base and their associated ids.
+    """
+
+    def __init__(
+        self,
+        candidates: Iterable[EntityCandidate],
+        dataset_name: Optional[str] = None,
```
alanakbik, line 32: Please add comments what `candidates` and `dataset_name` are.

flair/datasets/entity_linking.py
```diff
@@ -1,20 +1,470 @@
...
+        return self._idx_to_candidates[item]
+
+    def __contains__(self, item: str) -> bool:
+        return item in self._idx_to_candidates
+
+    def to_in_memory_dictionary(self) -> "InMemoryEntityLinkingDictionary":
+        return InMemoryEntityLinkingDictionary(list(self._idx_to_candidates.values()), self._dataset_name)
+
+
+class InMemoryEntityLinkingDictionary(EntityLinkingDictionary):
```
alanakbik, line 79: Please add comment explaining this class.

flair/datasets/entity_linking.py
```diff
@@ -1760,3 +2210,398 @@ def __init__(
             banned_sentences=banned_sentences,
             sample_missing_splits=sample_missing_splits,
         )
+
+
+class BigbioCorpus(Corpus, abc.ABC):
```
alanakbik, line 2215: Please add comment to explain this class

flair/datasets/entity_linking.py
```diff
@@ -1760,3 +2210,398 @@ def __init__(
...
+        if isinstance(paths, Path):
+            paths = [paths]
+        all_sentences = []
+        for path in paths:
+            for entry in self._file_to_dicts(path):
+                all_sentences.extend(self._dict_to_sentences(entry))
+        return FlairDatapointDataset(all_sentences)
+
+
+class BIGBIO_NCBI_DISEASE(BigbioCorpus):
```
alanakbik, line 2326: Also for each of the BigbioCorpus subclasses please add a brief description of the dataset and link.

flair/datasets/entity_linking.py
```diff
@@ -1760,3 +2210,398 @@ def __init__(
...
+                            "normalized": normalized,
+                        }
+                    )
+
+                unified_example["entities"] = unified_entities
+
+                yield unified_example
+
+
+class BIGBIO_BC5CDR_CHEMICAL(BigbioCorpus):
```
alanakbik, line 2421: Also for each of the BigbioCorpus subclasses please add a brief description of the dataset and link.

flair/datasets/entity_linking.py
```diff
@@ -1760,3 +2210,398 @@ def __init__(
...
+            for passage in xdoc.passages:
+                for span in passage.annotations:
+                    ent = self._get_bioc_entity(span, db_id_key="MESH")
+                    if ent["type"].lower() == "chemical":
+                        data["entities"].append(ent)
+
+            yield data
+
+
+class BIGBIO_GNORMPLUS(BigbioCorpus):
```
alanakbik, line 2506: Also for each of the BigbioCorpus subclasses please add a brief description of the dataset and link.

flair/models/sequence_tagger_model.py
```diff
@@ -838,44 +833,7 @@ def _fetch_model(model_name) -> str:
...
-                log.error("-" * 80)
-                log.error(
-                    f"ERROR: The key '{model_name}' was neither found on the ModelHub nor is this a valid path to a file on your system!"
-                )
-                log.error(" -> Please check https://huggingface.co/models?filter=flair for all available models.")
-                log.error(" -> Alternatively, point to a model file on your local drive.")
-                log.error("-" * 80)
-                Path(flair.cache_root / "models" / model_folder).rmdir()  # remove folder again if not valid
-                raise
+            model_path = hf_download(model_name)
```
alanakbik, line 836: Good change!

flair/models/entity_mention_linking.py
```diff
@@ -0,0 +1,1174 @@
...
+        data_points: Union[List[Sentence], Dataset],
+        gold_label_type: str,
+        out_path: Optional[Union[str, Path]] = None,
+        embedding_storage_mode: str = "none",
+        mini_batch_size: int = 32,
+        main_evaluation_metric: Tuple[str, str] = ("accuracy", "f1-score"),
+        exclude_labels: List[str] = [],
+        gold_label_dictionary: Optional[Dictionary] = None,
+        return_loss: bool = True,
+        k: int = 1,
```
alanakbik, line 1141: Many of these parameters are unused (like `out_path`, `embedding_storage_mode`, etc.). This is a limitation of our current evaluate signature of the `Model` class. 

Some possibilities (not in this PR):
- Have the EntityMentionLinker inherit from Classifier instead of Model and reuse its evaluate function. This would also entail adapting the predict method and determining how many of these parameters make sense for an untrained model. For instance, is there a batch size?
- Implement functionality for as many parameters as possible. For instance, we could implement something for out_path.

alanakbik (changes requested):
Looks good! Justrequests for more docstrings in some new classes.

Additionally, the evaluate method is quite bare-bones, but I have no good idea how to better reuse existing evaluation code for more informative evaluation. So this issue should not block a merge.

sg-wbi:
@alanakbik I have added the requested docstrings/comments.

alanakbik (approved):

alanakbik:
@sg-wbi @helpmefindaname @mariosaenger thanks a lot for adding this major new feature to Flair!

Issue #3059, opened 2023-01-20
[Feature]: Make konoha optional


marctorsoc:
### Problem statement

I'm having a problem installing flair, since it **requires** https://github.com/flairNLP/flair/blob/master/requirements.txt#L20 `konoha<5.0.0,>=4.0.0`. This in turn makes this
```
  importlib-metadata<4.0.0,>=3.7.0 (from konoha==4.6.5->flair==0.11.3->my-repo (setup.py))
```

### Solution

I can see two solutions here, but lmk if there's another one (apart from pinning deps on my end, ofc):
* Given that most users (me included) won't use Japanese, make this an extra
* Fix `flair` to remove this pin and adapt to latest versions of `konoha`. I assume this is harder

alanakbik:
Hello @marctorsoc good point and I would tend towards making konoha optional. This fits well with our general rule that libraries that are only needed for 1 feature should be optional.

marctorsoc:
@alanakbik I'm happy to work on this one, if you explain how do you want to implement extras. At the moment I don't see any packages made optional

Shall I just remove `konoha` from `requirements.txt` and do something like this? https://github.com/globality-corp/globality-black/blob/master/setup.py#L45-L48

alanakbik:
Normally, we do a check whether the library is installed and a printout with instructions for objects that use specific libraries. See for instance the SpacyTokenizer: https://github.com/flairNLP/flair/blob/15776018b150a3b8500e64a3a474e3c52674170c/flair/tokenization.py#L216 

I see we actually already have such a check implemented for konoha: https://github.com/flairNLP/flair/blob/15776018b150a3b8500e64a3a474e3c52674170c/flair/tokenization.py#L153 

So the easiest option would be to just remove it from the dependencies.

I haven't worked with optional dependencies yet, but a solution like you suggest for the option "japanese" could also be interesting.

marctorsoc:
there you have it https://github.com/flairNLP/flair/pull/3060

PR #3394, opened 2024-01-13
deps: relax version constraint for konoha

himkt:
This PR updates the version constraint for konoha.

I'm so sorry that I didn't notice the issue https://github.com/flairNLP/flair/issues/3059.
I updated konoha that removed the dependency for `importlib-metadata`, let me update `requirements-dev.txt`.

Now konoha only depends on `requests` with the support for Python3.8+ (https://github.com/himkt/konoha/blob/v5.5.2/pyproject.toml#L27).
So I think it can be safely used with flair and I update the requirements file.

And I have one question for dependency of `janome`. `janome` is only used in `JapaneseTokenizer`, which depends on `konoha`, therefore `janome` should be installed in the same place of `konoha`, I think. There are multiple options:

- (1) move `janome` to `requirements-dev.txt`
- (2) remove `janome` from `requirements.txt` and install with `konoha[janome]` (extras)
- (3) move `konoha` to `requirements.txt`
- (4) remove `konoha` from `requirements-dev.txt` and install `konoha` with `konoha[janome]` (extras)

I'd like to a maintainer's opinion.

### DIFF
--- a/flair/tokenization.py
+++ b/flair/tokenization.py
@@ -159,7 +159,7 @@ class JapaneseTokenizer(Tokenizer):
             )
             log.warning('- Install konoha with "pip install konoha[{tokenizer_name}]"')
             log.warning('  - You can choose tokenizer from ["mecab", "janome", "sudachi"].')
-            log.warning("Note that we Flair support only konoha<5.0.0,>=4.0.0")
+            log.warning("Note that we Flair support only konoha<6.0.0")
             log.warning("-" * 100)
             sys.exit()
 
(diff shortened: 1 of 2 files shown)


helpmefindaname (approved):

alanakbik:
@himkt thanks for updating this!

helpmefindaname:
Hi @himkt 
For me the ideal would be to not have neither listed as dependency, as less unused dependencies are better.
However with the current implementation, `janome` is always required as it will be imported, while `konoha` is only needed if the `JapaneseTokenizer` is used. Hence we don't depend on `konoha` and handle it as an optional dependency instead. 

If you have time to refactor the `JapaneseTokenizer` to use lazy imports on `janome`, PR's are welcome

himkt:
Thank you @helpmefindaname, it makes sense! I'll work on that later. 👍

PR #3405, opened 2024-02-09
deps: make janome optional

himkt:
Make `janome` optional and install with `konoha` (for `JapaneseTokenizer`).
Based on the discussion https://github.com/flairNLP/flair/pull/3394#issuecomment-1935771503.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,6 +1,5 @@
 black[jupyter]>=23.3.0
-janome>=0.4.2
-konoha<6.0.0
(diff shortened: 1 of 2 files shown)


helpmefindaname (approved):

helpmefindaname:
I had in my head that it was more complicated but looking at the code again, this looks good.

Also I can clearly follow your example. Thank you for taking the time to solve this.

alanakbik:
@himkt thanks for improving this! And thanks @helpmefindaname for testing!

PR #3410, opened 2024-02-23
Update documentation for Hunflair2 release

mariosaenger:
This PR updates the documentation for HunFlair lifting it to new release. More specifically the PR contains:

- Documentation pages highlighting the new features and usage of HunFlair2
- Minor adaptations to smooth user experience
- Integration of warning messages if using HunFlair (version 1) models


flair/models/multitask_model.py
```diff
@@ -260,6 +260,14 @@ def _fetch_model(model_name) -> str:
 
         cache_dir = Path("models")
         if model_name in model_map:
+            if model_name.startswith("hunflair") or model_name == "bioner":
```
alanakbik, line 263: "hunflair2" also starts with "hunflair", so I think this warning would always be printed.

flair/models/sequence_tagger_model.py
```diff
@@ -781,6 +781,14 @@ def _fetch_model(model_name) -> str:
         elif model_name in hu_model_map:
             model_path = cached_path(hu_model_map[model_name], cache_dir=cache_dir)
 
+            if model_name.startswith("hunflair"):
```
alanakbik, line 784: Same here

flair/models/multitask_model.py
mariosaenger: True. However, HunFlair2 will not be loaded as MultitaskModel. I fix it anyway.

flair/models/sequence_tagger_model.py
mariosaenger: True. However, HunFlair2 will not be loaded as SequenceTaggerModel. I fix it anyway.

flair/models/entity_mention_linking.py
```diff
@@ -648,6 +649,8 @@ def p(text: str) -> str:
                         emb = emb / torch.norm(emb)
                     dense_embeddings.append(emb.cpu().numpy())
                     sent.clear_embeddings()
+
+                # empty cuda cache if device is a cuda device
                 if flair.device.type == "cuda":
```
alanakbik, line 654: @sg-wbi Is this really required?

alanakbik (approved):
All good, thanks for adding this and thanks for your patience! There are some smaller points that we will address with follow-up PRs. 

One question: is the manual deleting of the cuda cache really necessary?

helpmefindaname:
Hi @alanakbik @sg-wbi 
The GPU cache clear is done after each batch, as otherwise we would get Memory errors if the embeddings of the whole dataset is larger than the memory available on the GPU (besides the needs for the Model).
Since the search is always performed on CPU, we don't need to keep the embeddings cached on the GPU, hence there is no big advantage of not deleting it.

alanakbik:
Alright, thanks!

PR #3412, opened 2024-02-28
bugfix in test: double `data_folder`

ya0guang:


helpmefindaname (approved):
Interesting behavior of python. Apparently (`path / path`) is the same as just `path`, so no change required.
However I find using just `path` is way cleaner, hence approving this.

Thank you for your contribution.

alanakbik:
Thanks for improving this @ya0guang!

PR #3413, opened 2024-02-28
Bugfix in pairwise_regression_model.py

ya0guang:
It seems like `batch` should be used in iterations, rather than `pairs`.

I'm not 100% sure about this fix, please feel free to close it if I'm wrong.


helpmefindaname (approved):
Can comfirm. Thank you for finding this!

alanakbik:
Thanks for fixing this @ya0guang!

PR #3417, opened 2024-03-08
feat: allow cpu device even if gpu available

drbh:
This PR updates the init to check if `FLAIR_DEVICE="cpu"` and will use the cpu even on a system with cuda available. Currently its not possible to use flair on your cpu if you have a gpu which may be undesirable when managing memory


drbh:
bumping and tagging @alanakbik for visibility 🙏

PR #3423, opened 2024-03-11
ci: fix black and ruff

stefan-it:
Hi,

this PR pins both `black` and `ruff` versions and also applies tree-wide formatting.

Hopefully, CI will be green then :)

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,11 +1,11 @@
-black[jupyter]>=23.3.0
+black[jupyter]==24.2.0
 konoha[janome]<6.0.0
 mypy>=1.2.0
 pytest>=7.3.1
-pytest-black-ng>=0.4.1
+pytest-black-ng==0.4.1
 pytest-github-actions-annotate-failures>=0.1.8
 pytest-mypy>=0.10.3
-pytest-ruff>=0.0.5
+pytest-ruff==0.3.1
 types-dataclasses>=0.6.6
 types-Deprecated>=1.2.9.2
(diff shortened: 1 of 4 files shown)


stefan-it:
Green again :partying_face:

requirements-dev.txt
```diff
@@ -1,12 +1,12 @@
-black[jupyter]>=23.3.0
+black[jupyter]==24.2.0
 konoha[janome]<6.0.0
 mypy>=1.2.0
 pytest>=7.3.1
-pytest-black-ng>=0.4.1
+pytest-black-ng==0.4.1
 pytest-github-actions-annotate-failures>=0.1.8
 pytest-mypy>=0.10.3
-pytest-ruff>=0.0.5
```
helpmefindaname, line 8: Notice, that ruff itself is not pinned and pinning pytest-ruff is not enough for that.

So please add ruff as a version.
Also plead don't do hard pins but make softer ones:
`black[jupyter]==24.2.*` would allow us to accept bugfixes while requireing the dependency to behave the same.

helpmefindaname (changes requested):

requirements-dev.txt
stefan-it: Thanks for the feedback! I've just updated the requirements file :)

helpmefindaname (approved):

alanakbik:
Thanks @stefan-it!

stefan-it:
Thanks for merging @alanakbik ! I will rebase my other open PRs now :)

PR #3432, opened 2024-03-22
add prediction label type for span classifier

helpmefindaname:


alanakbik (commented):
This currently does not work. To reproduce: 

```python
from flair.models import MultitaskModel

# For comparison: This works since the label type is "ner" for both models in the multitask model
classifier: MultitaskModel = MultitaskModel.load("zelda")

sentence = Sentence("Kirk and Spock met on the Enterprise")

classifier.predict(sentence)

print(sentence)

# Giving them sensible label names, now made possible with this PR
classifier.tasks["Task_1"]._label_type = "nel"
classifier.tasks["Task_1"]._span_label_type = "ner"

# However, this no longer makes predictions
sentence = Sentence("Kirk and Spock met on the Enterprise")

classifier.predict(sentence)

print(sentence)
```

The problem lies in the `_filter_data_point` method of the `DefaultClassifier`.

alanakbik:
@helpmefindaname I've pushed a small fix for the issue. I'll approve now, but can you double beck before you merge?

helpmefindaname:
@alanakbik good catch, I did a small change so we allow both, setting the `span_label_type` immediately (but training on `label_type` if available) and having it set later instead.

alanakbik (approved):

alanakbik:
Thanks @helpmefindaname!

PR #3437, opened 2024-04-02
ci: fix ruff error on numpy deprecation

sg-wbi:
Replace deprecated `numpy.row_stack` cause ruff error.


alanakbik:
Thanks, saw this too late! I included it directly in the hunflair branch!

PR #3455, opened 2024-05-13
fix _all_scores_for_token in ViterbiDecoder

mauryaland:
There was a bug where in a batch, it took the last tag sequence of that batch from the Viterbi decoder instead of the tag sequence corresponding to its sentence.


helpmefindaname (approved):
Hi @mauryaland 
thank you for this fix, it looks good to me.
May I ask you to create a test for the specific behaviour that you fixed? That way we can ensure, that we won't run into the same issue in a year or so

alanakbik:

alanakbik:
To reproduce the problem that is fixed: 

```python
tagger: SequenceTagger = SequenceTagger.load("ner-fast")

sentence_1 = Sentence("Mr John Smith arrived")
sentence_2 = Sentence("Hey ho ho ho")

tagger.predict([sentence_1, sentence_2], force_token_predictions=True, return_probabilities_for_all_classes=True)

print(sentence_1[1])
print(sentence_1[1].get_tags_proba_dist("ner"))

print()

print(sentence_1[2])
print(sentence_1[2].get_tags_proba_dist("ner"))
```

The tag probability distribution gives a different probability of the predicted tag than the prediction. That is because it is using the last tag sequence (sentence_2) rather than the actual sentence. This is fixed in the PR.

Issue #3438, opened 2024-04-03
[Bug]: No Python 3.12 support


b3n4kh:
### Describe the bug

Trying to use Flair on python 3.12 fails, but this is neither documented in the README nor reflected in the `pyproject.toml`

The underlying issue is the dependency [transformer_smaller_training_vocab](https://github.com/helpmefindaname/transformer-smaller-training-vocab/blob/main/pyproject.toml) the dependency pinning here is so strict that [sentencepiece](https://github.com/google/sentencepiece/releases) 0.2.0 can not be installed, which is required for python 3.12 support.

### To Reproduce

```python
from flair.data import Sentence
```
```


### Expected behavior

Loading the model successfully.

### Logs and Stack traces

```stacktrace
File "/home/ben/git/taranis-ai/src/worker/worker/bots/nlp_bot.py", line 5, in <module>
    from flair.data import Sentence
  File "/home/ben/.venv/lib/python3.12/site-packages/flair/__init__.py", line 42, in <module>
    from . import (  # noqa: E402 import after setting device
  File "/home/ben/.venv/lib/python3.12/site-packages/flair/trainers/__init__.py", line 2, in <module>
    from .trainer import ModelTrainer
  File "/home/ben/.venv/lib/python3.12/site-packages/flair/trainers/trainer.py", line 21, in <module>
    from flair.trainers.plugins import (
  File "/home/ben/.venv/lib/python3.12/site-packages/flair/trainers/plugins/__init__.py", line 5, in <module>
    from .functional.reduce_transformer_vocab import ReduceTransformerVocabPlugin
  File "/home/ben/.venv/lib/python3.12/site-packages/flair/trainers/plugins/functional/reduce_transformer_vocab.py", line 5, in <module>
    from transformer_smaller_training_vocab import reduce_train_vocab
ModuleNotFoundError: No module named 'transformer_smaller_training_vocab'
```

### Environment

This fails because importing flair is not possible without the flair import:

```
#### Versions:
##### Pytorch
2.2.2+cu121
##### Transformers
4.39.3
#### GPU
False
```

Additional:

```
$ pip freeze | grep flair
flair==0.13.1
```

b3n4kh:
While debugging this, watch out for errors regarding scipy triu `ImportError: cannot import name 'triu' from 'scipy.linalg'` which is an [issue with gensim](https://github.com/piskvorky/gensim/issues/3525)

Issue #3441, opened 2024-04-04
ImportError: cannot import name 'triu' from 'scipy.linalg'[Question]: 


Sritharan-racap:
### Question

All of a sudden I'm getting this error. please help. I did `pip install flair`

```
 File "/Users/smahendra/opt/anaconda3/envs/gpt11/lib/python3.11/site-packages/flair/embeddings/document.py", line 16, in <module>
    from flair.embeddings.token import FlairEmbeddings, StackedEmbeddings, TokenEmbeddings
  File "/Users/smahendra/opt/anaconda3/envs/gpt11/lib/python3.11/site-packages/flair/embeddings/token.py", line 10, in <module>
    import gensim
  File "/Users/smahendra/opt/anaconda3/envs/gpt11/lib/python3.11/site-packages/gensim/__init__.py", line 11, in <module>
    from gensim import parsing, corpora, matutils, interfaces, models, similarities, utils  # noqa:F401
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/smahendra/opt/anaconda3/envs/gpt11/lib/python3.11/site-packages/gensim/corpora/__init__.py", line 6, in <module>
    from .indexedcorpus import IndexedCorpus  # noqa:F401 must appear before the other classes
    ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/Users/smahendra/opt/anaconda3/envs/gpt11/lib/python3.11/site-packages/gensim/corpora/indexedcorpus.py", line 14, in <module>
    from gensim import interfaces, utils
  File "/Users/smahendra/opt/anaconda3/envs/gpt11/lib/python3.11/site-packages/gensim/interfaces.py", line 19, in <module>
    from gensim import utils, matutils
  File "/Users/smahendra/opt/anaconda3/envs/gpt11/lib/python3.11/site-packages/gensim/matutils.py", line 20, in <module>
    from scipy.linalg import get_blas_funcs, triu
ImportError: cannot import name 'triu' from 'scipy.linalg' (/Users/smahendra/opt/anaconda3/envs/gpt11/lib/python3.11/site-packages/scipy/linalg/__init__.py)

```

Update:
flair is importing gensim which is importing scipy. On April-2-24, scipy removed the triu from linalg module.
[...]

PR #3461, opened 2024-05-26
quick fix for py3.12

tavallaie:
adding genism@da2f388 for supporting python 3.12

Solving #3438

https://blog.techbend.io/overcoming-the-hazm-compatibility-challenge-with-python-312


helpmefindaname:
Hi @tavallaie 
thank you for taking kind of this issue, but I disagree that this should be merged into libraries.
The root issue lies in gensim not having released the fixes that are on their current develop branch. The fix to this is, that gensim needs to release a new version that is stable, which they are also [planning to do soon](https://github.com/piskvorky/gensim/issues/3510).
Until then, nothing prevents you from using flair by running `pip install flair git+https://github.com/piskvorky/gensim.git@da2f388` or any of the other solutions suggested [here](https://github.com/flairNLP/flair/issues/3441).
Pinning a version within a library would be bad, as it would force everyone to use that specific version, no matter if they are using python3.12 or were going for older python3.8

david-waterworth:
This also  doesn't have anything to do with Python 3.12 either as far as I understand. The issue is with gensim using a deprecated scipy function but not setting the requirements correctly (if you knowingly use a deprecated function you should at least set an upper bound on the library you're using until you're ready to fix) - it occurs on all version of python.

PR #3463, opened 2024-05-30
Support for NoisyNER dataset

teresaloeffelhardt:
Hi,

This PR adds support for the NoisyNER dataset as proposed in [this paper](https://ojs.aaai.org/index.php/AAAI/article/view/16938) and released in [this repo](https://github.com/uds-lsv/NoisyNER):

> NoisyNER is a dataset for the evaluation of methods to handle noisy labels when training machine learning models. It is from the NLP/Information Extraction domain and was created through a realistic distant supervision technique. Some highlights and interesting aspects of the data are:
> * Seven sets of labels with differing noise patterns to evaluate different noise levels on the same instances
> * Full parallel clean labels available to compute upper performance bounds or study scenarios where a small amount of gold-standard data can be leveraged
> * Skewed label distribution (typical for Named Entity Recognition tasks)
> * For some label sets: noise level higher than the true label probability
> * Sequential dependencies between the labels


flair/datasets/sequence_labeling.py
```diff
@@ -4839,6 +4839,137 @@ def __init__(
...
+    def __init__(
+        self,
+        version: int = 0,
+        base_path: Optional[Union[str, Path]] = None,
+        in_memory: bool = True,
+        **corpusargs,
+    ) -> None:
+        """Initialize the NoisyNER corpus.
+
+        :param version: Chooses the labelset for the data.
```
helpmefindaname, line 4855: please use the [google docstring](https://sphinxcontrib-napoleon.readthedocs.io/en/latest/example_google.html) format as described in our [docs](https://flairnlp.github.io/flair/v0.13.1/contributing/updating_documentation.html#docstrings)

helpmefindaname (changes requested):

flair/datasets/sequence_labeling.py
teresaloeffelhardt: Hi. I have made all the requested changes. Is the description in the correct format now?

alanakbik:
@teresaloeffelhardt thanks a lot for adding this!

PR #3466, opened 2024-06-06
QuickFix: extract hidden states of transformers by key

DavidReuschenberg:
Hi,

The extraction of hidden states within transformer embeddings was previously done by accessing the last item in the transformer model output. This causes some older models, e.g. **dvm1983/TinyBERT_General_4L_312D_de** to crash during inference, due to a different output ordering.
This is fixed by extracting the hidden states by the corresponding key instead, which is also more readable.

### DIFF
--- a/flair/embeddings/transformer.py
+++ b/flair/embeddings/transformer.py
@@ -1321,7 +1321,7 @@ class TransformerEmbeddings(TransformerBaseEmbeddings):
             model_kwargs["bbox"] = bbox
         if pixel_values is not None:
             model_kwargs["pixel_values"] = pixel_values
-        hidden_states = self.model(input_ids, **model_kwargs)[-1]
+        hidden_states = self.model(input_ids, **model_kwargs).hidden_states
         # make the tuple a tensor; makes working with it easier.
         hidden_states = torch.stack(hidden_states)
 


helpmefindaname:
Hi @DavidReuschenberg
Thank you for trying to improve this library, however as seen in the test pipeline, the changes crash on most models.
The issue is that the specific model you pointed out, has set different default values for `output_attention` and `output_hidden_states` in their [configuration](https://huggingface.co/dvm1983/TinyBERT_General_4L_312D_de/blob/main/config.json#L17).
For this to solve, I suggest one of 3 solutions:
* Specify those values when loading the embeddings, e.g.: `TransformerWordEmbeddings("dvm1983/TinyBERT_General_4L_312D_de", output_attentions=False, output_hidden_states=False)`
* [contribute](https://huggingface.co/dvm1983/TinyBERT_General_4L_312D_de/edit/main/config.json) to the model on the hf hub to not overwrite the defaults and therefore act the same as most models on the hf-hub.
* Ensure that the TransformerEmbeddings always set the atributes to false, so it will ignore the configurations of specific models

DavidReuschenberg:
Thank you @helpmefindaname , changing the parameter is a much easier and better solution of course.

Issue #3467, opened 2024-06-06
[Bug]: BytePairEmbeddings are not loading for English


pjdrm-unbabel:
### Describe the bug

Can't load BytePairEmbeddings, it throws a `UnicodeDecodeError`. Maybe the files from https://nlp.h-its.org/bpemb got updated and are no longer compatible.

### To Reproduce

```python
from flair.embeddings import BytePairEmbeddings
BytePairEmbeddings("en")
```

### Expected behavior

BytePairEmbeddings should load without errors.

### Logs and Stack traces

```stacktrace
Traceback (most recent call last):
  File "<stdin>", line 1, in <module>
  File "my_venvs/venv_flair13/lib/python3.9/site-packages/flair/embeddings/token.py", line 1438, in __init__
    self.embedder = BPEmb(
  File "/my_venvs/venv_flair13/lib/python3.9/site-packages/bpemb/bpemb.py", line 191, in __init__
    self.emb = load_word2vec_file(self.emb_file, add_pad=add_pad_emb)
  File "/my_venvs/venv_flair13/lib/python3.9/site-packages/bpemb/util.py", line 79, in load_word2vec_file
    vecs = KeyedVectors.load_word2vec_format(word2vec_file, binary=binary)
  File "/my_venvs/venv_flair13/lib/python3.9/site-packages/gensim/models/keyedvectors.py", line 1719, in load_word2vec_format
    return _load_word2vec_format(
  File "/my_venvs/venv_flair13/lib/python3.9/site-packages/gensim/models/keyedvectors.py", line 2058, in _load_word2vec_format
    header = utils.to_unicode(fin.readline(), encoding=encoding)
  File "/my_venvs/venv_flair13/lib/python3.9/site-packages/gensim/utils.py", line 365, in any2unicode
    return str(text, encoding, errors=errors)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0x8b in position 1: invalid start byte
```

##### Flair
0.13.1
##### Pytorch
2.3.1
##### Transformers
4.41.2
#### GPU
False

stefan-it:
Hi @pjdrm-unbabel ,

here's discussion about this bug in upstream repo:

https://github.com/bheinzerling/bpemb/issues/66

Fix is to just update to latest version of `bpemb` :)

pjdrm-unbabel:
Yep, that fixes it. Thank you for the pointer. Could be worth bumping the min version on the flair repo to avoid tripping into this.

PR #3468, opened 2024-06-06
pip: bump min. version of bpemb

stefan-it:
Hi,

due to a recent bugfix release of `bpemb` (see https://github.com/bheinzerling/bpemb/issues/66 ), we should use this version as minimum required one!

In versions < 0.3.5 the download of embeddings is no longer working.

Related to #3467.


helpmefindaname (approved):

alanakbik:
@stefan-it thanks for fixing this!

Issue #3474, opened 2024-06-21
[Feature]: add random seed to dataset functions for reproducibility


MattGPT-ai:
### Problem statement

When performing random splitting and downsampling of datasets, including when loading Flair datasets, there is no option to set the random seed specifically for that function. You can set random seed outside of the function, but this does not seem to provide as full control and reproducibility if you want to set specific different random seeds for each operation. I have had trouble controlling this when instantiating datasets that can be imported from Flair.

### Solution

Ideally there would be an option to pass a random seed to functions like `randomly_split_into_two_datasets` and `downsample`. Perhaps you could pass in a random seed to dataset objects like `flair.datasets.sequence_labeling.CONLL_03`

PR #3475, opened 2024-06-21
GH-3474: add random seed parameter to dataset splitting and downsampl…

MattGPT-ai:
https://github.com/flairNLP/flair/issues/3474

Adds random seed parameters to dataset splitting and downsampling functions

### DIFF
--- a/flair/data.py
+++ b/flair/data.py
@@ -1301,6 +1301,7 @@ class Corpus(typing.Generic[T_co]):
         test: Optional[Dataset[T_co]] = None,
         name: str = "corpus",
         sample_missing_splits: Union[bool, str] = True,
+        random_seed: Optional[int] = None,
     ) -> None:
         # set name
         self.name: str = name
@@ -1314,7 +1315,7 @@ class Corpus(typing.Generic[T_co]):
             test_portion = 0.1
             train_length = _len_dataset(train)
             test_size: int = round(train_length * test_portion)
-            test, train = randomly_split_into_two_datasets(train, test_size)
+            test, train = randomly_split_into_two_datasets(train, test_size, random_seed)
             log.warning(
                 "No test split found. Using %.0f%% (i.e. %d samples) of the train split as test data",
                 test_portion,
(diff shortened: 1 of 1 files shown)


helpmefindaname (changes requested):
Hi @MattGPT-ai
thank you for your contribution.
The changes look good to me, I just requests a small change. Besides that, I think this is ready to merge

flair/data.py
```diff
@@ -1879,11 +1882,21 @@ def iob2(tags):
     return True
 
 
-def randomly_split_into_two_datasets(dataset, length_of_first):
+def randomly_split_into_two_datasets(
+    dataset: Dataset, length_of_first: int, random_seed: Optional[int] = None
+) -> tuple[Subset, Subset]:
```
helpmefindaname, line 1887: Please use `Tuple` (from typing import Tuple) to allow compability with older python versions

MattGPT-ai:
Thanks for your review! Pushed the changes you requested

helpmefindaname (approved):

MattGPT-ai:
Thanks for the approval! How can we trigger the checks to run so that it can be merged?

flair/data.py
```diff
@@ -1879,11 +1882,21 @@ def iob2(tags):
...
+def randomly_split_into_two_datasets(
+    dataset: Dataset, length_of_first: int, random_seed: Optional[int] = None
+) -> Tuple[Subset, Subset]:
+    """Shuffles a dataset and splits into two subsets.
+
+    The length of the first is specified and the remaining samples go into the second subset.
+    """
     import random
 
     indices = list(range(len(dataset)))
```
alanakbik, line 1894:
```suggestion
    indices = list(range(_len_dataset(dataset)))
```

alanakbik (changes requested):
Thanks for the PR! Looks good, but there is a small mypy error (I added a suggestion to fix).

MattGPT-ai:
Applied the change. I had a question, I saw that step is necessary because Dataset doesn't have `__len__` implemented. Is there a way to implement `__len__`, even if it just calls `_len_dataset`? Does constructing the `DatasetLoader` class take any non-negligible time? I actually see the same warning for `return len(loader)` in my PyCharm that it "Expected type Sized, got DataLoader instead", and a similar warning for `_iter_dataset`

alanakbik:
Thanks! I agree that it would be better to implement `__len__` rather than having to call `_len_dataset`. 

The problem is that `Dataset` is a PyTorch class that does not have `__len__` or `__iter__` implemented, so we can't add them here. Alternatively, we might change the code that instead of `Dataset`, everything is based on either PyTorch's `IterDataset` or our own `FlairDataset`.

alanakbik (approved):

alanakbik:
@MattGPT-ai thanks a lot for adding this!

Issue #3459, opened 2024-05-19
[Bug]: Custom trained model not correctly loaded when using FlairEmbeddings


petermartens1992:
### Describe the bug

The F-scores shown after training a SequenceTagger model with FlairEmbeddings are way higher than when loading the model from file and doing an evaluation on it. Also the predictions from a loaded model seem way off compared to the initial F-scores.

The issue doesn't appear when initializing a new model and only loading the state from file:

```
sequence_tagger.load_state_dict(SequenceTagger.load(model_path).state_dict())
```

Also, when removing the FlairEmbeddings the F-scores between training and evaluation correctly align again.

See here the code to reproduce the issue (using flair 0.13.1):

```
import sys
import logging
import os

from typing import List
from pathlib import Path

from flair.datasets import UD_ENGLISH
from flair.data import Corpus
from flair.models import SequenceTagger
from flair.embeddings import TokenEmbeddings, WordEmbeddings, StackedEmbeddings, FlairEmbeddings, CharacterEmbeddings, OneHotEmbeddings, TransformerWordEmbeddings
from flair.datasets import ColumnCorpus
from flair.trainers import ModelTrainer

language = "nl"
model_type = "upos"

configs["train"] = {
    "language": language,
    "model_type": model_type,
    "tag_format": "BIOES",
    "rnn_hidden_size": 256,
    "rnn_layers": 1,
    "use_crf": True,
    "word_dropout": 0.2,
    "dropout": 0.2,
    "learning_rate": 0.1,
    "mini_batch_size": 32,
    "mini_batch_chunk_size": 32,
    "train_initial_hidden_state": True,
    "embeddings_storage_ratio": 1.0,
    "fine_tune_flair_embeddings": False,
}

output_model_name = configs["train"]["language"] + "-flair-" + configs["train"]["model_source"] + "-example"

output_path = "~/models/flair/" + configs["train"]["model_type"] + "/" + output_model_name
model_path = output_path + "/best-model.pt"
logger.info("model path: " + model_path)

# 1. what label do we want to predict?
label_type = configs["train"]["model_type"]

# 2. load the corpus
corpus = UD_ENGLISH().downsample(0.1)
print(corpus)

# 3. make the label dictionary from the corpus
label_dict = corpus.make_label_dictionary(label_type=label_type)
print(label_dict)

# 4. initialize embeddings
embedding_types: List[TokenEmbeddings] = [
    WordEmbeddings(configs["train"]["language"] + "-wiki"),

[...]

PR #3477, opened 2024-06-28
character embeddings store their embedding name too

helpmefindaname:
Fixes https://github.com/flairNLP/flair/issues/3459


alanakbik:
Can confirm that the issue in #3459 is fixed. Thanks @helpmefindaname!

alanakbik (approved):

PR #3480, opened 2024-07-01
Add PEFT training and explicit kwarg passthrough

janpf:
This PR adds the ability to train models using PEFT (LoRA and QLoRA) and some nicer handling for model and config explicit kwargs. For example, passing through kwargs to the model but not the config was not possible before.

If PEFT is not installed and not used, no error should be thrown either.


alanakbik:
Hello @janpf could you provide a small test script how to train a model (for instance for NER) using PEFT? That would make it easier to test.

janpf:
will do. hopefully this week :)

alanakbik:
Hi @janpf this looks good. I tested for a standard BERT model (for which quantization seems not to be available), and I'm getting competitive results to full fine-tuning when setting a slightly higher learning rate for LoRA: 

```python
from peft import LoraConfig, TaskType

document_embeddings = TransformerDocumentEmbeddings(
    "bert-base-uncased",
    fine_tune=True,
    # set LoRA config
    peft_config=LoraConfig(
        task_type=TaskType.FEATURE_EXTRACTION,
        inference_mode=False,
    ),
)

classifier = TextClassifier(document_embeddings, label_dictionary=label_dict, label_type=label_type)
trainer.fine_tune(
    "resources/taggers/question-classification-with-transformer",
    learning_rate=5.0e-4,
    mini_batch_size=4,
    max_epochs=1,
)
```

Unfortunately, I don't know what is causing the storage error. This is now affecting all PRs.

alanakbik:
Thanks again for adding this @janpf! Since the tests are now running through, we can merge!

PR #3481, opened 2024-07-01
Adds DataTriples which act just like DataPairs

janpf:
This adds a new DataPoint type consisting of three sentences instead of two. This could/should likely be parameterizable, but this is the easiest way to add this.


flair/data.py
```diff
@@ -1258,6 +1259,50 @@ def text(self):
...
+        self.first = first
+        self.second = second
+        self.third = third
+
+    def to(self, device: str, pin_memory: bool = False):
+        self.first.to(device, pin_memory)
+        self.second.to(device, pin_memory)
+        self.third.to(device, pin_memory)
+
+    def clear_embeddings(self, embedding_names: List[str] = None):
```
alanakbik, line 1274:
```suggestion
    def clear_embeddings(self, embedding_names: Optional[List[str]] = None):
```

alanakbik: Should be consistent with signature of supertype

flair/datasets/text_text.py
```diff
@@ -435,6 +444,264 @@ def __getitem__(self, index: int = 0) -> DataPair:
...
+        data_folder: Union[str, Path],
+        columns: List[int] = [0, 1, 2, 3],
+        train_file=None,
+        test_file=None,
+        dev_file=None,
+        use_tokenizer: bool = True,
+        max_tokens_per_doc=-1,
+        max_chars_per_doc=-1,
+        in_memory: bool = True,
+        label_type: str = None,
```
alanakbik, line 459:
```suggestion
        label_type: Optional[str] = None,
```

alanakbik: Implicit optionals are no longer allowed

flair/datasets/text_text.py
```diff
@@ -435,6 +444,264 @@ def __getitem__(self, index: int = 0) -> DataPair:
...
+class DataTripleDataset(FlairDataset):
+    def __init__(
+        self,
+        path_to_data: Union[str, Path],
+        columns: List[int] = [0, 1, 2, 3],
+        max_tokens_per_doc=-1,
+        max_chars_per_doc=-1,
+        use_tokenizer=True,
+        in_memory: bool = True,
+        label_type: str = None,
```
alanakbik, line 572:
```suggestion
        label_type: Optional[str] = None,
```

alanakbik (changes requested):
Thanks for the PR! Just some minor mypy problems, concerning implicit optionals. I've added some suggestions as comments to fix the issues.

alanakbik (approved):

alanakbik:
@janpf thanks a lot for adding this! Will you also add some prepared corpora for the SuperGLEBler datasets in future PRs?

janpf:
> @janpf thanks a lot for adding this! Will you also add some prepared corpora for the SuperGLEBler datasets in future PRs?

Thanks for the quick merge, @alanakbik :)
We are thinking about it. Currently, we provide the datasets via the repo, which was the fastest way to deliver the files. I think we'll make sure version 0.13 works nicely, and then we'll see :)

Is there an ETA for when the two PRs might land on pypi? :)

alanakbik:
I'm hoping to find enough time to prepare a new release once the semester ends (in two weeks).

Issue #3487, opened 2024-07-03
[Bug]: GPU memory leak in TextPairRegressor when embed_separately is set to `False`


MattGPT-ai:
### Describe the bug

When training a `TextPairRegressor` model with `embed_separately=False` (the default), via e.g. `ModelTrainer.fine_tune`, the GPU memory slowly creeps up with each batch, eventually causing an OOM even when the model and a single batch fits easily in GPU memory. 

The function `store_embeddings` is supposed to clear any embeddings of each DataPoint. For this model, the type of data point is `TextPair`. It actually does seem to handle clearing `text_pair.first` and `.second` when `embed_separately=True`, because it runs embed for each sentence (see `TextPairRegressor._get_embedding_for_data_point`), and that embedding is attached to each sentence so it can be referenced via the sentence.

However, the default setting is `False`; in that case, to embed the pair, it concatenates the text of both sentences (adding a separator), creates a new sentence, embeds that sentence, and then returns that embedding. Since it's never attached to the `DataPoint` object, `clear_embeddings` doesn't find it when you iterate over the data points. The function `identify_dynamic_embeddings` also always comes up empty

### To Reproduce

```python
import flair
from flair.data import DataPairCorpus
from flair.models import TextPairRegressor

search_rel_corpus = DataPairCorpus(Path('text_pair_dataset'), train_file='train.tsv', test_file='test.tsv', dev_file='dev.tsv', label_type='relevance', in_memory=False)
text_pair_regressor = TextPairRegressor(embeddings=embeddings, label_type='relevance')

embeddings = TransformerDocumentEmbeddings(
    model='xlm-roberta-base',
    layers="-1",
    subtoken_pooling='first',
    fine_tune=True,
    use_context=True,
    is_word_embedding=True,
)

trainer = ModelTrainer(text_pair_regressor, search_rel_corpus)

trainer.fine_tune(
    "relevance_regressor",
    learning_rate=1e-5,
    epoch=0,
    max_epochs=5,
    mini_batch_size=4,
    save_optimizer_state=True,
    save_model_each_k_epochs=1,
    use_amp=True,  # aka Automatic Mixed Precision, e.g. float16
)
```

### Expected behavior

The memory should remain relatively flat with each epoch of training if memory is cleared correctly. In other training, such as for a `TextClassifier`, it stays roughly the same after each mini-batch, 

### Logs and Stack traces

```stacktrace
OutOfMemoryError                          Traceback (most recent call last)
[...]

PR #3490, opened 2024-07-03
Fix GPU memory leak in TextPairRegressor

MattGPT-ai:
https://github.com/flairNLP/flair/issues/3487

This addresses the above memory leak by saving a reference to the `TextPair` when the concatenated `Sentence` is created. Its embeddings are explicitly cleared when `clear_embeddings` is called.

This also adds type hints to files relevant to this change.

### DIFF
--- a/flair/data.py
+++ b/flair/data.py
@@ -5,8 +5,9 @@ import typing
 from abc import ABC, abstractmethod
 from collections import Counter, defaultdict
 from operator import itemgetter
+from os import PathLike
 from pathlib import Path
-from typing import Dict, Iterable, List, NamedTuple, Optional, Tuple, Union, cast
+from typing import Any, DefaultDict, Dict, Iterable, List, NamedTuple, Optional, Tuple, Union, cast
 
 import torch
 from deprecated.sphinx import deprecated
@@ -49,7 +50,7 @@ class BoundingBox(NamedTuple):
 class Dictionary:
     """This class holds a dictionary that maps strings to IDs, used to generate one-hot encodings of strings."""
 
-    def __init__(self, add_unk=True) -> None:
+    def __init__(self, add_unk: bool = True) -> None:
         # init dictionaries
         self.item2idx: Dict[bytes, int] = {}
         self.idx2item: List[bytes] = []
@@ -143,21 +144,21 @@ class Dictionary:
     def start_stop_tags_are_set(self) -> bool:
         return {b"<START>", b"<STOP>"}.issubset(self.item2idx.keys())
 
-    def save(self, savefile):
+    def save(self, savefile: PathLike):
         import pickle
 
         with open(savefile, "wb") as f:
             mappings = {"idx2item": self.idx2item, "item2idx": self.item2idx}
             pickle.dump(mappings, f)
 
-    def __setstate__(self, d):
+    def __setstate__(self, d: Dict) -> None:
         self.__dict__ = d
         # set 'add_unk' if the dictionary was created with a version of Flair older than 0.9
         if "add_unk" not in self.__dict__:
             self.__dict__["add_unk"] = b"<unk>" in self.__dict__["idx2item"]
 
     @classmethod
-    def load_from_file(cls, filename: Union[str, Path]):
+    def load_from_file(cls, filename: Union[str, Path]) -> "Dictionary":
         import pickle
 
         with Path(filename).open("rb") as f:
@@ -174,7 +175,7 @@ class Dictionary:
         return dictionary
 
     @classmethod
-    def load(cls, name: str):
+    def load(cls, name: str) -> "Dictionary":
         from flair.file_utils import cached_path
 
         hu_path: str = "https://flair.informatik.hu-berlin.de/resources/characters"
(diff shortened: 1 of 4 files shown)


MattGPT-ai:
Just pushed a type hint change for a file that I didn't really touch, but I think is checking because I added type hints to `Sentence`:
`predicted: List[List[Union[int, float]]] = [[] for _ in range(number_tokens)]`

Not sure if it's the case that this should just be `List[List[int]]` and the line `predicted[i].append(input_indices[i].item())` should just cast `int(input_indices[i].item())` - but that seemed riskier since it's changing code and not just the type hint. 
Any comments here?

flair/data.py
```diff
@@ -282,11 +283,11 @@ class DataPoint:
     def __init__(self) -> None:
         self.annotation_layers: Dict[str, List[Label]] = {}
         self._embeddings: Dict[str, torch.Tensor] = {}
-        self._metadata: Dict[str, typing.Any] = {}
+        self._metadata: Dict[str, Any] = {}
 
     @property
     @abstractmethod
-    def embedding(self):
+    def embedding(self) -> None:
```
alanakbik, line 290: The `embedding` property returns an embedding tensor. All classes that inherit from `DataPoint` (such as `Sentence`) are embeddable and should thus be able to return an embedding tensor.

A snippet to show the property in action: 

```python

embeddings = TransformerDocumentEmbeddings("distilbert-base-uncased")

sentence = Sentence("hello")

embeddings.embed(sentence)

# print the embedding of the data point
print(sentence.embedding)
```

flair/models/pairwise_regression_model.py
```diff
@@ -185,6 +185,7 @@ def _get_embedding_for_data_point(self, prediction_data_point: TextPair) -> torc
                 use_tokenizer=False,
             )
             self.embeddings.embed(concatenated_sentence)
+            prediction_data_point.concatenated_data = concatenated_sentence
```
alanakbik, line 188: I like the idea of saving the concatenated version of the sentence pair on the data point itself. Since during training, we iterate over the same data points over and over again, you could include a check here to see whether the TextPair already has a concatenated version. 

I.e. the else statement could look like this: 

```python
        else:
            # If the concatenated version of the text pair does not exist yet, create it
            if prediction_data_point.concatenated_data is None:
                concatenated_sentence = Sentence(
                    prediction_data_point.first.to_tokenized_string()
                    + self.sep
                    + prediction_data_point.second.to_tokenized_string(),
                    use_tokenizer=False,
                )
                prediction_data_point.concatenated_data = concatenated_sentence

            # Embed the concatenated version of the data point and return embedding
            self.embeddings.embed(prediction_data_point.concatenated_data)
            return prediction_data_point.concatenated_data.get_embedding(embedding_names)
```

alanakbik (changes requested):
Thanks for adding this! Mostly looks good. Just please correct the return signature of the embedding property, as indicated in the comment above. 

I also added a suggestion on checking for the concatenated data point, which may give a minimal speed-up during training (though likely negligible so up to you if you want to include it).

Regarding the lemmatizer, I'm not so sure. Probably best to err on the side of caution and use the old, underspecified signature.

mattb-zip:
I have addressed all of the requests I believe, let me know if you can approve or request any further changes

MattGPT-ai:
I just realized that this same issue exists for `TextPairClassifier`

The simple fix is to just apply the same logic to `_get_embedding_for_data_point` -- however, since these two functions are identical, it might be good to do a small refactor where they each call the same common function for embedding a `DataPair`.

Since these inherit from different base classes, it doesn't seem like the best idea to try to do this via inheritance

alanakbik (approved):

alanakbik:
@MattGPT-ai thanks for fixing this and adding the type hints! 

I'll merge this now. The same bug in `TextPairClassifier` could be fixed similarly. Agree that addressing this via inheritance would be overkill since this only affects two classes (and doing too much via inheritance tends to make code less intuitive).

MattGPT-ai:
I actually pushed an additional commit to fix it in `TextPairClassifier`, just did the same fix. I have an idea for refactoring this: make `_get_embedding_for_data_point` a function in `TextPair` - right now it's just a type definition, but it could be made a class that inherits `DataPair[Sentence, Sentence]`. Then the models that use `TextPair` would just use its own function to embed. Could have a single function or two functions for embedding separately or jointly. 

This would give better control over how these embeddings are set in the `TextPair`. One problem I thought of is that if we're saving the concatenated sentence, you could have an inconsistency if you make a change to one of the texts in the pair, and it would not be updated accordingly. This refactor would allow us to have a setter function for each member of the pair that would automatically update the concatenated sentence. I think right now we're relying too much on the models to handle all of this, but all this functionality should exist in the data class independent of the models. 

Currently in `DataPair`, you have the property `embedding` that returns the concatenation of the two member embeddings concatenated, which is more in line with doing separate embeddings. That might be a decent default, but I think it makes sense to override that for the case of `TextPair`, and it might even be good to have a property that sets if its saved embedding is separate or joint. Not sure about that, but your thoughts would be much appreciated. If you think that makes sense, I could push a PR and maybe create a separate issue

alanakbik:
Hi @MattGPT-ai, is it a good idea and agree that saving the concatenated data point may lead to problematic inconsistencies.

I'm a bit worried though that moving too much logic into the data class will cause the logic to become too distributed. Right now, the `TextPairClassifier` makes all model-relevant decisions through `_get_embedding_for_data_point`: whether or not to concat, what separator to use, etc. Perhaps you could prepare a PR and we can discuss it?

Issue #3482, opened 2024-07-02
[Feature]: upgrade urllib3


david-waterworth:
### Problem statement

flair requires `urllib3<2.0.0` which is quite old (latest is 2.2, the 2.0 alpha was released in late 2022)

https://github.com/flairNLP/flair/blob/59bd7053a1a73da03293b0bdb68113cf383d6b9e/requirements.txt#L26C1-L26C75

I have other dependencies that require urllib3 >= 2.0 (i.e. `tritonclient[http] (>=2.47.0,<3.0.0) requires urllib3 (>=2.0.7)`)

### Solution

I'm not sure that flair directly uses `urllib3`, I suspect the require was added as per the comment due to 3rd party libraries not requiring urllib3 < 2 so resulting in a broken installation when urllib3 v2 was released. I would hope this is no longer a constraint and it can be removed?

### Additional Context

In general I think flair imports to much by default, i.e. I have to downgrade `scipy` because even though I'm not using any `gensim` based features the fact that gensim uses deprecated scipy features that are now removed (triu) I get import errors (gensim has fixed this but not released the fix yet). Also note a recent "fix" added by gensim pins `numpy < 2` which flair will inherit which could cause issues down the line.

I think most 3rd party (gensim, spacy, huggingface) functionality should be "optional" - i.e. `pip install flair[gensim]` to enable gensim features. I'm not a big fan of getting errors due to flair importing gensim when I'm training a tagger with transformer embeddings.

alanakbik:
Hi @david-waterworth, generally agree that we have too many dependencies.

@helpmefindaname did an analysis of all dependencies and identified a core set of dependencies plus optional ones, also wanting to get rid of gensim. He proposed a solution essentially exactly like you, e.g. using something like `pip install flair[gensim]` to install bigger libraries. 

My main problem is that gensim is used in some of our main taggers (for instance 'ner') and then this starter example would not work anymore: https://flairnlp.github.io/docs/intro#requirements-and-installation (installation plus example 1).

david-waterworth:
The example appears to work for me without using gensim features? My assumption is gensim is used by the `WordEmbeddings` class in `flair/embeddings/token.py` (specifically `gensim.models.KeyedVectors` to load word2vec serialized files). My testing was a bit ad-hoc though so I could very well have missed something.

[...]

PR #3493, opened 2024-07-07
make gensim optional

helpmefindaname:
Closes https://github.com/flairNLP/flair/issues/3482 

This requires all public models to be in serialized format, so they can be loaded without attempt to load gensim.

Now you can use `WordEmbeddings`, `BytePairEmbeddings` in inference without having gensim/bpe installed.

`FasttextEmbeddings` and `MuseEmbeddings` will only work when gensim is installed, I find this justifyable, as those embeddings are not commonly used anymore.

When instanciating new `WordEmbeddings` or `BytePairEmbeddings`, gensim/bpe is required. You can install them with `pip install flair[word-embeddings]` after the next release or `pip install -e .[word-embeddings]` when developing.


alanakbik (changes requested):
Thanks a lot for this @helpmefindaname. This generally looks good. 

I tested loading our standard 'ner' model that I re-serialized and pushed to the hub as "alanakbik/ner-new". 

I then tested loading with the following code:

```python

model = SequenceTagger.load("alanakbik/ner-new")

sentence = Sentence("Bill was born in New York")

model.predict(sentence)

print(sentence)
```

That works, but only for newer flair versions. It starts breaking down from Flair 0.10.0. 

To reproduce, in a fresh env, do: 

`pip install flair==0.10.0`

and run the above code. 

This throws the error: 

`AttributeError: 'dict' object has no attribute 'embedding_length'`

I think this actually has nothing directly to do with this PR, but affects all new models that were trained with newer Flair versions and are loaded with older ones. 

But to deploy this PR, we'd need to update all models. So if people are still using an old version of Flair, the regular 'ner' model would no longer work. 

@helpmefindaname can you take a look if backward compatibility can be improved?

alanakbik:
@helpmefindaname thanks a lot for adding this! I will now update all models on HF.

itsmedonttell:
The pip command recommended by the gensim import error message didn't work out of the box for me. I use Terminal and zsh on macOS. I had to put quotes around "flair[word-embeddings]" when using pip install (like so: `pip install "flair[word-embeddings]"`), or else zsh throws an error, and the install doesn't execute. Alternatively, `pip install genism` works.

Issue #3488, opened 2024-07-03
[Feature]: Function to write a ColumnCorpus instance to files


chelseagzr:
### Problem statement

It would be nice if there is a function to write a ColumnCorpus instance to files. 

Related issues:
https://github.com/flairNLP/flair/issues/988
https://github.com/flairNLP/flair/issues/2591

### Solution

Add a method `write_to_directory` to the ColumnCorpus class:

```
class ColumnCorpus(MultiFileColumnCorpus):

    def write_to_directory(self, label_types: list[str], output_directory: Path):
        """With a list of label_types specified, write the train, dev, test datasets of this corpus to files in CoNLL format."""
```

### Additional Context

Is there any reason the ColumnCorpus class should not have a function like this?
If not, I can draft a PR for this feature.

helpmefindaname:
Hi @chelseagzr

I don't see a reason to *not* have this. PR's are welcome, thank you for volunteering :)

PR #3497, opened 2024-07-11
GH-3488: Support for writing a ColumnCorpus instance to files

chelseagzr:
This PR addresses https://github.com/flairNLP/flair/issues/3488

Current limitations:
- Only token level or span level sequence tagging labels are supported by `write_to_directory`
- The `whitespace_after` attribute of tokens will not be preserved after saving the corpus to files; only `default_whitespace_after` attribute of datasets will be preserved.


flair/datasets/sequence_labeling.py
```diff
@@ -443,6 +444,161 @@ def __init__(
...
+    def get_token_level_label_of_each_token(sentence: Sentence, label_type: str) -> List[str]:
+        """Generates a label for each token in the sentence. This function requires that the labels corresponding to the label_type are token-level tokens.
+
+        Args:
+            sentence: a flair sentence to generate labels for
+            label_type: a string representing the type of the labels, e.g., "pos"
+        """
+        list_of_labels = ["O" for _ in range(len(sentence.tokens))]
+        for label in sentence.get_labels(label_type):
+            label_token_index = label.data_point._internal_index
```
mattb-zip, line 457: Not so much a comment on this PR, but I wonder why this member is protected when it is needed sometimes outside of the class

mattb-zip (approved):

flair/datasets/sequence_labeling.py
```diff
@@ -443,6 +444,161 @@ def __init__(
...
+            list_of_labels[label_token_index - 1] = label.value
+        return list_of_labels
+
+    @staticmethod
+    def get_span_level_label_of_each_token(sentence: Sentence, label_type: str) -> List[str]:
```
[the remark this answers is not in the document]
chelseagzr: I can make it a method of the `Sentence` class if the maintainers also agree.

flair/datasets/sequence_labeling.py
```diff
@@ -443,6 +434,167 @@ def __init__(
...
+        nonempty_dataset = self.train or self.dev or self.test
+        for _ in range(max_depth):
+            if type(nonempty_dataset) is ColumnDataset:
+                break
+            elif type(nonempty_dataset) is ConcatDataset:
+                nonempty_dataset = nonempty_dataset.datasets[0]
+            elif type(nonempty_dataset) is Subset:
+                nonempty_dataset = nonempty_dataset.dataset
+            else:
+                raise NotImplementedError("Unsupported type")
```
alanakbik, line 559: This part is a bit hard to read. Can you add some code comments to explain (i.e. that sometimes, a split is a concat or subset and you need to get the type of the original dataset)

flair/datasets/sequence_labeling.py
```diff
@@ -443,6 +434,167 @@ def __init__(
...
+                            return Token
+                        elif isinstance(label.data_point, Span):
+                            return Span
+                        else:
+                            raise NotImplementedError(
+                                f"The level of {label_type} is neither token nor span. Only token level labels and span level labels can be handled now."
+                            )
+        raise RuntimeError(f"There is no label of type {label_type} in this corpus.")
+
+    def write_corpus_meta_data(
```
alanakbik, line 531: This method has only one usage, i.e. from the `write_to_directory` method. It can be indicated as an internal method by adding a "_" to the method name (`_write_corpus_meta_data`).

flair/datasets/sequence_labeling.py
```diff
@@ -443,6 +434,167 @@ def __init__(
...
+        meta_data["column_format"] = {int(key): value for key, value in meta_data["column_format"].items()}
+
+        return cls(
+            data_folder=directory,
+            autofind_splits=True,
+            skip_first_line=False,
+            **meta_data,
+        )
+
+    def get_level_of_label(self, label_type: str) -> Union[Type[Token], Type[Span]]:
```
alanakbik, line 510: This method has only one usage, namely in the `write_to_directory` method. Making it public to the `ColumnCorpus` might be confusing to users. You can declare it as internal by prefixing the method name with a "_".

flair/datasets/sequence_labeling.py
```diff
@@ -443,6 +434,167 @@ def __init__(
...
+            tokens = label.data_point.tokens
+            start_token_index = tokens[0]._internal_index
+            list_of_labels[start_token_index - 1] = f"B-{label.value}"
+            for token in tokens[1:]:
+                token_index = token._internal_index
+                list_of_labels[token_index - 1] = f"I-{label.value}"
+        return list_of_labels
+
+    @staticmethod
+    def write_dataset_to_file(
```
alanakbik, line 470: This could also be a non-static method of ColumnDataset, directly callable in case someone wants to serialize one dataset only.

alanakbik (changes requested):
Thanks for adding this! 

One issue is that this adds many public methods to ColumnCorpus and ColumnDataset, which may be confusing to users since most of these are used only for a single functionality. I suggest leaving only the respective "save" and "load" method public (since these are the ones that users will call), and marking all others as internal.

MattGPT-ai (approved):

chelseagzr:
@alanakbik Could you please review the new changes when you have a chance? Thank you!

chelseagzr:
@alanakbik Could you please review the new changes when you have a chance? Thank you!

alanakbik:
hello @chelseagzr very sorry it took so long! I updated your branch for our current master, but one of the imports got lost. Could you add it back in?

PR #3502, opened 2024-07-20
Use built-in version parsing from packaging

adrianeboyd:
Use version parsing from `packaging` to support all valid python version identifiers instead of the external `semver` library.


helpmefindaname (approved):
Thank you for your contribution

alanakbik:
@adrianeboyd thanks for adding this!

Issue #3509, opened 2024-07-24
[Bug]: Support metric is reported incorrectly for multi-class classifier model evaluation


MattGPT-ai:
### Describe the bug

The classification report returned from running evaluate for a classifier model reports accuracy instead of support when it is not a multi-label classifier

### To Reproduce

```python
import flair

result = trainer.model.evaluate(test_set, gold_label_type=label_type)
print(result.classification_report['micro avg']['support'])
```

### Expected behavior

Support should report the integer number of samples 

##### Flair
0.13.1
##### Pytorch
2.3.1+cu121
##### Transformers
4.31.0
#### GPU
True

fkdosilovic:
@alanakbik Looks like this bug can be closed.

PR #3510, opened 2024-07-24
GH-3509: fix support metric

MattGPT-ai:
Addresses https://github.com/flairNLP/flair/issues/3509

Also adds more type hinting and fixes mutable default arguments.

### DIFF
--- a/flair/nn/model.py
+++ b/flair/nn/model.py
@@ -456,10 +456,13 @@ class Classifier(Model[DT], typing.Generic[DT], ReduceTransformerVocabMixin, ABC
             # Otherwise, it is identical to the "macro avg". In this case, we add it to the report.
             if "micro avg" not in classification_report_dict:
                 classification_report_dict["micro avg"] = {}
-                for precision_recall_f1 in classification_report_dict["macro avg"]:
-                    classification_report_dict["micro avg"][precision_recall_f1] = classification_report_dict[
-                        "accuracy"
-                    ]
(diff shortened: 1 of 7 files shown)


MattGPT-ai:
I'm not sure what the error occurring is, it doesn't seem to be related to my changes but maybe you are familiar with it

MattGPT-ai:
Looks like the test passed without me changing anything, so this should be ready for review

flair/nn/model.py
```diff
@@ -737,7 +743,8 @@ def _encode_data_points(self, sentences: List[DT], data_points: List[DT2]):
 
         return data_point_tensor
 
-    def _mask_scores(self, scores, data_points):
+    def _mask_scores(self, scores: Tensor, data_points) -> Tensor:
+        """This appears to have not been fully implemented and simply returns scores."""
```
alanakbik, line 747: In most models, scores are not masked, so the default implementation is to just return the scores. The `SpanClassifier` overwrites this method to account for "candidate lists". Anything not on these lists, will be masked.

alanakbik, line 747:
```suggestion
        """Classes that inherit from DefaultClassifier may optionally mask scores."""
```

alanakbik (changes requested):
Looks good, thanks for adding this! Just needs a tiny modification in one of the comment lines, as otherwise it would be misleading.

alanakbik:

PR #3512, opened 2024-07-24
Add support for NoiseBench

elenamer:
This PR adds support for [NoiseBench](https://arxiv.org/abs/2405.07609), a subset of CoNLL-03 for English with 7 different label sets, including 6 different types of label noise.

Example usage (intialize dataset with a crowdsourced version of the NER labels):
```
corpus = NER_NOISEBENCH(noise="crowd")
```

Potential issues:

To create the dataset, we need the [CleanCoNLL](https://aclanthology.org/2023.emnlp-main.533.pdf) labels as gold labels. This required running an external script in this implementation: https://github.com/flairNLP/CleanCoNLL/blob/main/create_cleanconll_from_conll03.sh


flair/datasets/sequence_labeling.py
```diff
@@ -4973,6 +4973,209 @@ def _write_instances(cls, version, base_path, split, data):
...
+                crowd,crowdbest,expert,distant,weak,llm : Different kinds of noisy labelsets (details: ...)
+            base_path (Optional[Union[str, Path]]): Path to the data.
+                Default is None, meaning the corpus gets automatically downloaded and saved.
+                You can override this by passing a path to a directory containing the unprocessed files but typically this
+                should not be necessary.
+            in_memory (bool): If True the dataset is kept in memory achieving speedups in training.
+            **corpusargs: The arguments propagated to :meth:'flair.datasets.ColumnCorpus.__init__'.
+        """
+        if noise not in ["clean", "crowd", "crowdbest", "expert", "distant", "weak", "llm"]:
+            raise Exception("Please choose a valid version")
```
fkdosilovic, line 5001: It is recommended to raise [ValueError](https://docs.python.org/3/library/exceptions.html#:~:text=Passing%20arguments%20of%20the%20wrong%20type%20(e.g.%20passing%20a%20list%20when%20an%20int%20is%20expected)%20should%20result%20in%20a%20TypeError%2C%20but%20passing%20arguments%20with%20the%20wrong%20value%20(e.g.%20a%20number%20outside%20expected%20boundaries)%20should%20result%20in%20a%20ValueError.) for wrong and unexpected values passed to the argument.

Also, an error message would be more helpful if it listed valid values. Something like:
```python
if noise not in VALID_NOISE_VALUES:
    raise ValueError(f"Unsupported value for noise argument. Got {noise}, expected one of {VALUE_NOISE_VALUES}!")
```

flair/datasets/sequence_labeling.py
```diff
@@ -4973,6 +4973,209 @@ def _write_instances(cls, version, base_path, split, data):
...
+
+        self._set_path(base_path)
```
fkdosilovic, line 5003: I would remove this function and just set the `base_path` here in `__init__`:
```python
if base_path:
    self.base_path = Path(base_path)
else:
    self.base_path = flair.cache_root / "datasets" / "noisebench"
```

flair/datasets/sequence_labeling.py
```diff
@@ -4973,6 +4973,209 @@ def _write_instances(cls, version, base_path, split, data):
...
+            column_delimiter="\t",
+            document_separator_token="-DOCSTART-",
+            **corpusargs,
+        )
+
+    def _set_path(self, base_path):
+        self.base_path = flair.cache_root / "datasets" / "noisebench" if not base_path else Path(base_path)
+
+    @staticmethod
+    def read_column_file(filename):
```
fkdosilovic, line 5070: You are adding a new public method to `NER_NOISEBENCH`. I would avoid this. Also, you are relying on this method only in non-public methods in the class (except in `generate_data_files`, which I would also put as non-public.)

My suggestion is to rename the method to `_read_column_file` and add type annotations for `filename` argument.

flair/datasets/sequence_labeling.py
```diff
@@ -4973,6 +4973,209 @@ def _write_instances(cls, version, base_path, split, data):
...
+                    point = []
+
+        if len(point) > 0:
+            all_x.append(point)
+
+        all_x = all_x
+        return all_x
+
+    @staticmethod
+    def save_to_column_file(filename, list):
```
fkdosilovic, line 5092: 1. Rename to `_save_to_column_file`
2. Add type annotations to arguments
3. Avoid using the name of built-in modules/types/etc. for argument and variable names. Here, you are using `list` as the name for the arguments. Seems like a more appropriate name is `sentences` which is of type `list`.

flair/datasets/sequence_labeling.py
```diff
@@ -4973,6 +4973,209 @@ def _write_instances(cls, version, base_path, split, data):
...
+
+            for token, label in zip(clean_sentence, sentence):
+                label[0] = token[0]  # token[0] -> text, token[1] -> BIO label
+        if self.SAVE_TRAINDEV_FILE:
+            self.save_to_column_file(os.path.join(self.base_path, f"{corpus}.traindev"), noisy_labels)
+        return noisy_labels
+
+    def generate_data_files(self, filename):
+
+        with open(os.path.join(self.base_path, "annotations_only", "index.txt")) as index_file:
```
fkdosilovic, line 5154: You can avoid this `os.path.join` since `base_path` is of type Path. Just use `/` to concatenate the path. Also, we should avoid relying on implicit default values, it is better to explicitly mention that we want to read the file with utf-8:
```python
with open(self.base_path / "annotations_only" / "index.txt", "r", encoding="utf-8") as fp:
    # ...
```

flair/datasets/sequence_labeling.py
```diff
@@ -4973,6 +4973,209 @@ def _write_instances(cls, version, base_path, split, data):
...
+            token_indices = index_file.readlines()
+
+            all_clean_sentences = self.read_column_file(os.path.join(self.cleanconll_base_path, "cleanconll.train"))
```
fkdosilovic, line 5157: Same here, use `self.cleanconll_base_path / "cleanconll.train"` instead of `os.path.join`.

flair/datasets/sequence_labeling.py
```diff
@@ -4973,6 +4973,209 @@ def _write_instances(cls, version, base_path, split, data):
...
+
+            # os.makedirs(os.path.join('data','noisebench'), exist_ok=True)
+
+            noisy_sentences = self._merge_tokens_labels(filename, all_clean_sentences, token_indices)
+            self._create_train_dev_splits(
+                all_sentences=noisy_sentences, filename=os.path.join(self.base_path, f"{filename}.traindev")
+            )
+
+        # copy test set
+        all_clean_test_sentences = self.read_column_file(os.path.join(self.cleanconll_base_path, "cleanconll.test"))
```
fkdosilovic, line 5167: Same here, use `self.cleanconll_base_path / "cleanconll.test"` instead of `os.path.join`.

flair/datasets/sequence_labeling.py
```diff
@@ -4973,6 +4973,209 @@ def _write_instances(cls, version, base_path, split, data):
...
+
+        test_sentences = []
+        for s in all_clean_test_sentences:
+            new_s = []
+            for token in s:
+                new_s.append([token[0], token[4]])
+            test_sentences.append(new_s)
+
+        self.save_to_column_file(os.path.join(self.base_path, "clean.test"), test_sentences)
```
fkdosilovic, line 5176: Same here, use `self.base_path / "clean.test"` instead of `os.path.join`.

flair/datasets/sequence_labeling.py
```diff
@@ -4973,6 +4973,209 @@ def _write_instances(cls, version, base_path, split, data):
...
+
+        # copy test set
+        all_clean_test_sentences = self.read_column_file(os.path.join(self.cleanconll_base_path, "cleanconll.test"))
+
+        test_sentences = []
+        for s in all_clean_test_sentences:
+            new_s = []
+            for token in s:
+                new_s.append([token[0], token[4]])
+            test_sentences.append(new_s)
```
fkdosilovic, line 5174: This can be refactored to a more idiomatic approach with list comprehensions:
```python
for sentence in all_clean_test_sentences:
    new_sentence = [[tokens[0], tokens[4]] for tokens in sentence]
    test_sentences.append(new_sentence)
```

flair/datasets/sequence_labeling.py
```diff
@@ -4973,6 +4973,209 @@ def _write_instances(cls, version, base_path, split, data):
...
+                sentence[0][0] = "-DOCSTART-"
+                continue
+            clean_sentence = all_clean_sentences[int(index.strip())]
+
+            assert len(clean_sentence) == len(sentence)  # this means indexing is wrong
+
+            for token, label in zip(clean_sentence, sentence):
+                label[0] = token[0]  # token[0] -> text, token[1] -> BIO label
+        if self.SAVE_TRAINDEV_FILE:
+            self.save_to_column_file(os.path.join(self.base_path, f"{corpus}.traindev"), noisy_labels)
```
fkdosilovic, line 5149: Same here, use `self.base_path / f"{corpus}.traindev"` instead of `os.path.join`.

flair/datasets/sequence_labeling.py
```diff
@@ -4973,6 +4973,209 @@ def _write_instances(cls, version, base_path, split, data):
...
+            os.sep.join(filename.split(os.sep)[:-1]) + os.sep + filename.split(os.sep)[-1].split(".")[0] + ".train",
+            train_sentences,
+        )
+
+    def _merge_tokens_labels(self, corpus, all_clean_sentences, token_indices):
+        # generate NoiseBench dataset variants, given CleanCoNLL, noisy label files and index file
+
+        noisy_labels = self.read_column_file(os.path.join(self.base_path, "annotations_only", f"{corpus}.traindev"))
+        # print(noisy_labels)
+        # print(token_indices)
```
fkdosilovic, line 5135: Remove these debugging prints.

flair/datasets/sequence_labeling.py
```diff
@@ -4973,6 +4973,209 @@ def _write_instances(cls, version, base_path, split, data):
...
+        )
+        self.save_to_column_file(
+            os.sep.join(filename.split(os.sep)[:-1]) + os.sep + filename.split(os.sep)[-1].split(".")[0] + ".train",
+            train_sentences,
+        )
+
+    def _merge_tokens_labels(self, corpus, all_clean_sentences, token_indices):
+        # generate NoiseBench dataset variants, given CleanCoNLL, noisy label files and index file
+
+        noisy_labels = self.read_column_file(os.path.join(self.base_path, "annotations_only", f"{corpus}.traindev"))
```
fkdosilovic, line 5133: Same here, use `self.base_path / "annotations_only" /  f"{corpus}.traindev"` instead of `os.path.join`.

flair/datasets/sequence_labeling.py
```diff
@@ -4973,6 +4973,209 @@ def _write_instances(cls, version, base_path, split, data):
...
+
+    @staticmethod
+    def read_column_file(filename):
+        with open(filename, errors="replace") as file:
+            lines = file.readlines()
+            all_x = []
+            point = []
+            for line in lines:
+                if "\t" in line.strip():
+                    stripped_line = line.strip().split("\t") if "\t" in line.strip() else line.strip().split(" ")
```
fkdosilovic, line 5077: Looks like `stripped_line = line.strip().split("\t") if "\t" in line.strip()` will always be set due to the condition `if "\t" in line.strip()` in line 5076. Refactor to:
```python
if "\t" in line.strip():
    stripped_line = line.strip().split("\t")
else:
    stripped_line = line.strip().split(" ")

File: flair/datasets/sequence_labeling.py [hunk 14]
```diff
@@ -4973,6 +4973,209 @@ def _write_instances(cls, version, base_path, split, data):
...
+            all_x.append(point)
+
+        all_x = all_x
+        return all_x
+
+    @staticmethod
+    def save_to_column_file(filename, list):
+        with open(filename, "w") as f:
+            for sentence in list:
+                for token in sentence:
```
fkdosilovic, line 5095: Rename the iteration variable to `tokens` to convert the meaning that this is a container of elements and not a single token.

alanakbik:
@fkdosilovic thanks a lot for your review, this is very helpful! Please note that this is still a draft PR and one major change is still planned.

elenamer:
@fkdosilovic Thanks a lot for the review, it was very helpful and I incorporated most of your feedback! 

Also, the CLEANCONLL Corpus is now a part of Flair, so I modified this PR to use it, instead of downloading and running an external .sh script.

alanakbik:
@elenamer thanks for adding this!

Issue #3524, opened 2024-08-09
[Feature]: Refactor fill_mean_token_embeddings for performance optimization on GPU


sheldon-roberts:
### Problem statement

The argument `subtoken_pooling="mean"` runs significantly slower than  `subtoken_pooling="first"` when fine-tuning a `TransformerWordEmbedding`.

### Solution

The function `fill_mean_token_embeddings` can be vectorized to avoid using any loops. This lets us use mean subtoken pooling with essentially no additional training time compared to first subtoken pooling.

### Additional Context

Here is a script demonstrating the potential performance gains with the refactor.

```
from flair.embeddings import TransformerWordEmbeddings
from flair.data import Sentence
import time
import torch
import flair

text = 'The grass is green ' * 50

# Test the speed of embedding with subtoken_pooling="first"
embedding = TransformerWordEmbeddings('xlm-roberta-base', subtoken_pooling="first", fine_tune=True)
sentences = [Sentence(text) for _ in range(100)]
start = time.time()
embedding.embed(sentences)
end = time.time()
print(f'[subtoken_pooling=first]: sentence embeddings took {end - start:.2f} seconds')

# Test the speed of embedding with subtoken_pooling="mean"
embedding = TransformerWordEmbeddings('xlm-roberta-base', subtoken_pooling="mean", fine_tune=True)
sentences = [Sentence(text) for _ in range(100)]
start = time.time()
embedding.embed(sentences)
end = time.time()
print(f'[subtoken_pooling=mean]: sentence embeddings took {end - start:.2f} seconds')

# ----------------------------------------------------------
# now add the optimized fill_mean_token_embeddings function
@torch.jit.script_if_tracing
def fill_mean_token_embeddings_new(
    all_token_embeddings: torch.Tensor,
    sentence_hidden_states: torch.Tensor,
    word_ids: torch.Tensor,
    token_lengths: torch.Tensor,
):
    batch_size, max_tokens, embedding_dim = all_token_embeddings.shape
    mask = word_ids >= 0
    
    # sum embeddings for each token
    all_token_embeddings.scatter_add_(
        1, 
        word_ids.clamp(min=0).unsqueeze(-1).expand(-1, -1, embedding_dim), 
        sentence_hidden_states * mask.unsqueeze(-1).float()
    )
    
    # calculate the mean of subtokens
    subtoken_counts = torch.zeros_like(all_token_embeddings[:, :, 0])
[...]

PR #3525, opened 2024-08-09
GH-3524: Refactor fill_mean_token_embeddings for performance optimization on GPU

sheldon-roberts:
#3524 

This PR implements the optimization of `fill_mean_token_embeddings`, as suggested in the linked issue.


helpmefindaname (approved):
Thank you for contributing this,
I could verify that this is way faster than before

PR #3526, opened 2024-08-09
Ignore FutureWarning

alanakbik:
Current builds are failing due to a new FutureWarning regarding the default behavior of tokenizer.cleanup_tokenization_spaces (see https://github.com/huggingface/transformers/issues/31884). 

This PR simply ignores these warnings since our code is unaffected by it.


alanakbik:
Also removed some dataset tests to reduce the disk usage.

PR #3535, opened 2024-08-22
Add proxies information to requests.head

diego-morientez:
request.head  also require use url_proxies variable

I have been working using this library with a proxy configuration, and I found that this change is necessary to achieve the download of the files


helpmefindaname (approved):
Thank you for your contribution

PR #3538, opened 2024-08-23
fix: use proper eval default main eval metrics for text pair regressor

MattGPT-ai:
The `TextPairRegressor` model uses the wrong default evaluation metric for a regression. This changes it from micro avg f1 score to pearson correlation coefficient


helpmefindaname (approved):
Thank you for this contribution

Issue #3543, opened 2024-09-03
[Bug]: Cannot load pre-trained models after fine-tuning (Transformers)


DhruvSondhi:
### Describe the bug

Hello,

I was trying to fine tune a mT5 model (google/mT5 series models) on a custom dataset that follows the text format given in your documentation for the column data loader. I have been trying to figure out what is happening but I think there is some problem in the way the model is being loaded/saved. I am sharing my files that have changes done to them (uses the base template of [this example](https://github.com/flairNLP/flair/blob/master/examples/ner/run_ner.py)). 

### To Reproduce
`run_ner.py` (I am trying to reproduce results from this repo: https://github.com/MLlab4CS/Astro-mT5/tree/main)
```python

import inspect
import json
import logging
import os
import sys
from dataclasses import dataclass, field

import torch
from transformers import HfArgumentParser

import flair
from flair import set_seed
from flair.embeddings import TransformerWordEmbeddings
from flair.models import SequenceTagger
from flair.trainers import ModelTrainer
from flair.datasets import ColumnCorpus

logger = logging.getLogger("flair")
logger.setLevel(level="INFO")


@dataclass
class ModelArguments:
    model_name_or_path: str = field(
        metadata={"help": "The model checkpoint for weights initialization."},
    )
    layers: str = field(default="-1", metadata={"help": "Layers to be fine-tuned."})
    subtoken_pooling: str = field(
        default="first",
        metadata={"help": "Subtoken pooling strategy used for fine-tuned."},
    )
    hidden_size: int = field(default=256, metadata={"help": "Hidden size for NER model."})
    use_crf: bool = field(default=False, metadata={"help": "Whether to use a CRF on-top or not."})


@dataclass
class TrainingArguments:
    num_epochs: int = field(default=10, metadata={"help": "The number of training epochs."})
    batch_size: int = field(default=8, metadata={"help": "Batch size used for training."})
    mini_batch_chunk_size: int = field(
        default=1,
        metadata={"help": "If smaller than batch size, batches will be chunked."},
    )
    learning_rate: float = field(default=5e-05, metadata={"help": "Learning rate"})
    seed: int = field(default=42, metadata={"help": "Seed used for reproducible fine-tuning results."})
    device: str = field(default="cuda:0", metadata={"help": "CUDA device string."})
    weight_decay: float = field(default=0.0, metadata={"help": "Weight decay for optimizer."})
[...]

PR #3544, opened 2024-09-06
fix T5 tokenizer loading

helpmefindaname:
Closes https://github.com/flairNLP/flair/issues/3543

The T5 tokenizer fast is required to load from the slow tokenizer when `add_prefix_space` is True. See [the code](https://github.com/huggingface/transformers/blob/673440d073d5f534a6d6bedeeca94869afd8d0a7/src/transformers/models/t5/tokenization_t5_fast.py#L113)
I am not sure why exactly this is implemented that way, as the traces do not show a specific reasoning, but with a small exception on our side, we can easily support all T5 tokenizers.

### DIFF
--- a/flair/embeddings/transformer.py
+++ b/flair/embeddings/transformer.py
@@ -26,6 +26,7 @@ from transformers import (
     LayoutLMv2FeatureExtractor,
     PretrainedConfig,
     PreTrainedTokenizer,
+    T5TokenizerFast,
 )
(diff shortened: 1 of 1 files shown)


DhruvSondhi:
Hello,

Yes, this indeed fixes the problem. I was able to load the best model from the set of models trained. Hoping that this could be merged into the master branch as soon as possible. Thanks, @helpmefindaname!

alanakbik:

Issue #3457, opened 2024-05-16
[Question]: Release of Biaffine/SpanTagger


stefan-it:
### Question

Hi everyone,

I would like to experiment with the Biaffine/SpanTagger implementation that is e.g. used in the great [[CleanCoNLL](https://aclanthology.org/2023.emnlp-main.533.pdf) paper. I could reproduce the results using FLERT, but I wanted to try out the Biaffine/SpanTagger approach as well.

In the Flair repo itself I could find some branches about this topic:

* https://github.com/flairNLP/flair/tree/span_tagger_refactoring
* https://github.com/flairNLP/flair/tree/span_tagger_gazetteer_mining
* https://github.com/flairNLP/flair/tree/span_tagger_fast
* https://github.com/flairNLP/flair/tree/span_classifier_ner_type
* https://github.com/flairNLP/flair/tree/bf_span_tagger_fast
* https://github.com/flairNLP/flair/tree/add_span_tagger
* https://github.com/flairNLP/flair/tree/add_span_tagger_mod

It would be really awesome to get this approach merged into Flair master, I could help integrating and testing as well :)

Many thanks!

PR #3557, opened 2024-10-14
Add CleanCoNLL object

susannaruecker:
Here's the PR for adding a CleanCoNLL object. Simple usage:

```python
from flair.datasets import CLEANCONLL

cleanconll = CLEANCONLL()
print(cleanconll)
```

When called for the first time, this will download the necessary files, so
* the original CoNLL-03 from https://data.deepai.org/conll2003.zip
* our patch files concerning new line breaks, and our annotation files (where the tokens are masked) from the [CleanCoNLL repo](https://github.com/flairNLP/CleanCoNLL/tree/main/data)

It then applies the patch files to the original CoNLL-03 tokens (for our new line break etc.) and then merges those new tokens with our new annotations.

Note: As requested, I replaced all previous usage of subcalling bash scripts with pure python. Especially the patching process which until now was 
```python
subprocess.run(['patch', str(file_path), str(patch_path), '-o', output_path])
```
is now done with own methods, but unfortunately rather lengthy now...

Please check if it works for you! :slightly_smiling_face: (If you already have the CleanCoNLL files in your `.flair/datasets` folder you should delete those before, otherwise those simply will be read and the new reconstruction is not tested)


flair/datasets/sequence_labeling.py
```diff
@@ -1425,6 +1430,233 @@ def __init__(
...
+
+        def parse_patch(patch_file_path):
+            """
+            Parses a patch file and returns a structured representation of the changes.
+            """
+
+            changes = []
+            current_change = None
+
+            with open(patch_file_path, 'r') as patch_file:
```
elenamer, line 1497: it's better to also specify an encoding when reading/writing files, as different os have different default encodings

elenamer (changes requested):
Thanks a lot for adding this and implementing the patching in python! :) 

I'm getting an error ```TypeError: CLEANCONLL.download_and_prepare_data() takes 1 positional argument but 2 were given```, and I saw that the same one is appearing when the tests are run. 

Also, so that all checks pass, you need to run mypy, ruff and black for checks and formatting

stefan-it:
Hi @susannaruecker ,

many thanks for adding this! I have trained [FLERT models](https://huggingface.co/collections/stefan-it/fine-tuned-cleanconll-models-6685d2e4852db86b9ca90dba) on the CleanCoNLL dataset with very great results.

I have one question about further experiments with that dataset: is it possible that the SpanTagger training also gets officially integrated into Flair - I opened this issue some time ago #3457 and it would be awesome to have support in Flair for this approach as well :)

Issue #3552, opened 2024-10-07
[Bug]: Error when cls_pooling="mean" or cls_pooling="max"


chelseagzr:
### Describe the bug

When `cls_pooling` is `"mean"` or `"max"`, document embeddings fail to be created.
After digging into the codes, I think the issue is:  `return result` is missing in the functions [document_mean_pooling](https://github.com/flairNLP/flair/blob/e17ab1234fcfed2b089d8ef02b99949d520382d2/flair/embeddings/transformer.py#L183) and [document_max_pooling](https://github.com/flairNLP/flair/blob/e17ab1234fcfed2b089d8ef02b99949d520382d2/flair/embeddings/transformer.py#L193).

### To Reproduce

```python
from flair.embeddings import TransformerDocumentEmbeddings
from flair.data import Sentence

embeddings = TransformerDocumentEmbeddings(
    model="xlm-roberta-base",
    layers="-1",
    cls_pooling="mean",
    allow_long_sentences=True
)
sentence = Sentence("Today is a good day")
embeddings.embed(sentence)
```

### Expected behavior

Document embeddings using mean pooling are created for the sentence.

### Logs and Stack traces

```stacktrace
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In[4], line 11
      4 embeddings = TransformerDocumentEmbeddings(
      5     model="xlm-roberta-base",
      6     layers="-1",
      7     cls_pooling="max",
      8     allow_long_sentences=True
      9 )
     10 sentence = Sentence("Today is a good day")
---> 11 embeddings.embed(sentence)

File /pyzr/active_venv/lib/python3.12/site-packages/flair/embeddings/base.py:50, in Embeddings.embed(self, data_points)
     47     data_points = [data_points]
     49 if not self._everything_embedded(data_points):
---> 50     self._add_embeddings_internal(data_points)
     52 return data_points

File /pyzr/active_venv/lib/python3.12/site-packages/flair/embeddings/transformer.py:709, in TransformerBaseEmbeddings._add_embeddings_internal(self, sentences)
    707 if self.document_embedding:
    708     document_embedding = embeddings["document_embeddings"]
--> 709     self.__extract_document_embeddings(document_embedding, sentences)
    711 if self.token_embedding:
    712     token_embedding = embeddings["token_embeddings"]

[...]

PR #3558, opened 2024-10-19
Fix error when cls_pooling="mean" or cls_pooling="max" for TransformerDocumentEmbeddings

fkdosilovic:
* Refactor `"cls"` pooling option into a function.
* Return `result` from `document_mean_pooling` and `document_max_pooling` functions.
* Add unit tests for testing each cls pooling option.
* Refactor `test_default_embeddings_stay_the_same_after_saving_and_loading` and `test_embeddings_stay_the_same_after_saving_and_loading` in `BaseEmbeddingsTest` to use `torch.allclose`

Fixes https://github.com/flairNLP/flair/issues/3552.


tests/embedding_test_utils.py
```diff
@@ -176,9 +184,17 @@ def test_default_embeddings_stay_the_same_after_saving_and_loading(self):
...
+                    token_old.get_embedding(names_old),
+                    token_new.get_embedding(names_new),
+                    atol=1e-6,
+                )
         if self.is_document_embedding:
-            assert (sentence_old.get_embedding(names_old) == sentence_new.get_embedding(names_new)).all()
+            assert torch.allclose(
+                sentence_old.get_embedding(names_old),
+                sentence_new.get_embedding(names_new),
+                atol=1e-6,
```
helpmefindaname, line 196: Can you please revert those changes?

The bug that transformer embeddings weren't the same after reloading are fixed on master now.

helpmefindaname (changes requested):

tests/embedding_test_utils.py
fkdosilovic: > Can you please revert those changes?

Just to clarify, do you want me to revert all the changes were I introduce `torch.allclose` or just the ones for `if self.is_document_embedding` branching?

helpmefindaname: yes exactly, the test is intended to ensure that embeddings work the same even when reloading them. Hence, it is expected to have the exact same values, as the same computer architecture is used

fkdosilovic: Done.

helpmefindaname:
Thank you for the bugfix @fkdosilovic
I will merge as soon as the tests complete

helpmefindaname (approved):

PR #3560, opened 2024-10-25
drop python 3.8

helpmefindaname:
Since python3.8 is no longer supported, we can now drop it and use python3.9 features in our code, especially improving typing.

This pr does:
* increase the minimum python version to 3.9 and uses it in gh-actions
* update ruff to version 0.7 & fixes new linting errors
* uses the python3.9 typing features
* removes some methods that aren't used since longer
* ensure that transformer embeddings will stay with the attention implementation that were saved, preventing precision issues after reloading.


alanakbik (approved):

alanakbik:
@helpmefindaname thanks a lot for the PR!

PR #3569, opened 2024-11-22
perf: optimize dictionary items check

MattGPT-ai:
perf: optimize dictionary items by providing function to check for presence of label in constant time, rather than requiring the creation of a list to be checked. update slow uses of get_items() to use this new function.

### DIFF
--- a/flair/data.py
+++ b/flair/data.py
@@ -122,18 +122,18 @@ class Dictionary:
         return list(results)
 
     def get_items(self) -> list[str]:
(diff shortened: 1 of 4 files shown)


helpmefindaname (approved):

helpmefindaname:
Thank you for your contribution :)

alanakbik:
Awesome, thanks a lot for improving this @MattGPT-ai!

PR #3573, opened 2024-12-02
Test new docstrings and apidocs deployment

alanakbik:
Adds docstrings to the `get_label` method and tests whether the "Returns:" description in the docstrings is correctly displayed.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -143,3 +143,7 @@ smv_outputdir_format = "{ref.name}"
 smv_prefer_remote_refs = False
 
 html_favicon = "_static/favicon.ico"
+
+# Define a mapping for shorter module names in the documentation
+# (for example, `car` instead of `longpath.to.module.car`)
+modindex_common_prefix = ["flair.", "flair.data"]
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -112,8 +112,6 @@ autodoc_default_options = {
     "inherited": True,
 }
 
-autodoc_add_module_names = False
-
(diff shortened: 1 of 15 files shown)


alanakbik:
FYI @helpmefindaname @konstantin-lukas - examples are not yet formatted in the template. For now, I'll just exclude examples from the code comments. 

Here's how the following docstring looks like:

```python
 def add_label(self, typename: str, value: str, score: float = 1.0, **metadata) -> "DataPoint":
        """Adds a label to the :class:`DataPoint` by internally creating a :class:`Label` object.

        Examples:
            >>> sentence = Sentence("this is good")
            >>> sentence.add_label("sentiment", "positive")
            >>> print(sentence)
            Sentence[3]: "this is good" → positive (1.0000)

            >>> # make a sentence and get the last token
            >>> sentence = Sentence("this is good")
            >>> token = sentence[2]
            >>> # add a label to token
            >>> token.add_label("part-of-speech", "adjective")
            >>> print(token)
            Token[2]: "good" → adjective (1.0000)

        Args:
            typename: A string that identifies the layer of annotation, such as "ner" for named entity labels or "sentiment" for sentiment labels
            value: A string that sets the value of the label.
            score: Optional value setting the confidence level of the label (between 0 and 1). If not set, a default confidence of 1 is used.
            **metadata: Additional metadata information.

        Returns:
            A pointer to itself (DataPoint object, now with an added label).
        """
```

image

helpmefindaname (approved):

PR #3583, opened 2024-12-17
Multigpu: Fix gradient accumulation and learning rate aggregation

jeffpicard:
Hi! This PR fixes 3 bugs with multi-gpu training:
- When using gradient accumulation (mini_batch_chunk_size), this PR disables gradient syncing for allforward/backward passes except the last one before the step. These intermediate syncs are unnecessary and add communication that decreases efficiency. In my experience, this effect was significant enough to go from multi_gpu not giving any speedup, to a 80%-of-linear speedup (mini_batch_chunk_size ~= 4, sentence_length ~= 500, model=xlm-roberta-base, gpu_memory ~= 24g). This is similar to what's done [here](https://discuss.pytorch.org/t/gradient-accumulation-with-ddp-no-sync-interface/169593).
- Sums rather than averages gradients from multiple processes. Flair tracks losses / grads as the sum of the whole batch. However, Torch DDP averages the whole batch when aggregating multiple processes. This led to the effective learning rate being 1 / n_gpus smaller when using `multi_gpu=True`, but ideally it should stay the same to let knobs be turned independently, and this PR does that.
- Fixes a deadlock in the checkpoint plugin. When using `multi_gpu=True` and `save_model_each_k_epochs>0` training will freeze. Currently the plugin runs `barrier` which waits for other processes to catchup, but the plugin is only run on the main process, so they never will.

### DIFF
--- a/flair/trainers/plugins/functional/checkpoints.py
+++ b/flair/trainers/plugins/functional/checkpoints.py
@@ -30,8 +30,6 @@ class CheckpointPlugin(TrainerPlugin):
             )
             model_name = "model_epoch_" + str(epoch) + ".pt"
             self.model.save(self.base_path / model_name, checkpoint=self.save_optimizer_state)
-            if torch.distributed.is_initialized():
-                torch.distributed.barrier()  # Prevent any process from loading a model until writing is complete
 
     @property
     def attach_to_all_processes(self) -> bool:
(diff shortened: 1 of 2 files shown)


alanakbik:
@jeffpicard thanks for improving this!

Issue #3596, opened 2025-01-14
[Feature]: Adding an encoder-decoder model where encoder and decoder are both transformers


chelseagzr:
### Problem statement

Encoder-decoder models, where both the encoder and decoder are transformers, have proven to be highly effective for seq2seq tasks, such as translation. It would be extremely helpful if Flair supports this type of model.

### Solution

It would be nice if a user can easily create an encoder-decoder model by specifying the name of the encoder (e.g., xlm-roberta-base) and the name of the decoder (e.g., GPT2). Developers can make use of the `EncoderDecoderModel` class from `transformers` module to facilitate this.

PR #3597, opened 2025-01-14
Adding a transformer based encoder-decoder model

chelseagzr:
This PR is aimed to address this [feature](https://github.com/flairNLP/flair/issues/3596) I proposed.
Please let me know if you have any suggestions for this model! Thank you!

### DIFF
--- a/flair/models/encoder_decoder_model.py
+++ b/flair/models/encoder_decoder_model.py
@@ -1,0 +1,590 @@
+import logging
+import math
+from pathlib import Path
+from typing import List, Dict, Tuple, Union, Optional, Any, Callable
+import inspect
+import ast
+
+import torch
+from torch import nn
+from torch.nn.utils.rnn import pad_sequence
+from transformers import (
+    AutoModelForCausalLM,
+    AutoTokenizer,
+    EncoderDecoderModel,
+    EncoderDecoderConfig,
+    GenerationMixin,
+    PreTrainedModel
+)
+from transformers.modeling_outputs import Seq2SeqLMOutput
+
+import flair
+from flair.data import DataPoint
+from flair.datasets import FlairDatapointDataset
+from flair.nn import Model
+from flair.training_utils import Result
+from flair.embeddings.base import load_embeddings
+
+
+
+logger = logging.getLogger("flair")
+
+
+# copied from the _tie_encoder_decoder_weights method of transformers.PreTrainedModel
+# but added shape check to make it safer
+@staticmethod
+def _tie_encoder_decoder_weights(
+    encoder: nn.Module, decoder: nn.Module, base_model_prefix: str, base_encoder_name: str
+):
+    uninitialized_encoder_weights: List[str] = []
+    tied_weights: List[str] = []
+    if decoder.__class__ != encoder.__class__:
+        logger.info(
+            f"{decoder.__class__} and {encoder.__class__} are not equal. In this case make sure that all encoder"
+            " weights are correctly initialized."
+        )
+
+    def tie_encoder_to_decoder_recursively(
+        decoder_pointer: nn.Module,
+        encoder_pointer: nn.Module,
+        module_name: str,
+        base_encoder_name: str,
+        uninitialized_encoder_weights: List[str],
+        depth=0,
+        total_decoder_name="",
+        total_encoder_name="",
+    ):
+        assert isinstance(decoder_pointer, nn.Module) and isinstance(
+            encoder_pointer, nn.Module
+        ), f"{decoder_pointer} and {encoder_pointer} have to be of type nn.Module"
+        if hasattr(decoder_pointer, "weight"):
+            assert hasattr(encoder_pointer, "weight")
+            # added codes start here
+            if encoder_pointer.weight.shape != decoder_pointer.weight.shape:
+                uninitialized_encoder_weights.append(f"{base_encoder_name}{total_encoder_name}")
+                return
+            # added codes end here
+            encoder_pointer.weight = decoder_pointer.weight
+            tied_weights.append(f"{base_encoder_name}{total_encoder_name}.weight")
+            if hasattr(decoder_pointer, "bias"):
+                assert hasattr(encoder_pointer, "bias")
+                tied_weights.append(f"{base_encoder_name}{total_encoder_name}.bias")
+                encoder_pointer.bias = decoder_pointer.bias
+            return
+
+        encoder_modules = encoder_pointer._modules
+        decoder_modules = decoder_pointer._modules
+        if len(decoder_modules) > 0:
+            assert (
+                len(encoder_modules) > 0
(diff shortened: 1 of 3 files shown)


flair/models/encoder_decoder_model.py
```diff
@@ -0,0 +1,556 @@
...
+
+    if len(uninitialized_encoder_weights) > 0:
+        logger.warning(f"The following encoder weights were not tied to the decoder {uninitialized_encoder_weights}")
+    # added codes start here
+    logger.warning(f"The following encoder weights were tied to the decoder {tied_weights}")
+    # added codes end here
+    return tied_weights
+
+
+PreTrainedModel._tie_encoder_decoder_weights = _tie_encoder_decoder_weights
```
helpmefindaname, line 129: Is the shape check really necessary?
I would prefer to need as little patching of external functions as possible, as that function is always subject of change, increasing the bourdon of maintainance.
Maybe it would be an better idea to make a PR to transformers and add the shapecheck there?

flair/models/encoder_decoder_model.py
```diff
@@ -0,0 +1,556 @@
...
+
+
+def retrieve_function_from_definition(func_definition: str) -> Callable:
+    """Creates a function object from the definition of a function.
+
+    Args:
+        func_definition (str): the definition of a single function
+    """
+    local_scope = {}
+    exec(func_definition, local_scope)
```
helpmefindaname, line 139: This allows remote code execution and is something that we should be avoided.

I think it would be fair to always expect a SentencePair as input and add an optional parameter `prompt_template`.
Then you can set `prompt_template="[TASK] translate German to English: [TASK] {text}"` for the opus example.
Is that enough?

helpmefindaname (changes requested):

flair/models/encoder_decoder_model.py
chelseagzr: I see. I think it's good to have the shape check, but I can remove this patch for now. It might be better to make a PR to transformers.

flair/models/encoder_decoder_model.py
chelseagzr: Thank you for the feedback!
In the second example above, I believe an additional parameter (`use_first_as_input`) is required to train a model to translate between English and German using a single underlying corpus. This parameter specifies which sentence in the sentence pair should be treated as the input and which as the output.
Here are some code to illustrate the use of the additional parameter `use_first_as_input`:


```
class EncoderDecoderLanguageModel(Model, GenerationMixin):
    """A language model based on an encoder-decoder architecture using HuggingFace's Transformers."""

    label_pad_token_id: int = -100  # The index to ignore when calculating cross_entropy loss

    def __init__(
        self,
        encoder_embeddings: Any,
        decoder: CausalLanguageModelDecoder,
        label_type: str,
        use_first_as_input: bool = True,
        task_prompt: str = "",
        tie_encoder_decoder: bool = False,
    ) -> None:
        super().__init__()

        self.encoder_embeddings = encoder_embeddings
        self.decoder = decoder

        self._label_type = label_type
        self._use_first_as_input = use_first_as_input
        self._task_prompt = task_prompt
        self.tie_encoder_decoder = tie_encoder_decoder

        # Initialize EncoderDecoderModel
        ......

    def forward_loss(self, datapoints: list[DataPoint]) -> tuple[torch.Tensor, int]:
        if len(datapoints) == 0:
            raise RuntimeError("No datapoints provided")

        if self._use_first_as_input:
            input_texts = [self._task_prompt + s.first.text for s in datapoints]
            target_texts = [s.second.text for s in datapoints]
        else:
            input_texts = [self._task_prompt + s.second.text for s in datapoints]
            target_texts = [s.first.text for s in datapoints]

        encoder_inputs = self.encoder_tokenizer(
            input_texts,
            padding="longest",
            truncation=True,
            return_tensors="pt",
        ).to(flair.device)

        ...... 
```
(similar for `evaluate` and `predict`)

Then, we instantiate tasks by providing `use_first_as_input` and `task_prompt`:

```
edm1 = EncoderDecoderLanguageModel(
    encoder_embeddings=encoder_embedding,
    decoder=decoder,
    label_type="translate_de_to_en",
    use_first_as_input=True,
    task_prompt="[TASK] translate German to English: [TASK] "
)

edm2 = EncoderDecoderLanguageModel(
    encoder_embeddings=encoder_embedding,
    decoder=decoder,
    label_type="translate_en_to_de",
    use_first_as_input=False,
    task_prompt="[TASK] translate English to German: [TASK] "
)
```

chelseagzr: The `prompt_template` and `use_first_as_input` parameters should be sufficient to replace the `generate_input_text_fn` and `generate_output_text_fn` functions in the examples I provided earlier. However, I’m considering using the functions for something beyond simple string concatenation.
For example, I store a list of strings in the `_metadata` attribute of a datapoint and I want to use the concatenation of these strings as the input for the encoder-decoder model. To make the model invariant to the order of the strings in the input, I want to shuffle the list each time the datapoint is used.

Assume I have a list of strings under the key "ingredients" in self._metadata of a datapoint:
`self._metadata = {"ingredients": ["apple", "banana", "pear"]}`
For each epoch, I want to see different permutations of the concatenated strings, such as "apple banana pear", "banana apple pear", "pear apple banana", etc.
With the `generate_input_text_fn` function, this could be achieved by defining:
```
def input_text_fn(datapair):
    ingredients = datapair.first.get_metadata("ingredients").copy()
    random.shuffle(ingredients)
    return " ".join(ingredients)
```

Do you have any suggestions on how to accommodate this use case without saving and loading functions? Thank you!

helpmefindaname: I think designwise it is not necessary to have `use_first_as_input` as you can just create different corpora:
```
corpus_de_en = OpusParallelCorpus(
    dataset="tatoeba",  # "tatoeba" or "subtitles" are supported in this example
    l1="de",
    l2="en",
    max_tokens_per_doc=512,
)

corpus_en_de = OpusParallelCorpus(
    dataset="tatoeba",  # "tatoeba" or "subtitles" are supported in this example
    l1="en",
    l2="de",
    max_tokens_per_doc=512,
)
```

When you want to do data-augmentation like with the metadata, you should build your dataset in that way. E.g.:'

```
class DataAugmentatedDataset(FlairDataset):
   def __init__(self, orig_dataset: FlairDataset) -> None:
       self.orig_dataset = orig_datset
       
   def __len__(self) -> int:
       return len(self.orig_dataset)
       
   def __getitem__(i: int) -> DataPoint:
       return self.transform(self.orig_dataset[i])
       
   def transform(dp: dataPoint) -> DataPoint:
        ....
```

As you wouldn't want that data-augmentation on inference (and also not in the test set).

chelseagzr: Thank you for the response! I'll remove `generate_input_text_fn` and `generate_output_text_fn` and only add `prompt_template`.

PR #3601, opened 2025-01-25
fix: cast indices tensor to int to fix bug

MattGPT-ai:
This fixes a bug where `index_select` is expecting an int type:

`scores = torch.index_select(scores, 0, torch.tensor(filtered_indices, device=flair.device))`

`RuntimeError: index_select(): Expected dtype int32 or int64 for index`


alanakbik:
Thanks for fixing this!

PR #3602, opened 2025-01-26
fix: use proper eval default main eval metrics for text regression model

MattGPT-ai:
This fixes the default metrics used in the text regression model to be more appropriate for regression, rather than the classification default. There are "correlation" and "pearson" metrics.

This is mostly a copy of https://github.com/flairNLP/flair/pull/3538 but for the `TextRegressor` model

### DIFF
--- a/flair/models/pairwise_regression_model.py
+++ b/flair/models/pairwise_regression_model.py
@@ -345,7 +345,7 @@ class TextPairRegressor(flair.nn.Model[TextPair], ReduceTransformerVocabMixin):
                 f"spearman: {metric.spearmanr():.4f}"
             )
 
-            scores = {
+            eval_metrics = {
                 "loss": eval_loss.item(),
                 "mse": metric.mean_squared_error(),
                 "mae": metric.mean_absolute_error(),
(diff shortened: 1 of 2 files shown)


alanakbik:
@MattGPT-ai thanks for fixing this! Tested with: 

```python
from flair.datasets import WASSA_FEAR
from flair.models import TextRegressor

# load pairwise regression dataset
corpus = WASSA_FEAR().downsample(0.1)

# simple text pair regressor
model = TextRegressor(document_embeddings=TransformerDocumentEmbeddings("distilbert-base-uncased"),
                      label_name="class")

# train the model and be sure to pass the correct main_evaluation_metric
trainer = ModelTrainer(corpus=corpus, model=model)

trainer.fine_tune("resources/taggers/wassa-regression",
                  main_evaluation_metric=("correlation", "spearman"),
                  train_with_dev=True,
                  monitor_test=True,
                  max_epochs=2,
                  )
```

Only indirectly connected to this PR, but from a usability standpoint, it might be nicer to not have to explicitly set the `main_evaluation_metric `in the trainer for regression models. Perhaps in a future PR we could put in a check that - if no `main_evaluation_metric `is provided - takes the default metric from the Model.

PR #3605, opened 2025-01-27
Add per-task metrics

ntravis22:
This will make per-task metrics (from MultitaskModel) available to _publish_eval_result. Without this, the metrics are not recorded, so no logging is possible.


alanakbik:
@ntravis22 thanks for adding this!

PR #3606, opened 2025-01-28
fix: update scipy .A to toarray()

sg-wbi:
scipy has removed support for the .A method in sparse arrays (https://github.com/scipy/scipy/issues/21049). Replaced this with .toarray().


alanakbik:
@sg-wbi thanks for fixing this!

Issue #3620, opened 2025-02-21
[Bug]: Language Order in OpusParallelCorpus


chelseagzr:
### Describe the bug

When instantiating an OpusParallelCorpus with language pairs such as ("de", "en") or ("en", "de"), the resulting corpus consistently has German as the first language and English as the second language, regardless of the input order.

Upon reviewing the source code of OpusParallelCorpus, it appears that the [following code](https://github.com/flairNLP/flair/blob/c6b053d69fc5f5676e49e06f7354bd7757864685/flair/datasets/text_text.py#L83):
```
if l1 > l2:
    l1, l2 = l2, l1
```
forces the languages to be ordered lexicographically based on their language codes. This results in German being treated as the first language in both ("de", "en") and ("en", "de").

### To Reproduce

```python
from flair.datasets import OpusParallelCorpus

corpus_de_en = OpusParallelCorpus(
    dataset="tatoeba",
    l1="de",
    l2="en",
    max_tokens_per_doc=512,
)

corpus_en_de = OpusParallelCorpus(
    dataset="tatoeba",
    l1="en",
    l2="de",
    max_tokens_per_doc=512,
)

# Both corpora consist of (German, English) pairs
print(corpus_de_en.train[0])
>> DataPair: 'Sentence[5]: "Ich muss schlafen gehen."' + 'Sentence[7]: "I have to go to sleep."'
print(corpus_en_de.train[0])
>> DataPair: 'Sentence[5]: "Ich muss schlafen gehen."' + 'Sentence[7]: "I have to go to sleep."'
```

### Expected behavior

Sentence pairs in `corpus_de_en` are (German, English) while sentence pairs in `corpus_en_de` are (English, German)

### Logs and Stack traces

```stacktrace

```

##### Flair
0.15.1
##### Pytorch
2.6.0+cu124
##### Transformers
4.49.0
#### GPU
False

PR #3624, opened 2025-02-28
fix opus corpus and fix ruff errors

helpmefindaname:
fixes https://github.com/flairNLP/flair/issues/3620


alanakbik:

Issue #3623, opened 2025-02-26
[Bug]: Can not load model using embeddings with PEFT


MattGPT-ai:
### Describe the bug

When using a LoRA config with transformer embeddings, a model such as `TextClassifier` will fail to load, as the state dict keys have changed

### To Reproduce

```python
import flair
from peft import LoraConfig, TaskType

peft_config = LoraConfig(
    task_type=TaskType.FEATURE_EXTRACTION,
    inference_mode=False,
)

peft_embeddings = flair.embeddings.TransformerEmbeddings(peft_config=peft_config)

model = flair.models.TextClassifier(embeddings=peft_embeddings, label_type='label', label_dictionary=flair.data.Dictionary())

model.save('/tmp/bert_perf.pt')

model = flair.models.TextClassifier.load('/tmp/bert_perf.pt')
```

### Expected behavior

The model to load successfully 

### Logs and Stack traces

```stacktrace
---------------------------------------------------------------------------
RuntimeError                              Traceback (most recent call last)
Cell In[3], line 1
----> 1 model = flair.models.TextClassifier.load('/tmp/bert_perf.pt')

File /pyzr/active_venv/lib/python3.12/site-packages/flair/models/text_classification_model.py:139, in TextClassifier.load(cls, model_path)
    135 @classmethod
    136 def load(cls, model_path: Union[str, Path, dict[str, Any]]) -> "TextClassifier":
    137     from typing import cast
--> 139     return cast("TextClassifier", super().load(model_path=model_path))

File /pyzr/active_venv/lib/python3.12/site-packages/flair/nn/model.py:1012, in DefaultClassifier.load(cls, model_path)
   1008 @classmethod
   1009 def load(cls, model_path: Union[str, Path, dict[str, Any]]) -> "DefaultClassifier":
   1010     from typing import cast
-> 1012     return cast("DefaultClassifier", super().load(model_path=model_path))

File /pyzr/active_venv/lib/python3.12/site-packages/flair/nn/model.py:583, in Classifier.load(cls, model_path)
    579 @classmethod
    580 def load(cls, model_path: Union[str, Path, dict[str, Any]]) -> "Classifier":
    581     from typing import cast
--> 583     return cast("Classifier", super().load(model_path=model_path))

File /pyzr/active_venv/lib/python3.12/site-packages/flair/nn/model.py:200, in Model.load(cls, model_path)
    197 if "__cls__" in state:
    198     state.pop("__cls__")
[...]

PR #3627, opened 2025-03-04
fixes https://github.com/flairNLP/flair/issues/3623: save PEFT config…

MattGPT-ai:
Fixes Flair bug: https://github.com/flairNLP/flair/issues/3623

This simply saves the PEFT config to the transformer embeddings object and exports it as a parameter so it is saved with the model file and loaded when instantiating the embeddings.


MattGPT-ai:
We might actually need a way to load a model in inference mode so that it can set `inference_mode=True` in a model that was saved after training but is being loaded for inference. It would have to set that in the peft config after it was loaded in the state dict but before it was provided to the model instantiation.

alanakbik:

PR #3639, opened 2025-03-13
Add Danish NER corpus

alanakbik:
This PR adds the DANSK corpus for NER in Danish.


alanakbik:
@stefan-it there are some discrepancies in the dataset statistics, but I think this is probably explained by the removal of the OpenSubtitles part of the corpus, as noted here: https://huggingface.co/datasets/chcaa/dansk-ner#domain-distribution

stefan-it:
Hey @alanakbik , thanks for this PR!

I compared it quickly to my draft implementation, and I spotted a few differences - right side is output of this PR, training split:

image

image

 or:

image

alanakbik:
@stefan-it thanks for noticing this! There was an issue with handling of newline characters that were previously simply excluded, causing the token offsets to not line up anymore. The first sentence in your example was prefaced by a single newline token, which was excluded, explaining the offsets. 

With this fix, the statistics of the dataset now fully line up with the statistics posted on https://huggingface.co/datasets/chcaa/dansk-ner#domain-distribution, minus the OpenSubtitles part that was excluded.

flair/datasets/sequence_labeling.py
```diff
@@ -5618,3 +5618,109 @@ def __init__(
...
+                    debug_json_path = base_path / f"{split}_debug.json"
+                    import json
+
+                    # Convert dataset to a list of dictionaries and save with nice formatting
+                    dataset_for_json = [
+                        {"text": item["text"], "tokens": item["tokens"], "ents": item["ents"]} for item in ds
+                    ]
+
+                    with open(debug_json_path, "w", encoding="utf-8") as f_debug:
+                        json.dump(dataset_for_json, f_debug, ensure_ascii=False, indent=2)
```
stefan-it, line 5669: I think lines 5659 - 5669 can be removed due to debugging-only :)

stefan-it, line 5669: I think lines 5659 - 5669 can be removed due to debugging-only :)

stefan-it:
Thanks @alanakbik , current version looks really good, I am just wondering, if we really need empty tokens + `O` in the dataset:

image

I was able to avoid this by using this code fragment:

```
# Write tokens and tags
                            for token, tag in zip(tokens, tags):
                                token_text = text[token["start"] : token["end"]]  # Don't strip the token
                                if token_text:  # Still skip empty tokens
                                    # Replace newlines with space in output to maintain CoNLL format
                                    token_text = token_text.replace("\n", " ")

                                    if not token_text.strip():
                                        continue

                                    f_out.write(f"{token_text}\t{tag}\n")
```

:thinking:

stefan-it:
So I think for reproducibility reasons, we should then stick to the current solution, to really match the numbers from their documentation ☺️

PR #3651, opened 2025-03-25
fix: cast src to the correct type in scatter_add call

MattGPT-ai:
There is a bug that can arise from the assumption of `float32` when casting mask tensors to float when calling `scatter_add`, causing the error
`RuntimeError: scatter(): Expected self.dtype to be equal to src.dtype`
This fix casts them to the float type according to the caller of the function.


alanakbik:
Hi @MattGPT-ai is this ready for merge? I would need to first merge the pther PEFT PR, correct?

alanakbik:

Issue #3636, opened 2025-03-12
[Feature]: CRF decoder for TokenClassifier


alanakbik:
### Problem statement

Our CRF logic is currently tied to the `SequenceTagger` class. However, in most cases we now recommend using the `TokenClassifier` instead, as it inherits from `DefaultClassifier` and thus has access to all decoder classes designed for the `DefaultClassifier`.

### Solution

This PR will explore the viability of adding CRF logic to the `TokenClassifier`, either by directly tying the logic to this class, or by enabling the logic for the `DefaultClassifier`, or by creating a new decoder that can be used by any class.

alanakbik:
No good solution could be found (see #3656)

PR #3656, opened 2025-03-30
Small TokenClassifier optimization and many new docstrings

alanakbik:
This PR came about while investigating #3636 (CRF decoder for the TokenClassifier). Unfortunately, no good solution for the original issue could be found. 

While looking at the code, I found a suboptimal conversion that is called too often. This PR fixes this conversion, however no runtime benefits could be observed in small-scale experiments. It is possible that on larger datasets, this PR gives a very slight speedup during training. 

Additionally, using gemini-2.5-pro, I generated docstrings for all methods and classes in data.py. It worked fairly well, though much manual cleanup was required.

### DIFF
--- a/flair/data.py
+++ b/flair/data.py
@@ -14,6 +14,7 @@ import torch
 from deprecated.sphinx import deprecated
 from torch.utils.data import Dataset, IterableDataset
 from torch.utils.data.dataset import ConcatDataset, Subset
+from torch import device as torch_device  # Import torch.device for type hint
 
 import flair
 from flair.file_utils import Tqdm
@@ -25,6 +26,14 @@ log = logging.getLogger("flair")
 
 
 def _iter_dataset(dataset: Optional[Dataset]) -> typing.Iterable:
+    """Iterates over a Dataset yielding single data points.
+
+    Args:
+        dataset (Optional[Dataset]): The dataset to iterate over.
+
+    Returns:
+        typing.Iterable: An iterable yielding individual data points.
+    """
     if dataset is None:
         return []
     from flair.datasets import DataLoader
@@ -33,6 +42,14 @@ def _iter_dataset(dataset: Optional[Dataset]) -> typing.Iterable:
 
 
 def _len_dataset(dataset: Optional[Dataset]) -> int:
+    """Calculates the length (number of data points) in a Dataset.
+
+    Args:
+        dataset (Optional[Dataset]): The dataset whose length is required.
+
+    Returns:
+        int: The number of data points in the dataset, or 0 if the dataset is None.
+    """
     if dataset is None:
         return 0
     from flair.datasets import DataLoader
@@ -42,6 +59,8 @@ def _len_dataset(dataset: Optional[Dataset]) -> int:
 
 
 class BoundingBox(NamedTuple):
+    """Represents a bounding box with left, top, right, and bottom coordinates."""
+
     left: str
     top: int
     right: int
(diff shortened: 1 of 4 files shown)


alanakbik:
Other points of potential improvement identified by Gemini: 


1. Dictionary.get_idx_for_items Caching: Make the item2idx_not_encoded cache an instance variable computed once, rather than recomputing it on every call. This might save UTF-8 decoding time if the same items are looked up repeatedly.

2. Dictionary.save/load_from_file: Ensure all relevant flags (add_unk, multi_label, span_labels) are saved and loaded consistently. Add defaults/inference during loading for backward compatibility with older dictionary files.

3. Dictionary.load: Use a centralized URL store (like FlairLocator) for resource paths. Improve error handling for invalid names or failed downloads.

4. Label.__init__ / set_value: Add validation to ensure score is always within \[0.0, 1.0].

5. DataPoint.get_embedding: Clarify or refine the sorting/concatenation order when multiple embedding names are provided. Add error handling for incompatible tensor shapes during concatenation.

6. DataPoint.to: Use non_blocking=True for potentially faster async transfers. Handle potential RuntimeError if pin_memory() is called on non-CPU tensors or when CUDA is unavailable.

7. DataPoint.get_label / .tag / .score: Add warnings or deprecate .tag and .score as they rely on ambiguous "first" label ordering. Encourage using get_label() with optional label_type. Ensure get_label() returns a sensible default Label (with score 0.0) when none is found.

8. _PartOfSentence Label Propagation: Ensure that when labels are added/set/removed on a part (Token/Span/Relation), the corresponding update in the parent Sentence.annotation_layers correctly handles label instances and avoids duplicates or dangling references.

9. Token.__init__ / _initialize_from_tokens: Validate whitespace_after is non-negative. Improve handling when adding a Token that already belongs to another Sentence.

10. Span.__new__ / Relation.__new__: Improve handling of identifier collisions in the _known_spans cache (e.g., log warnings more clearly). Ensure _known_spans cache is initialized on the Sentence if accessed. Add checks for empty token lists in constructors.
Sentence._tokenize: Handle potential errors if tokenizer.tokenize_with_offsets isn't available or if _text.index fails (e.g., due to tokenizer outputting strings not found verbatim in the original text).

11. Sentence.get_span/__getitem__ (slice): Handle edge cases like empty slices or invalid indices more gracefully (e.g., return None or raise specific errors).

12. Sentence.get_language_code: Make language detection more robust (handle short texts, import errors, detection failures gracefully, potentially use a more robust library if langdetect is problematic).

13. Sentence.truncate: Ensure robust cleanup of _known_spans and annotation_layers by correctly identifying which spans/relations/labels are invalidated by token removal. Consider resetting internal token indices (_internal_index) after truncation, although this might have side effects.

14. Sentence.retokenize: Improve the logic for mapping old character offsets to new tokens, potentially using more sophisticated alignment algorithms if simple overlap checks are insufficient. Clarify which types of labels (sentence, token, span) are preserved vs. discarded. Consider re-applying sentence-level labels.

15. Corpus.__init__: Refine the logic for sampling missing splits to better handle small dataset sizes (ensure minimum numbers of samples in splits, avoid errors when sizes are too small). Log sampling actions more clearly.

16. Corpus Filtering Methods: Ensure filtering methods correctly handle empty datasets and return the appropriate type (Dataset or Subset). Add progress bars (like Tqdm) for potentially long operations.

17. Corpus._obtain_statistics_for: Make statistics calculation more comprehensive (e.g., add standard deviation for token lengths, provide options for how to count token labels - only specific type, all types, include 'O'). Add progress bars.

18. Corpus.make_label_dictionary: Improve logging to be more informative about counts and filtering. Ensure correct handling of the 'O' tag, especially for token/span tasks. Add clearer error messages if label type isn't found.

19. Corpus.add_label_noise: Add type hints. Improve validation of the noise transition matrix. Ensure random number generation is properly seeded for reproducibility. Clarify score handling for flipped labels.

20. get_spans_from_bio: Improve clarity of the state machine logic. Ensure correct handling of edge cases and score calculation. Consider alternative methods for determining the final span type if simple weighting is insufficient.

PR #3678, opened 2025-06-28
Diagram visualization of the flair codebase

ivanmilevtues:
This PR adds diagram representation of the codebase for flair.
You can see how it renders here: https://github.com/CodeBoarding/GeneratedOnBoardings/blob/main/flair/on_boarding.md

With over 2k forks it is clear that a lot of people play around with the codebase. This said most of them probably want to just adjust certain aspect of the codebase, and with the proposed diagrams they can navigate it much faster and contribute with better understanding. Let me know if you share my view here.

Usually I'd open a discussion first, but they are not enabled for this repo! This said any feedback is more than welcome!

I saw that you are using .rst files for docs (Sphinx?). We have a free ghub action to keep the diagrams up-to-date. So if you like what you see I'd be more than happy to integrate it for you!

Full transparency: we’re exploring this idea as a potential startup, but we’re still early and figuring out what’s actually useful to developers.

### DIFF
--- a/.codeboarding/flair_data_Corpus.md
+++ b/.codeboarding/flair_data_Corpus.md
@@ -1,0 +1,217 @@
+```mermaid
+
+graph LR
+
+    Corpus_Core_Management["Corpus Core Management"]
+
+    Data_Sampling_and_Splitting["Data Sampling and Splitting"]
+
+    Data_Filtering["Data Filtering"]
+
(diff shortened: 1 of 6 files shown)


ivanmilevtues:
Hey a quick update from our side, we've just released our diagram generation engine as an open-source repo: https://github.com/CodeBoarding/CodeBoarding

If you are interested to run it yourself or how exactly the generation works you can take a look!

